PR #3, opened 2017-06-15
remove Swap to BubleSort

nxexox:


0oo0:
Thanks for the PR 🎉. This repository is at initial stages and does not have many algorithms. Can we have more contributions from you please?

nxexox:
Hi!) Of course, I'm only happy to help, and add algorithms there. To begin with, you probably need to rewrite algorithms from the same python repository.

PR #8, opened 2017-10-01
Added linkedlist

dcariotti:
I added linkedlist with these methods: pushback, pushfront, popback and popfront.


dcariotti:
@dynamitechetan Done

Issue #4, opened 2017-09-28
Add more Algorithms


AnupKumarPanwar:
Please add more algorithms to this repo. 

See the structure and README of this repo for reference
https://github.com/TheAlgorithms/Python

DanGarcia595:
I have some implementations that I made in my CIS class last year that I can add.

dynamitechetan:
sure

brainsnail:
Anything in particular you'd like to see added?

jrolheiser-va:
I'd like to contribute, but also think this project could use some restructuring to make it more similar to the other ones in https://github.com/TheAlgorithms. Like breaking out files into common groups like `sorting`, `searching`, `network flow`, etc..

I am also a big fan of unittests for stuff like this. Not sure what others think.

dynamitechetan:
@brainsnail Feel free to add whatever you like.

EyesBear:
Interested.

himanshub16:
Hey,
I would like to add the following topics:
* LinkedList
* Stack
* Queue
* Binary Tree

Will add more to the list if possible. :smiley:

niranjan92:
I can add topological sort :)

shady831213:
Add some data structure:).
add fibonacci heap, reduced-space-vanEmdeBoas-tree and disjoint-set-tree #31

LockGit:
have some easy go [implementation](https://github.com/LockGit/Go)
https://github.com/LockGit/Go

anand371:
would adding kadanes largest sum contiguous sub array algorithm be a good idea?

PR #12, opened 2017-10-01
adds single linked list under data_structures folder

Riturajcse:
https://github.com/TheAlgorithms/Go/issues/4

Adding single linked list under data_structures folder


dynamitechetan:
I think it is already present.

PR #45, opened 2019-03-06
Find min

BooGluten:
A function that finds the minimum value in a list.

### DIFF
--- a/findmin.go
+++ b/findmin.go
@@ -1,0 +1,22 @@
+package main
+
+import("fmt")
+
+func findmin(in []int) int{
+	smallest := in[0]
+    for i := 0; i < len(in); i ++ {
+        var x = in[i]
+
+        if x <= smallest {
+            smallest = x
+        }
+
+    }
+
+    return smallest
+}
+
+func main() {
+	slice := []int{1,2,3, 0, 5}
+	fmt.Println(findmin(slice))
+}


findmin.go
```diff
@@ -0,0 +1,22 @@
...
+
+import("fmt")
+
+func findmin(in []int) int{
+	smallest := in[0]
+    for i := 0; i < len(in); i ++ {
+        var x = in[i]
+
+        if x <= smallest {
+            smallest = x
```
katkad: Hi. No need to assign new smallest value, if x  == smallest.

x < smallest
will do

findmin.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import("fmt")
+
+func findmin(in []int) int{
+	smallest := in[0]
+    for i := 0; i < len(in); i ++ {
+        var x = in[i]
```
katkad: Hi. You can start from in[1], because in[0] is already in smallest variable. You do not need to compare in[0] and in[0].

Mystigan (changes requested):
Hi,
Please work on the changes suggested by @katkad . Other than that:
- Please move this file into either `math` or `other` folder`. Rename the package accordingly.
- Remove the main function. Add a test file.

tjgurwara99:
Most comments requesting changes are still valid, even now. Please take a look at our new [contribution guidelines](https://github.com/TheAlgorithms/Go/blob/d2f12849df76203d80762ceca5105531d3424384/CONTRIBUTING.md) and make changes accordingly. Currently this PR is not up to standard. If no changes are made in the next two weeks, I will consider this PR to be stale and will close this PR then.

PR #74, opened 2019-10-26
Stack

george-e-shaw-iv:
* Implemented the stack data structure.

### DIFF
--- a/data-structures/stack/stack.go
+++ b/data-structures/stack/stack.go
@@ -1,0 +1,33 @@
+// Package stack implements the Stack data structure, LIFO.
+package stack
+
+// Stack is the type that implements the Stack data structure.
+type Stack struct {
+	items []interface{}
+}
+
+// Push pushes an item onto the stack.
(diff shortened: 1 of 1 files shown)


brayo-pip:
@cclauss this is more ur domain

data-structures/stack/stack.go
```diff
@@ -0,0 +1,33 @@
+// Package stack implements the Stack data structure, LIFO.
+package stack
+
+// Stack is the type that implements the Stack data structure.
+type Stack struct {
```
Mystigan, line 5: Why enclose it in a struct?
You could try something like `type Stack []interface{}`

Mystigan (changes requested):
- Keep the stack type unexported and add an exported func `InitStack` or `New` to initialize an empty stack. This would ensure that a user can import and use this library, at the same time not modify the stack instance directly.
- Add some tests.

tjgurwara99:

PR #100, opened 2019-12-01
Removed main from all sort functions, changed packages to "sorts" package

c4llmeco4ch:
Other than minor changes made to files to deal with a removed main function, this should help to deal with #37 . I have tests I'm currently working on and we should remove some of the files that are duplicates as referenced in #99 , as this leads to build issues.

As an aside, I don't know enough about topological sort to know how to rework the algorithm so I simply commented out main and someone else might need to go back in and condense the program to accommodate.

PR #102, opened 2019-12-31
Pull request regarding benchmarks and tests

c4llmeco4ch:
As a follow-up to PR #100 . There are still some issues I am working out presently.

The main one is topological sort might need to be reworked in how it receives tests or it may need to have its own tests written up specifically for it, different from the normal format. This is due to topological taking in a slice of edges rather than a slice of ints.

Also, while not a huge issue, Selection sort seems unable to handle a sufficiently large sort so I may need to dig in and rewire the large tests to account for selection sort being slow. 

I'll be passing some updates in the near future, but just wanted to get this on the books ASAP.


c4llmeco4ch:
Checking the linter, a lot of the issues seem to stem from general assignment issues. I'll try to tackle those

c4llmeco4ch:
Everything else is a part of topological

c4llmeco4ch:
Gonna sleep on it. The main problem was this sort was written to accommodate multiple possible input possibilities and has multiple supporting functions to facilitate those options. 

Furthermore, because it will require its own special test suite utilizing a graph rather than a list, what boils down to the whole topological_sort.py is more or less causing problems for the linter (see the note on how ```topSort(input []Edge) ([]int, [][]int);``` is unused. 

A bandaid fix is to effectively comment out the whole file while I build a suite for it then deal with the remainder of the tests and put TopSort off for a different PR. Up to you on how you want to proceed.

c4llmeco4ch:
Wrapped the whole thing in a giant comment to get things passing. Happy to tackle this however is deemed most professional, timely, and applicable, but figured something passing seemed to be the highest priority.

PR #115, opened 2020-03-25
Stack Data Structure

3dw1nM0535:
# What is this pull request about?
 - I was able to create a minimal tutorial about stack data structure with test-driven development
 - Methods can be run against their respective tests
 - Also, you can change variables in test cases to understand how stacks work and its implementation


tjgurwara99:
This has already been implemented, please could you provide where this algorithm is different than the one that is already implemented. It would be cool to see different implementations of a data structure but if the implementation is similar to the one already in the repo we may have to close this PR.

PR #123, opened 2020-05-19
Avoids Overflow Errors common in Binary search

brayo-pip:
Adding two large integers then dividing the sum, (a+b)/2 ,might end up creating a sum that can't fit in memory. However this overflow can be avoided by never adding the numbers first instead, adding the difference of the two numbers to the least i.e (a+b)/2 == a+(b-a)/2 == b-(b-a)/2 ... Please comment or ask for further clarification instead of just rejecting the proposal. Thanks


brayo-pip:
Can someone merge this already

PR #140, opened 2020-06-10
add solution of napsack problem with dp

rxaru:
Hi
I added script of solving napsack problem with dp in dynamic-programming directory.


rajneesh44:
@rxaru Please make the following changes.
1. Change ```napsack``` to ```0-1 Knapsack```.
2.  Don't pass a struct, rather than pass two different slices, one of value one of weight.
3. Don't print all the output. Simply print the max profit.

rxaru:
@rajneesh44
I fixed it.
Please review fixed script.

PR #144, opened 2020-06-15
Implemented RSA using math\big library

brayo-pip:
The code is more practical generates 1024 bit keys, unlike my previous implementation which took forever with just a 20 bit key


brayo-pip:
@StepfenShawn do ya thing

PR #151, opened 2020-06-23
implemented dfs, bfs

rxaru:
Hi
I implemented depth first search, breadth first search algorith.

### DIFF
--- a/searches/breadth_first_search.go
+++ b/searches/breadth_first_search.go
@@ -2,7 +2,6 @@ package main
 
 import "fmt"
 
-// getIdx return target node index
 func getIdx(target int, nodes []int) int {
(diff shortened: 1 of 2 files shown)


searches/breadth_first_search.go
```diff
@@ -0,0 +1,76 @@
+package main
+
+import "fmt"
+
+// getIdx return target node index
```
mateuszz0000, line 5: Could you remove unecessary comments? Code should be self documented :)

rxaru:
Sure.
I removed unecessary comments. But, I had no idea whether sample graph comment should be removed or not, so I left it.

mateuszz0000:
@rxaru Thank you for the contribution! :rocket:

PR #153, opened 2020-07-02
Standard formatting

brayo-pip:
I ran "go fmt" on most of the code. Then I figured we can just automate with GitHub actions anyway. Also created a Math folder


cclauss:
I think we need to remove the unused `swap()` function because of
https://github.com/TheAlgorithms/Go/search?q=swap&type=Commits

brayo-pip:
I agree, I'll rewrite that bit later

brayo-pip:
Could u merge this PR btw

PR #158, opened 2020-07-11
added the "=" sign to the condition to first for loop

Ejiro-Edwin:
added the "=" sign to the condition to the first for loop to include the last value which return 18

### DIFF
--- a/strings/naiveStringSearch.go
+++ b/strings/naiveStringSearch.go
@@ -16,7 +16,8 @@ import (
 
 func naivePatternSearch(text string, pattern string) []int {
 	var positions []int
-	for i := 0; i < len(text)-len(pattern); i++ {
+	// added the "=" sign to the condition in for loop i to include the last value which return 18
+	for i := 0; i <= len(text)-len(pattern); i++ {
 		var match bool = true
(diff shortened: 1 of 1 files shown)


strings/naiveStringSearch.go
```diff
@@ -16,7 +16,8 @@ import (
 
 func naivePatternSearch(text string, pattern string) []int {
 	var positions []int
-	for i := 0; i < len(text)-len(pattern); i++ {
+	// added the "=" sign to the condition in for loop i to include the last value which return 18
```
nomisrevol, line 19: We don't need this comment. Correct the algorithm is enough.

Ejiro-Edwin: Correct the algorithm.

nomisrevol: I mean we should remove this comment out of codebase. Because it doesn't help to provide any information related to the rest of the algorithm implementation.

Ejiro-Edwin: Correct the algorithm

Mystigan:
This has been fixed upstream. Hence closing the PR.

PR #161, opened 2020-07-14
Implemented Dijkstra algorithm

dbejanishvili:
* added a structure for saving graphs
* implemented Dijkstra algorithm
* added test file for testing Dijkstra implementation
* added go.mod file so tests can run properly


tjgurwara99:
Could you pull the current repo so that conflicts (which surely exists because we restructured the repo) can be resolved.

PR #162, opened 2020-07-15
Added coin change problem.

himadriganguly:
Added coin change algorithm code in Golang to find the total number of combinations that can be achieved with the given number of coins to get the specified sum.

### DIFF
--- a/dynamic-programming/coinChange.go
+++ b/dynamic-programming/coinChange.go
@@ -1,4 +1,4 @@
-/* Coin Change problem 
+/* Coin Change problem
 1. Finding the total number of combination that can be achived with the given number of coins to get the specified sum.
 */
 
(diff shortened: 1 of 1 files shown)


himadriganguly:
Checked on my local system and it is working exactly fine. golangci-lint check also returns no error.
If anyone can point out the problem of the **All checks have failed** error it will be helpful.

dynamic-programming/coinChange.go
```diff
@@ -0,0 +1,29 @@
+/* Coin Change problem
+1. Finding the total number of combination that can be achieved with the given number of coins to get the specified sum.
```
nomisrevol, line 2: Please provide a more informative description. About the problem, how to solve it, and reference also.

PR #169, opened 2020-08-25
Go : Sorting algorithm updates

Alfex4936:
Test Cases: Added counting sort test, test array now has negative values too. (didn't use crypto/rand as it's not for a product)

Bubble Sort : the name of function was gone

Counting Sort : Made it to work on even with negative values

Quick Sort : using lomuto's partition scheme (with random pivot range from low to high

### DIFF
--- a/sorts/bubble_sort.go
+++ b/sorts/bubble_sort.go
@@ -1,14 +1,16 @@
 //Package sorts a package for demonstrating sorting algorithms in Go
 package sorts
+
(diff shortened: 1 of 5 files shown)


nomisrevol (changes requested):
It is better to separate each improvement into a single commit and PR. Could you follow the lastest commit and create new pull requests, assign me to review. Thank you for your hard work.

PR #170, opened 2020-09-04
Added miller rabin probability test

etimoz:
Added the miller-rabin probability test:

https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test

### DIFF
--- a/other/MillerRabinTest.go
+++ b/other/MillerRabinTest.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"fmt"
+	"math/rand"
+)
+
+
+// Miller–Rabin primality test
+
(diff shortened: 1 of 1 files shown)


tjgurwara99:
There already exists a robust implementation of MR primality test. If you think your approach is different than the current implementation, let us know. In the meantime, please take a look at our new [contribution guidelines](https://github.com/TheAlgorithms/Go/blob/d2f12849df76203d80762ceca5105531d3424384/CONTRIBUTING.md) and make changes accordingly. Currently this PR is not up to standard. If no changes are made in the next two weeks, I will consider this PR to be stale and will close this PR then.

PR #171, opened 2020-09-05
Improved bubble sort

halilylm:
Number of iteration added to reduce number of iterations where possible. More optimized version than previous one


nomisrevol:
Could you take a benchmark to compare your improvement with the original? If you have a benchmark result, please update the Pull Request also. Thank you.

PR #172, opened 2020-09-10
Added roman to integer conversion

Navid2zp:
Added roman to integer conversion and it's file link to DIRECTORY.md

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -6,6 +6,9 @@
   * [Rsacipher](https://github.com/TheAlgorithms/Go/blob/master/ciphers/RSAcipher.go)
   * [Xorcipher](https://github.com/TheAlgorithms/Go/blob/master/ciphers/xorCipher.go)
 
+## Conversions
+  * [Roman To Integer](https://github.com/TheAlgorithms/Go/blob/master/conversions/roman_to_integer.go)
+
 ## Data-Structures
   * Binary-Tree
     * [Binary-Search-Tree](https://github.com/TheAlgorithms/Go/blob/master/data-structures/binary-tree/binary-search-tree.go)
--- a/conversions/conversions_case_test.go
+++ b/conversions/conversions_case_test.go
@@ -1,0 +1,14 @@
+package conversions
+
+type romanToIntegerConversionTest struct {
+	input    string
+	expected int
+	name     string
+}
+
+var romanToIntegerTests = []romanToIntegerConversionTest{
+	{input: "DCCLXXXIX", expected: 789, name: "DCCLXXXIX-789"},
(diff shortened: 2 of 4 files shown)


conversions/roman_to_integer.go
```diff
@@ -0,0 +1,41 @@
...
+			holder += 2
+		} else {
+			total += romans[string(roman[holder])]
+			holder++
+		}
+	}
+	return total
+}
+
+func main() {
```
nomisrevol, line 29: Remove `main` function and create a new test file instead.

DIRECTORY.md
```diff
@@ -6,6 +6,9 @@
   * [Rsacipher](https://github.com/TheAlgorithms/Go/blob/master/ciphers/RSAcipher.go)
   * [Xorcipher](https://github.com/TheAlgorithms/Go/blob/master/ciphers/xorCipher.go)
 
+## Conversions
```
nomisrevol, line 9: You don't need to update this file manually. Please remove this.

nomisrevol (changes requested):

conversions/roman_to_integer.go
Navid2zp: The main function is removed and test files created.

DIRECTORY.md
Navid2zp: Not sure how to remove file changes. will deleting the file or just removing the added lines work?

conversions/conversions_case_test.go
```diff
@@ -0,0 +1,14 @@
+package conversions
+
+type romanToIntegerConversionTest struct {
+	input    string
+	expected int
+	name     string
+}
+
+var romanToIntegerTests = []romanToIntegerConversionTest{
```
nomisrevol, line 9: Move the testcases inside of the test file. We do not need it to separated in a standalone file.

conversions/conversions_test.go
```diff
@@ -0,0 +1,28 @@
...
+		actual := romanToInteger(test.input)
+		if actual != test.expected {
+			t.Errorf("roman to integer test %s failed", test.name,)
+		}
+	}
+}
+
+//END TESTS
+
+//BEGIN BENCHMARKS
```
nomisrevol, line 18: No need to have benchmark for a single implementation, please remove all unnecessary like `BEGIN TESTS` or `END TESTS` also.

nomisrevol (changes requested):

conversions/conversions_case_test.go
Navid2zp: Should I rename the test file to `roman_to_integer_test.go` or you want all the tests in one file like the way it is now?

PR #180, opened 2020-10-01
Create radixSort.go

vedantmamgain:


vedantmamgain:
@nomisrevol You mind taking a look at this.

vedantmamgain:
@Mystigan

sorts/radixSort.go
```diff
@@ -0,0 +1,41 @@
+//Radix sort is an integer sorting algorithm that sorts data with integer keys by grouping the keys by individual digits
+//that share the same significant position and value (place value).
+//Radix sort uses counting sort as a subroutine to sort an array of numbers.
+//Source Wikipedia - https://en.wikipedia.org/wiki/Radix_sort
+package main
+
+func LargetSum(array []int) int {
+	largestNum := 0
```
Mystigan, line 8: Rather than assigning 0, initialize it to the first value of the slice. Because right now, if the passed slice only has negative numbers, the func will return 0 which is incorrect.

Issue #149, opened 2020-06-18
Naming issues in the strings directory


cclauss:
@halafi Can you please make some changes in https://github.com/TheAlgorithms/Go/tree/master/strings ?

1. The directories contain spaces instead of dashes in the directory names.  This forces us to [quote the directory names](https://github.com/TheAlgorithms/Go/blob/master/.github/workflows/golangci-lint.yml#L33) when using command line tools such as linters.
2. File names need work also.  Is `ac.go` a calculation for alternating current or is it controller for an air conditioner?  Is `bom.go` a tool for building a bill of materials or does it generate a byte order mark?  Why does the reader have to guess??  `aho-corasick.go` is a much more self-documenting filename than `ac.go`.
3. Can you help us to get the last two lines of `.github/workflows/golangci-lint.yml` to pass without `|| true`?

piyushs2894:
@cclauss I have fixed the issues 1 and 2 and looking for 3rd as well. Can u please review and merge below PR for the same with "hacktoberfest-accepted" label.
https://github.com/TheAlgorithms/Go/pull/209

piyushs2894:
@cclauss Regarding golandci-lint failure, it is failing as in single_string_matching and multiple_string_matching directory, some methods like main, commandLineInput, and many more more are redeclared. Go does not allow same method name in single package. Is it has to be fixed ?

PR #187, opened 2020-10-02
Clean up code a bit

reverse:
-Tried to fix the main re-declaration issues 
-Rename some directories for better clarity #149 
Tried to fix go-lint throwing err on whitespace in the dir name #149


mckinnonn:
This is a really inadequate fix, but it fixes the issues people are having with constant build fails.

.github/workflows/golangci-lint.yml
```diff
@@ -29,6 +29,6 @@ jobs:
       - run: golangci-lint run --no-config other                              || true
       - run: golangci-lint run --no-config searches                           || true
       - run: golangci-lint run --no-config sorts
-      - run: golangci-lint run --no-config strings
-      - run: golangci-lint run --no-config "strings/multiple string matching" || true
+      - run: golangci-lint run --no-config strings                            || true
```
cclauss, line 32: I object to this change.  We need more tests to be mandatory, not fewer.

PR #193, opened 2020-10-05
Added stooge sort and test cases

JianLoong:
Added stooge sort and test cases.  #4 

More information can be found [here](https://en.wikipedia.org/wiki/Stooge_sort)

Since it is stooge sort, the test cases are expected to fail at the large values (500k values) and the benchmark would also perform poorly.

### DIFF
--- a/sorts/sorts_test.go
+++ b/sorts/sorts_test.go
@@ -95,6 +95,21 @@ func TestShell(t *testing.T) {
 	}
 }
 
+// Stoogesort is bad and performs poorly.
+// This is true on bigger arrays as the test case will time out.
+func TestStooge(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tjgurwara99:
Could you pull our new current master and resolve any conflicts that arise. Also, please take a look at our new [contribution guidelines](https://github.com/TheAlgorithms/Go/blob/d2f12849df76203d80762ceca5105531d3424384/CONTRIBUTING.md) and make changes accordingly if you think you need to make any changes to satisfy our contribution guidelines. If I do not get any correspondence from you in 2 weeks, I will consider this PR to be stale and close it. If you read this after I've closed it, then you are very much free to open it again.

PR #195, opened 2020-10-06
add mean, median

SotirisGP:
added mean, harmonic mean and median


Mystigan (changes requested):
- Remove the main package name.
- Separate them into different files, all under the `math` package.
- Remove main function and add tests. Recommend using table driven tests.
- Also export the function names so that it can be used as a library by anyone.

tjgurwara99:

PR #196, opened 2020-10-06
add min, max, range, distances

SotirisGP:
added min, min distance from 0, max, max distance from 0 and range on a list of numbers

### DIFF
--- a/math/find_min_max.go
+++ b/math/find_min_max.go
@@ -1,0 +1,74 @@
+// Find the minimum and maximum numbers of a list.
+// Find the smallest and largest distance from 0.
+// Find the range.
+
+package main
+
+import (
+	"fmt"
+)
+
+// absolute value, also implemented in "math" as math.Abs
+func abs(a float64) float64 {
+	if a < 0 {
+		return -a
+	}
+	return a
+}
+
+// find the minimum
+func min(numbers []float64) float64 {
+	min := numbers[0]
+	for _, n := range numbers {
+		if n < min {
+			min = n
+		}
+	}
+	return min
+}
+
+//find the smallest distance from 0
+func minDist(numbers []float64) float64 {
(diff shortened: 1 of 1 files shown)


math/find_min_max.go
```diff
@@ -0,0 +1,74 @@
...
+// Find the range.
+
+package main
+
+import (
+	"fmt"
+)
+
+// absolute value, also implemented in "math" as math.Abs
+func abs(a float64) float64 {
```
nomisrevol, line 12: Please reuse the existence function.

math/find_min_max.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}
+	return max
+}
+
+func findRange(numbers []float64) float64 {
+	return max(numbers) - min(numbers)
+}
+
+func main() {
```
nomisrevol, line 67: Remove the `main` function and write the test file instead.

math/find_min_max.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, n := range numbers {
+		if n > max {
+			max = n
+		}
+	}
+	return max
+}
+
+//find the largest distance from 0
+func maxDist(numbers []float64) float64 {
```
nomisrevol, line 53: Should rename to `maxDistToZero`, I think it is more meaningful.

nomisrevol (changes requested):
Overall, you are naming things not really good, trying to see this doc as reference: https://github.com/unknwon/go-code-convention/blob/master/en-US/naming_rules.md#:~:text=Variable%20name%20is%20generally%20using,e.g.%20APIClient%20%2C%20repoID%20%2C%20UserID%20.

Also, find a more meaningful variable name for your variables.

tjgurwara99:

Issue #145, opened 2020-06-18
Need more test files 


rajneesh44:
Need to write test.go files for the following ```unmarked``` directories as we are using non-executable packages. (no ```package main``` at the top)
* [ ] ciphers
* [ ] data-structures
* [ ] dynamic-programming
* [ ] other
* [x] searches
* [x] sorts
* [ ] strings

tjgurwara99:
Closing this to keep track of this issue under a unified issue #272

PR #197, opened 2020-10-06
add tests for Caesar cipher

mcaci:
Hi, I would like to contribute to this project by helping a bit with the tests to add as per issue #145. I needed to move the ceasar code in a separate package to be able to run the newly written tests without build errors that happen when working with the full repository. If this PR is ok with you I can add some more tests after this one.


ciphers/caesar/caesar_test.go
```diff
@@ -0,0 +1,66 @@
...
+		"Encrpyt sentence with key 10",
+		"the quick brown fox jumps over the lazy dog.",
+		13,
+		"gur dhvpx oebja sbk whzcf bire gur ynml qbt.",
+	},
+}
+
+func TestCeasarCipher(t *testing.T) {
+	for _, test := range caesarTestData {
+		actual := caesarCipher(test.input, test.key)
```
Mystigan, line 59: I would suggest to run each iteration in a t.Run() func. This would ensure that even one of the test cases fail, rest of them will still get executed. Currently, if a test case fails, the test terminates and does not execute the remaining cases.

Mystigan (changes requested):
Enclose each iteration in a t.Run() to ensure all the cases get executed in case of failure.

ciphers/caesar/caesar_test.go
mcaci: Yes, I agree that for now it will fail at the first failing test case. I'm changing it so that it executes the rest. Thanks for your feedback!

Mystigan (approved):
LGTM

Mystigan:
Merging changes as the lint issue is unrelated and has been fixed upstream.
@mcaci , the PR format is fine. You may refer to the format followed in #219. I feel it is a bit more easy on the eyes.

PR #205, opened 2020-10-08
Added Pythagoras Theorem

localleon:
Calculates the distance between to vectors/points in space with the Pythagoras Theorem.

### DIFF
--- a/math/pythagoras/pythagoras.go
+++ b/math/pythagoras/pythagoras.go
@@ -1,4 +1,4 @@
-package main
+package pythagoras
 
 import (
 	"fmt"
@@ -19,8 +19,4 @@ func Distance(a, b vector) float64 {
 	res := float64(math.Pow(abs, 2.0) + math.Pow(float64(b.y-a.y), 2.0) + math.Pow(float64(b.z-a.z), 2))
 	fmt.Println(abs, res)
 	return math.Sqrt(res)
-}
-
-func main() {
-
 }
--- a/math/pythagoras/pythagoras.go
+++ b/math/pythagoras/pythagoras.go
@@ -6,15 +6,15 @@ import (
 	"math"
 )
 
-// vector defines a tuple with 3 values in 3d-space
-type vector struct {
+//Vector defines a tuple with 3 values in 3d-space
+type Vector struct {
 	x int
 	y int
 	z int
 }
 
 //Distance calculates the distance between to vectors with the   Pythagoras theorem
-func Distance(a, b vector) float64 {
+func Distance(a, b Vector) float64 {
 	abs := float64(math.Abs(float64(b.x - a.x)))
 	res := float64(math.Pow(abs, 2.0) + math.Pow(float64(b.y-a.y), 2.0) + math.Pow(float64(b.z-a.z), 2))
 	fmt.Println(abs, res)
--- a/math/pythagoras/pythagoras.go
+++ b/math/pythagoras/pythagoras.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+
+	"math"
+)
+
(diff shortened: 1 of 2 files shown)


vedantmamgain:
Remove the main function.

localleon:
> Remove the main function.
Done. Also renamed the package to make it more universal

math/pythagoras/pythagoras.go
```diff
@@ -0,0 +1,22 @@
...
+	x int
+	y int
+	z int
+}
+
+//Distance calculates the distance between to vectors with the   Pythagoras theorem
+func Distance(a, b Vector) float64 {
+	abs := float64(math.Abs(float64(b.x - a.x)))
+	res := float64(math.Pow(abs, 2.0) + math.Pow(float64(b.y-a.y), 2.0) + math.Pow(float64(b.z-a.z), 2))
+	fmt.Println(abs, res)
```
Mystigan, line 20: Please remove this print statement.

Mystigan (changes requested):
- Please remove the print statement.
- Let's keep the power factor consistent (Line 19: currently it's 2.0 for x and y components and 2 for z).
- Please add some more test cases and consider using table driven tests.( Refer: https://github.com/golang/go/wiki/TableDrivenTests )

math/pythagoras/pythagoras.go
localleon: Sure thing.

localleon:
>     * Please remove the print statement.
> 
>     * Let's keep the power factor consistent (Line 19: currently it's 2.0 for x and y components and 2 for z).
> 
>     * Please add some more test cases and consider using table driven tests.( Refer: https://github.com/golang/go/wiki/TableDrivenTests )

Everything should be fixed now. I needed to round the results of the Distance Function for the tests because i wasn't able to find nice number for round results and i also wanted to test edge cases. Writing down the whole 64bit of the float number would be kinda difficult to do in a testfile

math/pythagoras/pythagoras.go
```diff
@@ -0,0 +1,19 @@
...
+//Vector defines a tuple with 3 values in 3d-space
+type Vector struct {
+	x float64
+	y float64
+	z float64
+}
+
+//Distance calculates the distance between to vectors with the   Pythagoras theorem
+func Distance(a, b Vector) float64 {
+	abs := float64(math.Abs(b.x - a.x))
```
Mystigan, line 16: Why are calculating absolute value for the x component. We are guaranteed to get a positive number when we square it.

math/pythagoras/pythagoras_test.go
```diff
@@ -0,0 +1,31 @@
...
+	{Vector{2, -1, 7}, Vector{1, -3, 5}, 3.0},
+	{Vector{4, 10, 9}, Vector{4, 3, 5}, 8.06},
+	{Vector{8, 5, 5}, Vector{1, 1, 12}, 10.67},
+	{Vector{1, 1, 1}, Vector{2, 2, 2}, 1.73},
+}
+
+//TestDistance tests the Function Distance with 2 vectors
+func TestDistance(t *testing.T) {
+	for _, tt := range distanceTest {
+		res := Distance(tt.v1, tt.v2)           // Calculate result for
```
Mystigan, line 23: Please enclose each iteration in a t.Run() func. This would ensure that even if one of the test cases fail, the others will still be executed. The func will require a name for the test case which you can add as a part of distanceTest slice. You may use [this test file](math/primecheck/prime_check_test.go) for  reference.

Mystigan (changes requested):
Please work on the suggested changes and we should be ready to go.

Mystigan (approved):
LGTM. Thanks for your contribution!

PR #206, opened 2020-10-08
Calculating Surface Area of Sphere and Cube

localleon:
Calculates the Surface Area of some geometric shapes.

### DIFF
--- a/math/area/area.go
+++ b/math/area/area.go
@@ -1,0 +1,24 @@
+package main
+
+import (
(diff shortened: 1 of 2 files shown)


vedantmamgain:
The addition seems too specific. I would suggest, you implement other topics of Geometry. 
Closing this for now.

localleon:
What kind of geometric topics are you interested in? More simple shapes?

PR #209, opened 2020-10-11
Changed directory naming of string sub directories

piyushs2894:


cclauss:
I disagree with this change... String algorithms can be conversion, formatting, etc. not just matching.

PR #216, opened 2020-10-11
tests for rot13 algorithm

mcaci:
Hi, continuing my work on issue #145 and using the structure used in PR #197. With this PR i'm adding some tests for the rot13 algorithm. As in my recent PRs the tests work on the created package but not when running them on the whole project.

### DIFF
--- a/ciphers/rot13/rot13.go
+++ b/ciphers/rot13/rot13.go
@@ -1,8 +1,7 @@
-package main
+package rot13
 
 import (
 	"bytes"
-	"fmt"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


Mystigan (changes requested):
- Please rename package appropriately for both files.
- Remove main function from Rot13.go.

- This one is more like an open ended suggestion. The first parameter of t.Run() takes the description of the test case. So, in case of failure, it will be displayed with FAIL status. So, the additional t.Logf() entry seems redundant.

mcaci:
Hi @Mystigan, I did the changes you asked on the review. Since the main function was doing an example of encrypting and decrypting I eventually created two tests, one for encrypting and one for decrypting so that this behavior is not lost while testing the behavior. About the last suggestion, I agree that the description of the test appears twice in the output so it makes sense to me to remove it. Cheers!

Mystigan (approved):
Changes LGTM. 
Merging the PR. Thanks for your contribution!

mcaci:
Glad to be helping :)

PR #220, opened 2020-10-12
add rot13 unit test

enrinal:
add Rot13 unit test.


Mystigan:
Closing this as #216  is already open for the same algorithm(and it was opened first).
@enrinal , moving forward, when opening a new PR, please take a look at the other open PRs to avoid such clashes. It would also be helpful when PRs reference an issue that is fixes(fully or partially) for better tracking.
Do not be discouraged. There are many other items that you can work on.
Thanks for your contribution. 👍

PR #227, opened 2020-10-16
Add tests on KMP algorithm on single-string-matching

WilliamChang80:
I've added some test case on kmp algorithm on single-string-matching and also refactored some duplicate code so it's easier to test and improved later. This PR is also part of issue #145


vedantmamgain:
Squash all the commits.

WilliamChang80:
Done, you can check it again

vedantmamgain:
LGTM

PR #231, opened 2020-10-17
added test for radix sort

addy1997:
The sorts.go file doesn't have a test for radix sort. I have added it. please can anyone run it and verify.

### DIFF
--- a/sorts/sorts_test.go
+++ b/sorts/sorts_test.go
@@ -106,6 +106,19 @@ func TestShellSort(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cclauss:
There are several contenders https://github.com/TheAlgorithms/Go/pulls?q=is%3Apr+is%3Aopen+radix so the maintainers will need to select which one they like most.  We will certainly not break the build by merging an implementation with failing tests.

vedantmamgain:
Code has been copied from [here](https://www.cnblogs.com/CherryTab/p/12825780.html)

PR #233, opened 2020-10-17
cycle sort impl

Adwait1997:
cycle sort implemented [#4 ]


Mystigan (commented):
Code seems to be copied from [here](https://ispycode.com/GO/Sorting/Sorting-Algorithms/Cycle-Sort). It is clearly authored by a different person.
Hence, closing the PR.

PR #237, opened 2020-10-19
data-structures/binary-tree: Move main into tests as per #145

willemvds:
This moves the ```main``` package and func into a test file so it can be executed without changing the source file and requiring an executable as per #145.

In order to fully utilize the testing mechanisms the results of the operations on the trees need to be asserted like in the depth test. Doing that requires changes to the operations themselves which falls outside the scope of this PR but is easily enabled in future by these test files.

As part of these changes the binary search tree was moved into a separate folder so that the package names no longer clash and lastly the package name was changed from ```binarySearchTree``` to ```binarysearchtree``` to confirm more closely with golang package naming standards.

### DIFF
--- a/data-structures/binary-search-tree/binary-search-tree.go
+++ b/data-structures/binary-search-tree/binary-search-tree.go
@@ -1,9 +1,7 @@
 // Binary search tree
 // https://en.wikipedia.org/wiki/Binary_search_tree
 
-package binarySearchTree
-
-// package main
+package binarysearchtree
 
 import "fmt"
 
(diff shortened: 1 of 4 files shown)


tjgurwara99:
This algorithm has been restructured completely. Could you take a look and see whether you can write a test for the new implementation. Also, Please take a look at our new [contribution guidelines](https://github.com/TheAlgorithms/Go/blob/d2f12849df76203d80762ceca5105531d3424384/CONTRIBUTING.md) and make changes accordingly. Currently this PR seems to be okay but personally I don't think it is up to standard. If no changes are made in the next two weeks, I will consider this PR to be stale and will close this PR then. If you see this message after the PR has been closed you are free to open this PR again. Thanks.

PR #241, opened 2020-10-20
Implemented interpolationSearch in Go

Supuna-Kaumada:
Implemented interpolationSearch in Go


vedantmamgain:
This has been plagiarized from [here](https://www.golangprograms.com/golang-program-for-implementation-of-interpolation-search.html)

PR #245, opened 2020-10-25
Update iterBinarySearch

Crazy3lf:
I added test cases for binary and linear searches and I notice the `iterBinarySearch` function does not passes the test cases. I make some minor changes to the algorithm.

PR #247, opened 2020-10-25
Update iterBinarySearch and add test cases

Crazy3lf:
This is an update to my previous PR #245 .
I am not sure what name the test cases should be so I just use the same name for those cases.


cclauss:
Please resolve the conflict.

cclauss:
Please resolve conflicting files.

cclauss:
@tjgurwara99 Can I please get your review of this PR?

tjgurwara99:
@cclauss The code seems to be fine, do you think there should be a benchmark test for these algorithms?

tjgurwara99 (changes requested):
The code is fine but I think there should also be a benchmark test in the test file for people to compare the efficiency of these algorithms.

searches/binarysearch.go
```diff
@@ -5,7 +5,7 @@ func BinarySearch(array []int, target int, lowIndex int, highIndex int) int {
 	if highIndex < lowIndex || len(array) == 0 {
 		return -1
```
tjgurwara99, line 6: Consider adding a returning a proper error by changing the return type to include the answer and an error, for example:

```
func BinarySearch(array []int, target int, lowIndex int, highIndex int) (int, error) {
   ...
}

```

and return something like this

```suggestion
return -1, errors.New("The value not found in the array")
```

searches/search_test.go
```diff
@@ -25,7 +32,7 @@ func TestBinarySearch(t *testing.T) {
 	for _, test := range searchTests {
 		actual := BinarySearch(test.data, test.key, 0, len(test.data)-1)
 		if actual != test.expected {
-			t.Errorf("test %s failed", test.name)
+			t.Errorf("test '%s' failed: input array '%v' with key '%d', expected '%d', get '%d'", test.name, test.data, test.key, test.expected, actual)
```
tjgurwara99, line 35: If you add the error as a return type then it would be worth checking for the expected errors.

searches/binarysearch.go
Crazy3lf: I think -1 is enough to imply not found. Is it not?

cclauss: I agree with @tjgurwara99 on this one.  Delivering an error rather than implying an error is the correct idem in Go.

Crazy3lf:
> 
> 
> The code is fine but I think there should also be a benchmark test in the test file for people to compare the efficiency of these algorithms.

I am rather inexperienced in writing benchmark or testing for efficiency of search algorithms so I am not sure I am capable for this😅.
Do you suggest we should have something like [this](https://github.com/golang/go/blob/618b66e16d5adedbeccb7e7ed35b26e439a9f86f/src/sort/search_test.go#L73) from Go [sort](https://golang.org/pkg/sort/) package?

searches/search_test.go
```diff
@@ -1,49 +1,65 @@
...
-			t.Errorf("test %s failed", test.name)
+		actualValue, actualError := BinarySearch(test.data, test.key, 0, len(test.data)-1)
+		if actualValue != test.expectedValue {
+			t.Errorf("test '%s' failed: input array '%v' with key '%d', expected '%d', get '%d'", test.name, test.data, test.key, test.expectedValue, actualValue)
+		}
+		if actualError != test.expectedError {
+			t.Errorf("test '%s' failed: input array '%v' with key '%d', expected error '%s', get error '%s'", test.name, test.data, test.key, test.expectedError, actualError)
 		}
 	}
 }
```
tjgurwara99, line 45: it should be something along the lines of

```suggestion

func BenchmarkBinarySearch(b *testing.B) {
	var testCase []int
	for i := 0; i < b.N; i++ {
		testCase = append(testCase, i)
	} // You can take this completely outside with a reasonably large size say 100 - that way it is easier to take this and use it inside other benchmarks too for comparison of efficiency and speed.
	b.ResetTimer() // this is important because the above loop is expensive
	for i := 0; i < b.N; i++ {
		_, _ := BinarySearch(testCase, 10, 0, len(testCase) - 1)
	}
}

```

In the comment I meant this

```suggestion

func generateBenchmarkTestCase() []int {
	var testCase []int
	for i := 0; i < 100; i++ {
		testCase = append(testCase, i)
	}
	return testCase
}

func BenchmarkBinarySearch(b *testing.B) {
	testCase := generateBenchmarkTestCase()
	b.ResetTimer() // this is important because the generateBenchmarkTestCase() is expensive
	for i := 0; i < b.N; i++ {
		_, _ := BinarySearch(testCase, 10, 0, len(testCase) - 1)
	}
}

```

Make sure to run `gofmt -s` on the changed files for any formatting issue.

You should read up on the [testing](https://golang.org/pkg/testing/) documentation. It's quite good and very (very) useful.

In the go standard sort library that you linked, you should look at [this](https://github.com/golang/go/blob/618b66e16d5adedbeccb7e7ed35b26e439a9f86f/src/sort/search_test.go#L143) line, this is where they test the actually benchmark the sorting algorithms on in-built types.

Note that the efficiency testing that they are doing is a `test` not a `benchmark` thus it is testing something not benchmarking and you can't use cpu profilers on them (easily).

tjgurwara99 (commented):
I think the changes are fine now. Once you have added the benchmark, I will approve the changes after testing.

searches/search_test.go
Crazy3lf: Thanks for the guidance!

raklaptudirm:
This pr currently has merge conflicts. Please fix them.

raklaptudirm:
This pr currently has merge conflicts. Please fix them.

searches/search_test.go
```diff
@@ -1,49 +1,89 @@
...
+	{[]int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, 2, 1, nil, "Sanity"},
+	{[]int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, 1, 0, nil, "Sanity"},
 	//Absent
-	{[]int{1, 4, 5, 6, 7, 10}, -25, -1, "Absent"},
-	{[]int{1, 4, 5, 6, 7, 10}, 25, -1, "Absent"},
+	{[]int{1, 4, 5, 6, 7, 10}, -25, -1, errNotFound, "Absent"},
+	{[]int{1, 4, 5, 6, 7, 10}, 25, -1, errNotFound, "Absent"},
 	//Empty slice
-	{[]int{}, 2, -1, "Empty"},
+	{[]int{}, 2, -1, errNotFound, "Empty"},
```
tjgurwara99, line 31: Could you move the test cases inside the test to make sure that there is no conflict with new searching algorithms that are created in this package. Also, do a git pull from our master branch and resolve conflicts. We have changed the structure of the repository which would create conflicts for this PR. For example, you can see that the package name has been changed to `search` instead of `searches` and same applies for the directory. Let me know when you are ready for this PR to be reviewed. If you do not interact with this PR for two weeks then I would consider this PR to be stale and close it. You will be able to reopen it in the case that you did not see this message.

tjgurwara99 (changes requested):

searches/search_test.go
Crazy3lf: The new structure have a single `testcases.go` file share across tests for different search functions. Since currently `Binary` and `BinaryIterative` function return error for when value not found but the other search functions did not, I have opted to ignore error on the test for now. Should I add a error return value for other searches?

search/binary_test.go
```diff
@@ -1,23 +1,45 @@
...
-	"testing"
-)
+import "testing"
 
 func TestBinarySearch(t *testing.T) {
 	for _, test := range searchTests {
-		actual := Binary(test.data, test.key, 0, len(test.data)-1)
-		if actual != test.expected {
-			t.Errorf("test %s failed", test.name)
+		actualValue, _ := Binary(test.data, test.key, 0, len(test.data)-1)
```
tjgurwara99, line 7: I think you're currently ignoring the errors raised because of the other implementations. could you instead create a new test case locally inside this test function to check the expected errors as well. Going forward it would be good to cover everything implemented. If you can change the other implementations, I'd be happy to review but if not, I suggest that you create a local struct for the test cases and test it on them to make sure everything works as expected 😄

search/binary_test.go
```diff
@@ -1,23 +1,45 @@
...
 		}
 	}
 }
 
 func TestIterBinarySearch(t *testing.T) {
 	for _, test := range searchTests {
-		actual := BinaryIterative(test.data, test.key, 0, len(test.data)-1)
-		if actual != test.expected {
-			t.Errorf("test %s failed", test.name)
+		actualValue, _ := BinaryIterative(test.data, test.key, 0, len(test.data)-1)
```
tjgurwara99, line 16: same as above ^

tjgurwara99 (changes requested):
Consider the changes that I have requested. Overall, I like the changes you've made. Thanks for contributing 😄

siriak (dismissed):
LGTM, can merge once @tjgurwara99 approves too :)

raklaptudirm (dismissed):

searches/search_test.go
raklaptudirm: @tjgurwara99

search/binary_test.go
Crazy3lf: I think I will open a new PR to implement error and benchmark test on other search functions, is that ok?

siriak (approved):

tjgurwara99 (approved):
LGTM! Thanks for contributing!

PR #248, opened 2020-10-25
Added interpolation search and updated README.md

Kingarturs:
Hi there! I added interpolation search algorithm and it's description in the README.md

Hope it helps 🚀

### DIFF
--- a/searches/interpolation_search.go
+++ b/searches/interpolation_search.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"fmt"
+)
+
+func interpolationSearch(arr []int, x int) int {
+	var first = 0
+	var last = len(arr) - 1
+
+	for first <= last && x >= arr[first] && x <= arr[last] {
+		if first == last {
+			if arr[first] == x {
+				return first
+			}
+			return -1
+		}
+
+		tmp := first + int(((last-first)/(arr[last]-arr[first]))*(x-arr[first]))
+
+		if arr[tmp] == x {
+			return tmp
+		}
+
+		if arr[tmp] < x {
+			first = tmp + 1
+		} else {
+			last = tmp - 1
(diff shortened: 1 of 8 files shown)


searches/interpolation_search.go
```diff
@@ -0,0 +1,47 @@
+package main
```
nomisrevol, line 1: It should belong to `package searches`.

searches/interpolation_search.go
```diff
@@ -0,0 +1,47 @@
...
+			first = tmp + 1
+		} else {
+			last = tmp - 1
+		}
+	}
+
+	return -1
+}
+
+func main() {
```
nomisrevol, line 35: Remove `main` function, create test file instead.

nomisrevol (changes requested):
Change file name to `interpolationsearch.go` and add test to `search_test`. You can add more tests, it's very helpful and you're welcome.

searches/interpolation_search.go
Kingarturs: I noticed that every file on the searches folder is in the `main` package, should i leave mine there too or make it a separate package?

Kingarturs: It could imply to make a separate folder of the same name inside searches (as far as i know).
I can:
1. Change package of the go files inside the folder to searches
2. Leave my function in the main package the avoid import problems

nomisrevol: It's better to change it to `package searches` and the other search implementation should be changed to `package searches` too.

tjgurwara99:
Please resolve the conflicts. If you do not interact with this PR for two weeks, then I would consider this PR to be stale and close it. If for some reason you had not seen this message in the next two weeks, you are free to reopen this PR.

PR #249, opened 2020-10-27
Refactor repo with filename convention and folder convention

nomisrevol:


cclauss:
I am not a fan of this approach.  It is difficult to read and makes humans work hard so that computers don't have to.

Issue #228, opened 2020-10-16
Implement Fibonacci sequence in dynamic programming


Gituser143:
**Description**
Currently the dynamic programming examples, no implementation of Fibonacci sequence has been made. This is by far the most common and beginner friendly example for dynamic programming. I would like to implement the same.

For new implementation:
- Fibonacci sequence using dynamic programming.

abrahamfrancis:
Can I take this up?

IanvsPoplicola:
I've opened a PR with a pretty basic implementation of this 🙂

PR #252, opened 2020-11-01
Add fibonacci algorithm by dynamic programming

IanvsPoplicola:
Fixes https://github.com/TheAlgorithms/Go/issues/228

Add the basic dynamic programming algorithm for calculating Nth Fibonacci number from https://www.geeksforgeeks.org/program-for-nth-fibonacci-number/

### DIFF
--- a/dynamicprogramming/fibonacc_test.go
+++ b/dynamicprogramming/fibonacc_test.go
@@ -1,0 +1,39 @@
+package dynamicprogramming
+
+import "testing"
+
+func Test_NthFibonacci(t *testing.T) {
+	// source: http://www.maths.surrey.ac.uk/hosted-sites/R.Knott/Fibonacci/fibtable.html
+	var fibonacciNumbers = []struct {
+		nth       uint
+		fibonacci uint
+	}{
+		{0, 0},
+		{1, 1},
+		{2, 1},
+		{3, 2},
+		{4, 3},
+		{5, 5},
+		{6, 8},
+		{7, 13},
+		{8, 21},
+		{9, 34},
+		{10, 55},
+		{20, 6765},
+		{30, 832040},
+		{40, 102334155},
+		{50, 12586269025},
+		{60, 1548008755920},
+		{70, 190392490709135},
+		{80, 23416728348467685},
+		{90, 2880067194370816120},
+	}
+	t.Run("test", func(t *testing.T) {
+		for i := range fibonacciNumbers {
+			result := NthFibonacci(fibonacciNumbers[i].nth)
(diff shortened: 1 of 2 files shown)


dynamicprogramming/fibonacci.go
```diff
@@ -0,0 +1,21 @@
+package dynamicprogramming
+
+// Nth fibonacci Number
+// https://www.geeksforgeeks.org/program-for-nth-fibonacci-number/
+
+func fib(n int) int {
+	/* Declare an array to store Fibonacci numbers. */
+	f := make([]int, n+2) // 1 extra to handle case, n = 0
```
Mystigan, line 8: - This extra space is useful when the requirement is to return the first n fibonacci numbers.
- But since this program only returns the nth fibonacci number, we don't have to save all the previous numbers. We only need to keep track of the two previous numbers in the series.

dynamicprogramming/fibonacci.go
```diff
@@ -0,0 +1,21 @@
+package dynamicprogramming
+
+// Nth fibonacci Number
+// https://www.geeksforgeeks.org/program-for-nth-fibonacci-number/
+
+func fib(n int) int {
```
Mystigan, line 6: - Export the function and give it a more meaningful name.

Mystigan (changes requested):
- Add some tests.
- It is not advised to copy or port solutions from elsewhere. But considering your intentions were good, please ensure it doesn't happen in the future.
- Also, for the changes suggested, take a crack at it yourself.
@cclauss, @vedantmamgain, what do you think?

IanvsPoplicola:
Hi @Mystigan, thank you for the review! I've updated the implementation and added tests as per suggested.

dynamicprogramming/fibonacc_test.go
```diff
@@ -0,0 +1,39 @@
...
+		{20, 6765},
+		{30, 832040},
+		{40, 102334155},
+		{50, 12586269025},
+		{60, 1548008755920},
+		{70, 190392490709135},
+		{80, 23416728348467685},
+		{90, 2880067194370816120},
+	}
+	t.Run("test", func(t *testing.T) {
```
Mystigan, line 31: Please put the Run function inside the loop. This will ensure each case is run independently and one case failing will not block the execution of other cases.

Mystigan (changes requested):
Please make the suggested changes. Rest look good to me.

IanvsPoplicola:
Done!

Mystigan (approved):
LGTM.
Thanks for your contribution.

PR #255, opened 2020-11-06
Add moser-de bruijin sequence

cp-sumi-k:


math/moserdebruijnsequence/sequence.go
```diff
@@ -0,0 +1,37 @@
+// The Moser-de Bruijn sequence is the sequence obtained by
+// adding up the distinct powers of the number 4 (For example 1, 4, 16, 64, etc).
```
Navid2zp, line 2: You should probably add the Wikipedia URL:
https://en.wikipedia.org/wiki/Moser%E2%80%93de_Bruijn_sequence

math/moserdebruijnsequence/sequence.go
```diff
@@ -0,0 +1,37 @@
...
+
+package main
+
+import "fmt"
+
+func main() {
```
Navid2zp, line 8: Also, you should remove the main function.
You can change your `getSequenceNumber` function to `MoserDeBruijnSequence` instead and return the final result.

Navid2zp (changes requested):
You need to add the tests too. Check other algorithms for example.

cp-sumi-k:
Changes Done @Navid2zp. Please review it.

Navid2zp:
> Changes Done @Navid2zp. Please review it.

I'm not a maintainer but looks good to me. Wait for one of the maintainers to check.
Just wanted to point out the simple problems so you can fix it before one of the maintainers check the PR since it usually takes a while.

Navid2zp (dismissed):

iamrajiv (dismissed):

math/moserdebruijnsequence/sequence_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+var testCases = []struct {
+	name     string
+	inputNum int
+	expected []int
+}{
+	{"first 0 terms", 0, []int{}},
+	{"first 5 terms", 5, []int{0, 1, 4, 5, 16}},
+	{"first 10 terms", 10, []int{0, 1, 4, 5, 16, 17, 20, 21, 64, 65}},
```
siriak, line 15: No need to test the first 0, 5, and 10 terms if you're testing the first 15 terms too

siriak (approved):
LGTM except for 1 comment below. Please bring the branch up to date so that I could merge it in.

siriak (approved):

PR #256, opened 2020-11-14
Fix sort testcase input sorted unintentionally

vzvu3k6k:
As `getSortedVersion` and other sorting functions modify a given slice, testcase inputs are unintentionally sorted beforehand.

To avoid that, passes a cloned slices to these functions.

### DIFF
--- a/sorts/sorts_testcases.go
+++ b/sorts/sorts_testcases.go
@@ -69,6 +69,8 @@ func makeRandomSignedSlice(size int) []int {
 	return vals
 }
 
+// getSortedVersion Takes a slice of ints and returns a slice as a sorted (in 
+// ascending order) copy of the slice passed as the argument.
 func getSortedVersion(a []int) []int {
 	b := cloneIntSlice(a)
 	sort.Slice(b, func(i, j int) bool { return b[i] < b[j] })
--- a/sorts/sorts_testcases.go
+++ b/sorts/sorts_testcases.go
@@ -75,6 +75,8 @@ func getSortedVersion(a []int) []int {
 	return b
 }
 
+// cloneIntSlice returns a copy of a given slice of int.
+// Useful to protect a slice from in-place sorting functions.
 func cloneIntSlice(src []int) []int {
 	vals := make([]int, len(src))
 	copy(vals, src)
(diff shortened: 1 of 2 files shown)


tjgurwara99 (approved):
All seems good to me. Maybe add a comment on top of `getSortedVersion` and `cloneIntSlice` to make it easier to understand what is happening, even though it's mostly self explanatory.

vzvu3k6k:
@tjgurwara99
Thanks for review. I've added a comment for `cloneIntSlice` at 651bbbc, but can't think of a good comment for `getSortedVersion`. If you have any good ideas, please let me know.

Now that `go test` runs on CI thanks to #266, I've reduced the slice length of testcases (at 0bbc1db) to avoid timeout on GitHub Actions (happened on https://github.com/TheAlgorithms/Go/runs/1995000745). It would be great if you could review this change as well.

sorts/sorts_testcases.go
```diff
@@ -70,6 +70,15 @@ func makeRandomSignedSlice(size int) []int {
 }
 
 func getSortedVersion(a []int) []int {
```
tjgurwara99, line 72:
```suggestion
// getSortedVersion Takes a slice of ints and returns it slice as a sorted (in 
// ascending order) copy of the slice passed as an argument.
func getSortedVersion(a []int) []int {
```

or something along those lines. This function is essentially taking in a slice and sorting it with the in-built sorting algorithm. The in-built algorithms sorts it according to the second argument which is passed as a comparison `lambda` (not sure if its called lambda in Golang - mathematically its a lambda). So we can work with the sorted version of the unsorted random slices that we generated for testing to compare the answer of the algorithms with a properly sorted slice.

tjgurwara99 (changes requested):
Good work catching that unintentional sorting!

sorts/sorts_testcases.go
vzvu3k6k: Thanks for suggestion! Just committed.

tjgurwara99:
This issue has been resolved recently when I was restructuring the repo. You can close this PR now. Thank you.

PR #261, opened 2021-01-17
Add lacked tests for caesar cipher

task4233:
- add lacked processings
     - adapts not only lowercase but also uppercase
     - implements decrypt function
  - fix typo
    - package name
    - some words in test
  - add lacked test cases
    - coverage(64.7%=>100.0%) 

refers: #13 #197 #260

### DIFF
--- a/ciphers/caesar/CaesarCipher.go
+++ b/ciphers/caesar/CaesarCipher.go
@@ -1,41 +1,38 @@
-package ceaser
- 
-import (
-	"bytes"
-	"flag"
-	"fmt"
-	"strings"
-)
- 
-func main() {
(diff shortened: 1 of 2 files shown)


vedantmamgain:
LGTM. Can you squash the commits I'll merge .

task4233:
@vedantmamgain 
Thank you for your reviewing.
I'm afraid that I would appreciate if you could press "Squash and merge" button on GitHub.

Although I can squash commits on my local, I have to do forced push then. I don't think it's good idea.
In addition, pressing "Squash and merge" button on GitHub has same meaning which squashes commits on my local.

For the reason above, could you push "Squash and merge" button on GitHub?

vedantmamgain:
Sure, no problem.

Issue #198, opened 2020-10-07
Proposal: Testing revamp


willdavsmith:
I was looking over this repository and I noticed that lots of the files have commented out testing methods, however some have no proper tests. It seems to me that there's no centralized testing for the repository, and I think that it would be beneficial. 

Proposal:
* Centralized testing (go test) for entire project at once
* Integration with CI (Github actions - test and lint on each new PR)
* Code cleanup and migration of tests in main() methods to unit tests

Thoughts? I would be happy to work on this if others think it would be useful!

In the file [golangci-lint.yml](../blob/master/.github/workflows/golangci-lint.yml)...
* [x] Remove `|| true` from ciphers -- Fixed in #222
* [x] Remove `|| true` from data-structures/binary-tree
* [x] Remove `|| true` from data-structures/dynamic-array
* [x] Remove `|| true` from data-structures/hash-map -- Fixed in #222
* [x] Remove `|| true` from data-structures/linked-list
* [x] Remove `|| true` from data-structures/trie
* [x] Remove `|| true` from dynamic-programming
* [x] Remove `|| true` from other
* [x] Remove `|| true` from searches
* [x] Remove `|| true` from "strings/multiple string matching"
* [x] Remove `|| true` from math/sieve

vedantmamgain:
Alongside Centralized testing, I believe a README containing benchmark of all the sorts and searches must be added as well.

enrinal:
I think it's good idea for centralized testing.

cclauss:
Nice initiative -- I am highly supportive.  Let's get the continuous integration tests passing again.  Merging PRs when tests are failing is just dangerous.  @Mystigan, can you please help to make this happen?

Mystigan:
@cclauss , I agree with the points made.
- We can begin our cleanup with the files that are failing the CI test(from the strings folder AFAIK) and proceed from there.
- About centralized testing and adding it as part of CI, I would need some help with this as I haven't done this before.
- Also, to ensure it doesn't  turn ugly again, we should setup a set of guidelines for contributors so that they have an idea about the structure expected (for code as well as issues and PRs). To this end, so far, I have added some basic templates for issues and PRs. Please go through them and suggest changes, if any.

Mystigan:
So I have fixed the issues in the two files in strings folder which were causing the ci lint to fail. But for existing PRs, how do we update them? Will running `git pull upstream master` and then a `git push origin master` work? Or is there any other way to get this done?

[...]

PR #266, opened 2021-02-17
Fix: Major restructuring for fixing the linting workflow errors and module style code writing.

tjgurwara99:
Apologies for this big PR but I thought I should try and fix the automated issues that this repository faces.

I have refactored quite a lot of code in order for this repository to work the way I think the Go implementation is meant to be. For example, you can work with any function in this repository as long as you do `go get GitHub.com/tjgurwara99/Go`. The go mod file should be changed so that people can use `go get` command with TheAlgorithm handle. This particular feature helps avoid the presence of repeated code - however, there are still quite a lot of functions like max, min etc being redeclared locally in each package. Maintainer's should consider adding a proper contribution guidelines file which may explain how people should go about making the code more modular.

As I said before I have restructured quite a lot of things in this repository and there are a lot of test files missing. Once I get time, I will try and write some more tests for the repository. But so far with all the tests that exists, there are no issues except for the `strings/sbom` submodule test and `cipher/rsa` submodule - I believe that the implementations are incorrect, but I didn't have time to check properly. All the automated linting tests with golangci-lint are running without any issues (thus we don't need any || true) in the workflow files.

Also note that there are some files that were completely written as a main module which makes `go test` go crazy so I have decided to comment the code out for now and as soon as I get some time, I will try and convert them into functions and write tests so that the whole repository management can be completely automated. There are quick fixes which can sometime cause issues long term but I didn't want to cause any issues for the repository so I refrained from doing that.

Hope it helps the maintainers 😄

I also wanted to ask the maintainers if the commented out code can be removed (the files where there are some tests and the main function that is commented out is not required)?


ciphers/rot13/rot13.go
```diff
@@ -3,10 +3,10 @@
 package rot13
 
 import (
-	"TheAlgorithms/Go/ciphers/caesar"
+	"github.com/tjgurwara99/Go/ciphers/caesar"
```
tjgurwara99, line 6:
```suggestion
	"github.com/TheAlgorithm/Go/ciphers/caesar"
```
```suggestion
	"github.com/tjgurwara99/Go/ciphers/caesar"
```

ciphers/rsa/RSAcipher.go
```diff
@@ -130,44 +135,63 @@ func toASCII(slice []rune) []int {
...
+	var b []int
+	for _, v := range a {
+		temp, _ := strconv.Atoi(v)
+		b = append(b, temp)
 	}
+	return b
+}
 
-	n := p * q
+// func main() {
```
tjgurwara99, line 166: Do let me know if you think that the main function that have been commented out should be removed completely.

I kept them here because I think they are useful for making new test cases, but a lot of them already have test files so I wasn't sure I should completely remove these blocks.

datastructures/binary-tree/binarysearchtree.go
```diff
@@ -1,126 +1,38 @@
...
-	fmt.Print(t.depth(), "\n")
-}
-
+package binarytree
+
+// import (
+// 	"fmt"
+// )
+
+// func main() {
```
tjgurwara99, line 10: This file is now redundant

ciphers/rsa/RSAcipher.go
```diff
@@ -1,11 +1,13 @@
-package main
+package rsacipher
```
tjgurwara99, line 1: This package fails the test for the encryption and decryption - I though it was failing because the implementation of modular exponentiation was incorrect but the test for modular exponentiation passes without issues so I'm actually not sure what the problem is. My current theory is that the inverse exponentiation function is incorrect but I did't try to test it.

ciphers/rsaBig/RSAcipher(Big).go
```diff
@@ -1,10 +1,10 @@
-package main
+package bigrsacipher
```
tjgurwara99, line 1: The test for this file passes but because of the automated command in the golangtest.yml file, this folder is also skipped because it shares the same prefix as rsa folder.

datastructures/linkedlist/singlylinkedlist/singlylinkedlist2.go
```diff
@@ -0,0 +1,104 @@
+package singlylinkedlist
+
+// TODO not needed because there is already an implementation of SingleLinkedList
```
tjgurwara99, line 3: comment is self-explanatory

go.mod
```diff
@@ -1,3 +1,5 @@
-module TheAlgorithms/Go
+module github.com/tjgurwara99/Go
```
tjgurwara99, line 1:
```suggestion
module github.com/TheAlgorithm/Go
```

tjgurwara99: The benefit of this can be seen in the `rot13.go` file in the cipher directory

other/monte_carlo_pi/monte_carlo_pi.go
```diff
@@ -1,50 +1,50 @@
...
-		}
-		wg.Add(1)
-		go fuxc(flow)
-		wg.Wait()
-		fmt.Printf("took %v \n",time.Since(start))
-		return 4 * float64(total) / float64(iterations)
-	}//end of monteCarlo
-	fmt.Println(monteCarlo(2<<18, cpus))
-}
+// func main() {
```
tjgurwara99, line 12: As you can see this is fully a main module and would create a lot of conflicts. If I get some time this week, will rewrite it similar to other files.

ciphers/rsa/RSAcipher.go
vedantmamgain: Yes it should be removed completely.

vedantmamgain:
LGTM

cclauss:
Congratulations!  This is a huge step forward.  My only regret is that we have a ton of code that does not pass `golangci-lint`.  The original goal was that we could successfully run the linter on ___all___ code in the repo like we do for the other TheAlgorithms language repos.  Is `golangci-lint` the linter that we should be using for Go code?  How can we get the entire codebase to pass the linter?

tjgurwara99:
closes: #198

PR #275, opened 2021-02-19
feat: Added a proper Modular Exponentiation algorithm

tjgurwara99:
I have added this because this algorithm is being re-written again and again in almost all of the files in the cipher directory.

After this gets merged then I can begin cleaning up cipher directory by using the repository's import.


tjgurwara99:
@cclauss Can you review this request, please? This will help me clean up the Cipher directory code.

math/modulararithmetic/modularexponentiation_test.go
```diff
@@ -0,0 +1,57 @@
...
+type cases struct {
+	name        string
+	description string
+	base        int64
+	exponent    int64
+	mod         int64
+	expected    int64
+}
+
+var testCases = []cases{
```
cclauss, line 14: Should tests include negative numbers to prove that we pass and fail properly?

math/modulararithmetic/modularexponentiation.go
```diff
@@ -0,0 +1,42 @@
...
+
+// ModularExponentiation returns base^exponent % mod
+func ModularExponentiation(base, exponent, mod int64) int64 {
+	if mod == 1 {
+		return 0
+	}
+	_, err := Multiply64BitInt(mod-1, mod-1)
+
+	if err != nil {
+		panic(err)
```
cclauss, line 19: Code in this repo tends to avoid calling panic().

https://github.com/TheAlgorithms/Go/search?q=panic

Is that really the most elegant solution for production code?

math/modulararithmetic/modularexponentiation_test.go
tjgurwara99: That's a good shout, I think I should add that

math/modulararithmetic/modularexponentiation.go
tjgurwara99: Are you talking about graceful error handling? for example:

```Go
func ModularExponentiation(base, exponent, mod int64) (int64, error) {
...
}

```

math/modulararithmetic/modularexponentiation_test.go
tjgurwara99: Although, now that I think about it, I don't know how to do that. Any ideas?

cclauss (approved):

PR #280, opened 2021-02-24
adding openaddress hashtable

Zzocker:
Added Following things
- Hash Table interface
```
// HashTable represents a hash table interface
type HashTable interface {
	// Put will insert value with key
	Put(key interface{}, value interface{})

	// Get : returns value of a given key
	// return nil if key doesn't exists
	Get(key interface{}) interface{}

	// Delete : remove the key from the table
	Delete(key interface{})

	// Contains : return true if given exists in table and false otherwise
	Contains(key interface{}) bool
}
```
- Collision resolved with open-addressing (linear)
- Added Benchmark
```
Hash Table With Chaining :
BenchmarkChainingPut-4               10000            209244 ns/op          859210 B/op         22 allocs/op
Hash Table With open addressing:
BenchmarkOpenAddresPut-4             1000000              3704 ns/op              99 B/op          5 allocs/op
Go in-build map
BenchmarkGoPut-4                   2091596               598 ns/op             171 B/op          2 allocs/op
  ```
Signed-off-by: Zzocker <pkspritam16@gmail.com>

### DIFF
--- a/datastructures/hashmap/hashtable.go
+++ b/datastructures/hashmap/hashtable.go
@@ -1,0 +1,17 @@
+package hashmap
+
+// HashTable represents a hash table interface
+type HashTable interface {
+	// Put will insert value with key
+	Put(key interface{}, value interface{})
+
+	// Get : returns value of a given key
+	// return nil if key doesn't exists
+	Get(key interface{}) interface{}
+
+	// Delete : remove the key from the table
+	Delete(key interface{})
+
+	// Contains : return true if given exists in table and false otherwise
+	Contains(key interface{}) bool
+}
--- a/datastructures/hashmap/hashtable.go
+++ b/datastructures/hashmap/hashtable.go
@@ -14,4 +14,15 @@ type HashTable interface {
 
 	// Contains : return true if given exists in table and false otherwise
 	Contains(key interface{}) bool
+
+	// Len : returns length of the hash table
+	Len() int
+
+	// Cap : returns current capacity of the hash table
+	Cap() int
+}
+
+// NewHashTable : creates new HashTable
+func NewHashTable() HashTable {
+	return newTable()
 }
--- a/datastructures/hashmap/withProbing.go
+++ b/datastructures/hashmap/withProbing.go
@@ -1,3 +1,5 @@
+// Package hashmap provide hash table implementation with collsion resolved
+// using openaddressing
 package hashmap
 
 import (
@@ -24,6 +26,7 @@ type table struct {
 	capacity uint64 // cap = 2^m
 }
 
+// newTable is internal function for creating new hash table
 func newTable() *table {
 	return &table{
 		buckets:  make([]*entry, 1<<initMValue),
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: datastructures/hashmap/hashtable.go [hunk 1]
```diff
@@ -0,0 +1,17 @@
+package hashmap
```
tjgurwara99, line 1: Please add a description of the package in a comment immediately proceeding the package declaration

```suggestion
Package hashmap <desciption goes here>
package hashmap
```

datastructures/hashmap/withProbing.go
```diff
@@ -0,0 +1,136 @@
...
+	loadFactor      float64 = 0.667
+	initMValue              = 1
+	deletedEntryKey         = "U+0000"
+)
+
+type entry struct {
+	key   interface{}
+	value interface{}
+}
+type table struct {
```
tjgurwara99, line 20: Please capitalise your main structure that you will want to be used in other package level code 😄

```suggestion
type Table struct {
```

tjgurwara99: This would also require you to make its fields that someone may want to use outside for example, `size` to `Size` would be accessible in outside scope.

datastructures/hashmap/withProbing.go
```diff
@@ -0,0 +1,136 @@
...
+	buckets  []*entry
+	size     uint64
+	m        uint64
+	capacity uint64 // cap = 2^m
+}
+
+func newTable() *table {
```
tjgurwara99, line 27:
```suggestion
func NewTable() *table {
```

tjgurwara99 (changes requested):

datastructures/hashmap/withProbing.go
Zzocker: I'm providing Hash table interface to the user, so there is no point in exporting Table or Size, 
for Size , I will add Len - method to the interface

tjgurwara99: Ah I see. I didn't see it as an example implementation.

Zzocker: Should I export those fields and struct? and remove the Hash Table interface ?

tjgurwara99: Up to you actually, I think I don't see how it can be used more generally than what you've already implemented. From what I understand you are using `HashTable` interface as a `return` type in the function `NewHashtable` which is essentially sending the `newTable` from the other file. So it is redundant in some sense, but at the same time this is indeed the proper way of doing it - because then people can just use the `HashTable` as an interface in functions to make it more general.

These are my personal opinions on your implementation so I think you should wait for the maintainers to see your code before making any changes.

tjgurwara99: Consider having an `Example` (if you are familiar with `Test Examples` in go - [here](https://blog.golang.org/examples) is the link if not) implementation so that people can see how they can implement something general that would work with that interface.

Again, I would say that it is worth waiting for the maintainers to review your code before making these changes.

PR #286, opened 2021-02-28
Stack data structure and DFS algorithm

kAvEh--:
- Stack data structure (there is no native implementation in GO)
- Depth-first Search algorithm in graph/search package


graphs/search/depthFirstSearch.go
```diff
@@ -0,0 +1,36 @@
...
+// reference: https://en.wikipedia.org/wiki/Tree_traversal
+package search
+
+import (
+	"github.com/TheAlgorithms/Go/datastructures/stack"
+)
+
+/*DepthFirstSearch is an algorithm for traversing or searching graph data structures.
+	The algorithm starts at the start node and explores as far as possible along each branch before backtracking.
+	Worst-case performance			O(|V|+|E|)}O(|V| + |E|)
```
tjgurwara99, line 11: The comment is good, but could you add what you mean by `V` and `E` - not many students would know that this refers to vertices and edges in a graph

tjgurwara99 (changes requested):
All looks good, but I would suggest for you to add a test for your stack implementation!

graphs/search/depthFirstSearch.go
kAvEh--: I think there is no need for more explanation. for more info about the algorithm, big-O, and other extra data there is a reference in the last line of the comment. prolonging the comments may have a negative effect.
this is my opinion but if you think more comment is better I will add another commit.

tjgurwara99: The comments are useful for making the documentation for this repository self-contained. Just visit [godoc.org](https://pkg.go.dev/github.com/TheAlgorithms/Go), you can see that quite a lot of code is not very well documented - precisely because the comments are missing. I personally think it would be very good for the documentation to be as self-contained as possible.

tjgurwara99 (approved):
This looks good to me!

tjgurwara99:
Could you please resolve the conflict. Also take a look at our new [contribution guidelines](https://github.com/TheAlgorithms/Go/blob/d2f12849df76203d80762ceca5105531d3424384/CONTRIBUTING.md) and make changes accordingly. There is already a Depth First Search algorithm in the repo. If you think that your implementation is different, please do let us know and we would be very happy to include your implementation. If however, no changes are made in the next two weeks, I will consider this PR to be stale and will close this PR then.

PR #289, opened 2021-03-04
Merge pull request #1 from TheAlgorithms/master

UnkleTayo:
Updates

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -4,18 +4,31 @@ on: [push, pull_request]
 jobs:
   golangci-lint:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
     steps:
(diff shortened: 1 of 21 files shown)


tjgurwara99 (commented):
For some reason I don't see any of the files that have been changed - the GitHub delta's show 0 additions and 0 deletions. Furthermore, the diff shows so many things that I think we have removed from this Repository in PR #266. Before the merge of this pr, this repository used to fail many of the checks. Those have been fixed and now we are at the stage where we are adding test cases. Unless I'm missing something, I think this PR is redundant, in fact it takes it back to the previous stage of this repository which used to fail checks.

Let me know if I'm missing something! Cause I don't see any point in this PR because the files changed doesn't show anything.

UnkleTayo:
Sorry, I never Intended to make a PR i was trying to update my own forked repo, never knew this what what I did, I'm sorry about that,

PR #314, opened 2021-07-24
Added set data structure with possible operations on it

meetvalani:
Added `Set` data structure.
Added following methods on `Set`:
- Add
- Delete
- Len
- GetItems
- In
- IsSubsetOf
- IsSupersetOf
- Union
- Intersection
- Difference
- SymmetricDifference


dittrichlucas (approved):

meetvalani:
@dittrichlucas, Can you re-review it, it's asking for some workflow approval.

set/set.go
```diff
@@ -0,0 +1,147 @@
+package set
+
+// GetSet gives new set.
+func GetSet(items ...interface{}) Set {
```
tjgurwara99, line 4: This style of naming Go naming conventions. Go package names are enough to make it clear what you're getting and instead of creating a constructor like `set.GetSet` it would be much more natural to use it like `set.New(items)`. Don't you agree?

meetvalani: Nice idea actually, we should conventions. I'll look into it.

siriak (approved):
LGTM, thanks!

PR #316, opened 2021-08-14
create leetcode section

SarveshGulhane:
Solutions for problems on leetcode in golang


siriak:
Hi, thanks for submitting, but we don't accept solutions to problems from websites with challenges, only documented and known algorithms

PR #321, opened 2021-08-25
Remove design patterns

siriak:


cclauss:
Can you please describe why this change is being made?

raklaptudirm:
@cclauss 
Copied from https://golangbyexample.com/abstract-factory-design-pattern-go/
Also, Design patterns are not algorithms, therefore, not in the scope of this repository.

raklaptudirm (approved):

Miraddo:
> Can you please describe why this change is being made?

@cclauss, because design patterns are not what we looking for  (the main concepts are just algorithms and data structure)  in other repositories we don't have the design patterns either, and I guess it should be removed as well.

as  @raklaptudirm mentioned it was also copied from golangbyexample.

for the set folder is part of data_structures so, I guess it should be moved into that directory.

raklaptudirm:
@Miraddo That can be done in a separate Pull Request, as these are separate problems.

Issue #303, opened 2021-04-05
Quicksort doesn't sort in-place


brayo-pip:
Quicksort doesn't sort in place. I was studying algorithms the other day and decided to try and implement quicksort in go. It was apparent that implementing an in place sorting algorithm would be tedious involving pointers and pointer indirection, so I decided to check the one already implemented here. And it's implementation is incorrect. It's still close enough to quicksort but the additional work that goes on to combine slices adds an overhead that shouldn't be there with quicksort. I will make a PR to address this soon!

siriak:
Hi, could you add a PR number here? I'll gladly review it. I totally agree with you that this bug should be addressed.

tjgurwara99:
I think the author has solved this in this a26459b.

Issue #269, opened 2021-02-18
Fix the failing sbom tests


cclauss:
**Describe the bug**
Fix the failing `sbom` tests so they pass and then remove `| grep -v /sbom` from
* https://github.com/TheAlgorithms/Go/blob/master/.github/workflows/golangtests.yml#L13
**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Issue #152, opened 2020-06-24
Code style and tests


L3str4nge:
Hello, have you consider any code style like in https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md#coding-style ? I don't see any `CONTIRBUTING.md` file in this repo. Maybe it will be worth it? How about using `gofmt`:https://golang.org/doc/effective_go.html#formatting ?

My second question is: can we add unit tests to github action? For now we only have lint check.

brayo-pip:
Hey I 'go fmt-ed' most of the Repo maybe u can write the contributing.md?

cclauss:
`go fmt` is the standard for Go and fortunately, there is no other!

Are you folks OK with the GitHub Actions https://github.com/TheAlgorithms/Go/tree/master/.github/workflows ?

brayo-pip:
Am not good with GitHub actions, anyway could u find a way to automate 'go fmt' once new code is merged. I automating it with subprocess.run() in python but for some reason it doesn't work on linux. Maybe u will be luckier

brayo-pip:
Do we need custom packages such as 'sorts' package or would we rather have the main function for each file? I think most ppl don't have the time or expertise to import these custom packages and run them. Choosing to simply use the main package maybe the way to go... what do u guys think?

Issue #37, opened 2018-10-09
Remove func main in files and add tests


eduardogr:
An improve in this repository could be remove commented main functions and add tests for the code in order to provide examples and let the code be executable easily

PR #325, opened 2021-08-28
Fix: A few structural changes to the repository and added missing tests and fixed some failing tests.

tjgurwara99:
Apologies for the big PR, I tried to make minimal changes but one change/fix lead to another broken thing so... here we are.

I have made a few changes to the repo and added RSA algorithm that was failing tests and had an incorrect implementation.

This is a big PR because I have changed some of the directory structure inside the repo. For example, it made more sense to call the modular exponentiation function as `modular.Exponentiation` than `modular_arithmetic.ModularExponentiation` which was done before (by me). Similar thing applies to `cipher` directories different ciphers. 

I have also added a few algorithms that were needed for the RSA algorithm in the relevant directories, example `modular.Inverse`.

I also noticed that the sorting algorithms tests were not set up correctly and because of which every test passed even though the algorithms were incorrect, example `QuickSort` and `RadixSort`. Unfortunately however, because of the way I have fixed it, it has resulted in `golangci-lint` failing because of one of their [open issue](https://github.com/golangci/golangci-lint/issues/826). I can fix it by changing the implementation but I think we should try and stick to go's ecosystem and use the tools provided by go team, that is `golint`. `golint` is much more stricter than golangci-lint in most cases (as far as I have seen and I think its a good thing). These tools are actually more strict and have a better overall effect on other go tools. But if the maintainers want me to fix it as it is, I will fix it - albeit the code will be a bit uglier.

Closes: #152
Closes: #303

---

The following are open issues that are actually fixed adding it here so that merge automatically closes these issues
Closes: #37
by #266
Closes: #269
by #324
Closes: #36 
by #266


siriak:
Let's switch to `golint` if it's stricter and more robust. Could you update the `golang_lint_ant_test` script in this PR?

tjgurwara99:
@siriak I looked into it and the go team have archived golint for some reason (about 4 months ago) so its highly unlikely that they will make progress in it. There are other tools that I can look at, would you like me to?

tjgurwara99:
[this](https://github.com/reviewdog/action-staticcheck) seems to be a good alternative.

cclauss:
https://pkg.go.dev/golang.org/x/lint is deprecated and frozen.

tjgurwara99:
> https://pkg.go.dev/golang.org/x/lint is deprecated and frozen.

Yeah, and unfortunately there is no stand in replacement for it.

cclauss:
My vote would be that we stick with https://github.com/golangci/golangci-lint which continues to be well maintained.

raklaptudirm (dismissed):

raklaptudirm (approved):

raklaptudirm (approved):

siriak (approved):
LGTM

siriak:
@tjgurwara99 could you resolve the conflicts so that I could merge?

siriak (approved):
LGTM, thanks!

PR #328, opened 2021-09-02
Liyl

yaolin-Li:
Get the length of the input parameter outside the loop to avoid performing length calculation operation every cycle. And modified the logic of statistics to make the code easier.


siriak (approved):
Thanks!

PR #329, opened 2021-09-03
More Cleaner and Take Care of the Structure

Miraddo:
## changes 
* `datastructure` folder deleted and Cyclically Linked-list moved to `data_structurs/linkedlist` directory
* added a link to Cyclically Linked-list readme file to have a reference and fixed some grammars 
* `moserdebruijnsequence` renamed folder to `moserde_bruijn_sequence`
* `interpolation_search.go` renamed folder to` interpolationsearch_search`
* `interpolation_search_test.go` renamed file to` interpolationsearch_test.go`
* added  Cyclically Linked-list, `moserde_bruijn_sequence, interpolationsearch_search to the algorithms list in the main readme.md


data_structures/linkedlist/cyclically_linkedlist/readme.md
```diff
@@ -1,12 +1,14 @@
-# Linked Cyclic List
+# Cyclic Linked List
 ***
 ## What is it?
 ***
 Loop lists are single or doubly-linked lists that chase their own tail:
-A points to B, B points to C, C points to D, and D points to A. 
+A point to B, B point to C, C point to D, and D point to A.
```
tjgurwara99, line 6:
```suggestion
A points to B, B points to C, C points to D, and D points to A.
```

tjgurwara99: The previous one was grammatically correct 😄

data_structures/linkedlist/cyclically_linkedlist/readme.md
```diff
@@ -1,12 +1,14 @@
...
 They are better suited for cyclic data such as train schedules.
 These lists are missing the first and last items.
 Therefore, it is necessary to introduce the concept of the current position.
 
+know more : [https://en.wikipedia.org/wiki/Cycle_detection](https://en.wikipedia.org/wiki/Cycle_detection)
```
tjgurwara99, line 11:
```suggestion
Find out more about [cycle detection here](https://en.wikipedia.org/wiki/Cycle_detection).
```

tjgurwara99 (commented):
Overall the changes are minimal so, apart from that everything is good. Although I still disagree with directory naming convention that this repository has started since one of your previous commits (Go programmers tend not to use underscores "_" character since it is considered "non-idiomatic" - find out more in [this thread](https://stackoverflow.com/a/63446957), it provides some good links where this is explained a bit more in depth). Also, because of renaming the directory you may need to also change the package name at the top of every file within the directory.

[This](https://stackoverflow.com/questions/43579838/relationship-between-a-package-statement-and-the-directory-of-a-go-file) gives more information on how the directory structure and package names are linked with each other.

Can you see the discrepancy with the naming you have chosen? You have a directory which is `data_structures` and another directory which is `linkedlist`. One has underscore and the other doesn't. This also contains different directories even though they all share the same kind of purpose, i.e. they are all linkedlists. Now I propose that you consider something like this:

1. (depending on what your style of writing packages is) rename the `data_structures` to `datastructure` or simply `structure` and inside this we have `linkedlist` which is the name of the package.
2. Now in this `linkedlist` directory instead of creating subdirectories for different types of linked lists, create a new file for that linkedlist and name the structure such that it suits the package name, for example: `linkedlist.Cyclical()` or something along those lines instead of calling this as `linkedlist.CyclicLinkedList()` (the latter is harder to even pronounce). You can see how former approach is more readable and easily understandable by just the naming things appropriately.

data_structures/linkedlist/cyclically_linkedlist/readme.md
Miraddo: :D Oh really? I used Grammarly, which shows me that `points to` might be incorrect, so I thought `point to` might be correct. Sorry about that.

Miraddo:
> Overall the changes are minimal so, apart from that everything is good. Although I still disagree with directory naming convention that this repository has started since one of your previous commits (Go programmers tend not to use underscores "_" character since it is considered "non-idiomatic" - find out more in [this thread](https://stackoverflow.com/a/63446957), it provides some good links where this is explained a bit more in depth). Also, because of renaming the directory you may need to also change the package name at the top of every file within the directory.
> 
> [This](https://stackoverflow.com/questions/43579838/relationship-between-a-package-statement-and-the-directory-of-a-go-file) gives more information on how the directory structure and package names are linked with each other.
> 
> Can you see the discrepancy with the naming you have chosen? You have a directory which is `data_structures` and another directory which is `linkedlist`. One has underscore and the other doesn't. This also contains different directories even though they all share the same kind of purpose, i.e. they are all linkedlists. Now I propose that you consider something like this:
> 
> 1. (depending on what your style of writing packages is) rename the `data_structures` to `datastructure` or simply `structure` and inside this we have `linkedlist` which is the name of the package.
> 2. Now in this `linkedlist` directory instead of creating subdirectories for different types of linked lists, create a new file for that linkedlist and name the structure such that it suits the package name, for example: `linkedlist.Cyclical()` or something along those lines instead of calling this as `linkedlist.CyclicLinkedList()` (the latter is harder to even pronounce). You can see how former approach is more readable and easily understandable by just the naming things appropriately.

Taj @tjgurwara99 , absolutely you are right, but in our structure, I used _ for the folder to make them a little bit more readable. however, for everything that you said, I agree with you. we could change the structure. If you had time could you change them too in a different PR?

siriak (changes requested):
I tend to agree with @tjgurwara99, based on our (recently updated) guidelines, file names must not contain underscores except for some very specific cases. Please fix that and re-request my review, thanks!

tjgurwara99:
@Miraddo you can close this PR. As per our discussion on discord, I will start working on the structure of the repository.

PR #309, opened 2021-05-29
roman_to_int remove string casting

soypat:
easier to read without string casting all over the place

PR #330, opened 2021-09-04
Remove verbose string casting in roman_to_int

soypat:
Redoing #309 since the conflict was way to complicated to be resolved by the likes of me.


raklaptudirm (approved):

raklaptudirm:
All looks fine, but would suggest a bit more precise names, instead of `roman`, `romans`, `holder`, etc.

siriak (approved):
Please re-request a review once conflicts are resolved

PR #310, opened 2021-05-29
Update CaesarCipher.go

soypat:
Less casting. more readable code

PR #331, opened 2021-09-04
Update caesar.go

soypat:
Redoing #310 since the merge was more than I could take. 

I've changed the program to resemble idiomatic Go:

* Slices to which we append are declared using `var` keyword.
* Reduce casting verbosity, ditch `int` casts and `byte` casts where not needed


raklaptudirm (approved):

siriak (approved):
Thanks!

PR #334, opened 2021-09-05
Update some functions in the btree directory

yaolin-Li:
1.Add AccessNodesByLayer Function, which is used to access nodes layer by layer instead of printing the results as one line.
```
{
 [0],
 [1] [2],
 [3] [4] [5] [6],
 [10]
}
```
2.Update the case of the parameters name in the main function so that it can run normally without error.


raklaptudirm (approved):

data_structures/binary_tree/binarysearchtree.go
```diff
@@ -8,31 +8,30 @@ package binarytree
...
+//t.Root = BstDelete(t.Root, 30)
+//InOrder(t.Root)
+//fmt.Print("\n")
+//t.Root = BstDelete(t.Root, 15)
+//InOrder(t.Root)
+//fmt.Print("\n")
+//t.Root = BstDelete(t.Root, 20)
+//InOrder(t.Root)
+//fmt.Print("\n")
+//fmt.Print(t.Depth(), "\n")
```
siriak, line 36: Please remove commented out code

data_structures/binary_tree/binarytree.go
```diff
@@ -3,24 +3,36 @@ package binarytree
 
 /*
```
siriak, line 4: And here

data_structures/binary_tree/binarysearchtree.go
siriak: We'll fix this is #336

siriak (approved):

PR #348, opened 2021-09-17
Powers in terms of logarithms

i-redbyte:
Powers in terms of logarithms

### DIFF
--- a/math/pi/spigotpi.go
+++ b/math/pi/spigotpi.go
@@ -1,0 +1,57 @@
+// spigotpi.go
+// description: A Spigot Algorithm for the Digits of Pi
+// details:
+// implementation of Spigot Algorithm for the Digits of Pi - [Spigot algorithm](https://en.wikipedia.org/wiki/Spigot_algorithm)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see spigotpi_test.go
+
+package pi
+
+import "strconv"
+
+func PiSpigot(n int) string {
+	pi := ""
+	boxes := n * 10 / 3
+	reminders := make([]int, boxes)
+	for i := 0; i < boxes; i++ {
+		reminders[i] = 2
+	}
+	heldDigits := 0
+	for i := 0; i < n; i++ {
+		carriedOver := 0
+		sum := 0
+		for j := boxes - 1; j >= 0; j-- {
+			reminders[j] *= 10
+			sum = reminders[j] + carriedOver
+			quotient := sum / (j*2 + 1)
+			reminders[j] = sum % (j*2 + 1)
+			carriedOver = quotient * j
+		}
+		reminders[0] = sum % 10
+		q := sum / 10
+		if q == 9 {
+			heldDigits++
+		} else if q == 10 {
+			q = 0
+			for k := 1; k <= heldDigits; k++ {
+				replaced, _ := strconv.Atoi(pi[i-k : i-k+1])
+				if replaced == 9 {
+					replaced = 0
+				} else {
+					replaced++
+				}
+				pi = string(delChar([]rune(pi), i-k))
+				pi = pi[:i-k] + strconv.Itoa(replaced) + pi[i-k:]
+			}
+			heldDigits = 1
+		} else {
+			heldDigits = 1
+		}
+		pi += strconv.Itoa(q)
+	}
+	return pi
+}
+
+func delChar(s []rune, index int) []rune {
+	return append(s[0:index], s[index+1:]...)
+}
--- a/math/pi/spigotpi.go
+++ b/math/pi/spigotpi.go
@@ -16,7 +16,7 @@ func Spigot(n int) string {
 	for i := 0; i < boxes; i++ {
 		remainders[i] = 2
 	}
-	heldDigits := 0
+	digitsHeld := 0
 	for i := 0; i < n; i++ {
 		carriedOver := 0
 		sum := 0
@@ -31,10 +31,10 @@ func Spigot(n int) string {
 		q := sum / 10
 		switch q {
 		case 9:
-			heldDigits++
+			digitsHeld++
 		case 10:
 			q = 0
-			for k := 1; k <= heldDigits; k++ {
+			for k := 1; k <= digitsHeld; k++ {
 				replaced, _ := strconv.Atoi(pi[i-k : i-k+1])
 				if replaced == 9 {
 					replaced = 0
@@ -44,9 +44,9 @@ func Spigot(n int) string {
 				pi = delChar(pi, i-k)
 				pi = pi[:i-k] + strconv.Itoa(replaced) + pi[i-k:]
 			}
-			heldDigits = 1
+			digitsHeld = 1
 		default:
-			heldDigits = 1
+			digitsHeld = 1
 		}
 		pi += strconv.Itoa(q)
 	}
--- a/math/pi/spigotpi.go
+++ b/math/pi/spigotpi.go
@@ -12,22 +12,22 @@ import "strconv"
 func Spigot(n int) string {
 	pi := ""
 	boxes := n * 10 / 3
-	reminder := make([]int, boxes)
+	remainders := make([]int, boxes)
 	for i := 0; i < boxes; i++ {
-		reminder[i] = 2
+		remainders[i] = 2
 	}
 	heldDigits := 0
 	for i := 0; i < n; i++ {
 		carriedOver := 0
 		sum := 0
 		for j := boxes - 1; j >= 0; j-- {
-			reminder[j] *= 10
-			sum = reminder[j] + carriedOver
+			remainders[j] *= 10
+			sum = remainders[j] + carriedOver
 			quotient := sum / (j*2 + 1)
-			reminder[j] = sum % (j*2 + 1)
+			remainders[j] = sum % (j*2 + 1)
 			carriedOver = quotient * j
 		}
-		reminder[0] = sum % 10
+		remainders[0] = sum % 10
 		q := sum / 10
 		switch q {
 		case 9:
@@ -41,7 +41,7 @@ func Spigot(n int) string {
 				} else {
 					replaced++
 				}
-				pi = string(delChar(pi, i-k))
+				pi = delChar(pi, i-k)
 				pi = pi[:i-k] + strconv.Itoa(replaced) + pi[i-k:]
 			}
 			heldDigits = 1
@@ -53,7 +53,7 @@ func Spigot(n int) string {
 	return pi
 }
 
-func delChar(s string, index int) []rune {
+func delChar(s string, index int) string {
 	tmp := []rune(s)
-	return append(tmp[0:index], tmp[index+1:]...)
+	return string(append(tmp[0:index], tmp[index+1:]...))
 }
--- a/math/pi/spigotpi.go
+++ b/math/pi/spigotpi.go
@@ -9,29 +9,30 @@ package pi
 
 import "strconv"
 
-func PiSpigot(n int) string {
+func Spigot(n int) string {
 	pi := ""
 	boxes := n * 10 / 3
-	reminders := make([]int, boxes)
+	reminder := make([]int, boxes)
 	for i := 0; i < boxes; i++ {
-		reminders[i] = 2
+		reminder[i] = 2
 	}
 	heldDigits := 0
 	for i := 0; i < n; i++ {
 		carriedOver := 0
 		sum := 0
 		for j := boxes - 1; j >= 0; j-- {
-			reminders[j] *= 10
-			sum = reminders[j] + carriedOver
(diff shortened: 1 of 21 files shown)


raklaptudirm (dismissed):

math/power/powvialogarithm.go
```diff
@@ -0,0 +1,23 @@
...
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see powvialogarithm_test.go
+
+package power
+
+import (
+	"math"
+)
+
+func PowUseLog(a int, b int) int {
```
tjgurwara99, line 14: shouldn't you keep the arguments `float64`? The `math.Log` and `math.Exp` use float64 so all the type casting you use will not be necessary.

```suggestion
func PowUseLog(a float64, b float64) float64 {
```

tjgurwara99 (changes requested):
Please consider the following changes. Overall, as a mathematician I really like this approach :)

Also, I think this will be faster too from the other approach even though the expensive operations are being used. These expensive operations are assembly commands in most modern architecture which go takes advantage of. If it is possible could you add a benchmark for these `power` algorithms? If you can't, then just consider the recommended changes and we can approve this PR. Thanks!

siriak (dismissed):
LGTM apart from the one comment from @tjgurwara99

math/power/powvialogarithm.go
i-redbyte: Done

raklaptudirm (approved):

math/pi/spigotpi.go
```diff
@@ -0,0 +1,57 @@
...
+				} else {
+					replaced++
+				}
+				pi = string(delChar([]rune(pi), i-k))
+				pi = pi[:i-k] + strconv.Itoa(replaced) + pi[i-k:]
+			}
+			heldDigits = 1
+		} else {
+			heldDigits = 1
+		}
```
tjgurwara99, line 49: Wouldn't it be better to convert this to a simple switch case distinction?

math/pi/spigotpi.go
```diff
@@ -0,0 +1,57 @@
...
+		pi += strconv.Itoa(q)
+	}
+	return pi
+}
+
+func delChar(s []rune, index int) []rune {
+	return append(s[0:index], s[index+1:]...)
+}
```
tjgurwara99, line 57: Do we really need this?

tjgurwara99: On the other hand it makes this more readable - keep it there. It's fine.

math/pi/spigotpi_test.go
```diff
@@ -0,0 +1,51 @@
...
+	result string
+	n      int
+}{
+	{"314159", 6},
+	{"31415926535", 11},
+	{"314159265358", 12},
+	{"314159265358979323846", 21},
+	{"314", 3},
+	{"31415", 5},
+}
```
tjgurwara99, line 22: Can you make this local inside the `TestPiSpigot` function or create a factory function that generates these test cases? Without a factory or local initialisation it is possible to have conflict with other implementations. I'm aware that this is the only algorithm inside this directory but if someone adds another algorithm, it could create conflict - better to be safe then sorry 😄

math/power/powvialogarithm.go
```diff
@@ -0,0 +1,24 @@
...
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see powvialogarithm_test.go
+
+package power
+
+import (
+	"math"
+)
+
+func PowUseLog(a float64, b float64) float64 {
```
tjgurwara99, line 14: Can you change the name of the function? if someone was to use this in there code to try it out they would have to use it like this:

```Go
import github.com/TheAlgorithms/Go/math/power
...

func main() {
    ...
    power.PowUseLog(a, b)
    ...
}

```

which is arguably not very reader friendly - specially if someone is trying to learn from - the main purpose of this repo 😄 

My recommendation would be that you have something like this:

```Go
import github.com/TheAlgorithms/Go/math/power
...

func main() {
    ...
    power.UsingLog(a, b)
    ...
}
```

so it would be easier to parse through - sorry this is actually quite pedantic but believe me, readability makes a huge difference when parsing through code 😄 .

```suggestion
func UsingLog(a float64, b float64) float64 {
```

math/power/powvialogarithm_test.go
```diff
@@ -0,0 +1,44 @@
+// powvialogarithm_test.go
+// description: Test for PowUseLog
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see powvialogarithm.go
+
+package power
+
+import "testing"
+
+var tests = []struct {
```
tjgurwara99, line 10: Same as previous comment - could you move this to a local declaration/initialisation or factory generation instead? Same point as other test comment applies 😄

math/power/powvialogarithm_test.go
```diff
@@ -0,0 +1,44 @@
...
+	{"2^0", 2, 0, 1},
+	{"2^3", 2, 3, 8},
+	{"8^3", 8, 3, 512},
+	{"11^11", 11, 11, 285311670611},
+	{"5^5", 5, 5, 3125},
+	{"-7^2", -7, 2, 49},
+	{"-6^3", -6, 3, -216},
+}
+
+func TestPowUseLog(t *testing.T) {
```
tjgurwara99, line 28: reflect the names if you change the name of the PowUseLog function

math/power/powvialogarithm_test.go
```diff
@@ -0,0 +1,44 @@
...
+			result := PowUseLog(tc.base, tc.power)
+			t.Log(result)
+			if result != tc.expected {
+				t.Errorf("Expected %.2f to the power of %.2f to be: %.2f, but got: %.2f", tc.base, tc.power, tc.expected, result)
+			}
+		})
+	}
+}
+
+func BenchmarkPowUseLog(b *testing.B) {
```
tjgurwara99, line 40: Same for this!

tjgurwara99 (changes requested):
Nice implementations! As a mathematician I really liked your implementations. Please consider the changes that I have requested.

Also, could I ask you to make new PR's if you are submitting more than one implementation? This makes it easier to track changes and doesn't put a lot of load on maintainers 😄 If submitting another PR is not possible, then just update the PR's description and title to reflect the changes you have made.

If you have any troubles understanding the material, let me know and I'll be happy to discuss.

Thanks for contributing and happy coding!

math/pi/spigotpi.go
```diff
@@ -0,0 +1,57 @@
...
+// details:
+// implementation of Spigot Algorithm for the Digits of Pi - [Spigot algorithm](https://en.wikipedia.org/wiki/Spigot_algorithm)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see spigotpi_test.go
+
+package pi
+
+import "strconv"
+
+func PiSpigot(n int) string {
```
tjgurwara99, line 12: Could you also change the name of this function - `pi` is already in the package name so `Spigot` should suffice.

math/pi/spigotpi.go
```diff
@@ -0,0 +1,57 @@
...
+	pi := ""
+	boxes := n * 10 / 3
+	reminders := make([]int, boxes)
+	for i := 0; i < boxes; i++ {
+		reminders[i] = 2
+	}
+	heldDigits := 0
```
tjgurwara99, line 19: `digitsHeld` instead of `heldDigits`
```suggestion
	digitsHeld := 0
```

math/pi/spigotpi.go
```diff
@@ -0,0 +1,57 @@
...
+	for i := 0; i < n; i++ {
+		carriedOver := 0
+		sum := 0
+		for j := boxes - 1; j >= 0; j-- {
+			reminders[j] *= 10
+			sum = reminders[j] + carriedOver
+			quotient := sum / (j*2 + 1)
+			reminders[j] = sum % (j*2 + 1)
```
tjgurwara99, line 27: typo here
```suggestion
			remainder[j] = sum % (j*2 + 1)
```

math/pi/spigotpi.go
```diff
@@ -0,0 +1,57 @@
...
+		} else if q == 10 {
+			q = 0
+			for k := 1; k <= heldDigits; k++ {
+				replaced, _ := strconv.Atoi(pi[i-k : i-k+1])
+				if replaced == 9 {
+					replaced = 0
+				} else {
+					replaced++
+				}
+				pi = string(delChar([]rune(pi), i-k))
```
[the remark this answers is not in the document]
i-redbyte: Done

math/pi/spigotpi.go
i-redbyte: Done

math/pi/spigotpi.go
tjgurwara99: Don't worry - I'm also a non-native speaker 😄

math/power/powvialogarithm.go
i-redbyte: Totally agree with you. Fixed names

math/power/powvialogarithm_test.go
i-redbyte: Done

math/power/powvialogarithm_test.go
i-redbyte: Done

math/pi/spigotpi.go
i-redbyte: Done

math/pi/spigotpi.go
tjgurwara99: is it resolved? - I don't see it outdated. I don't think you have changed this variable - you might have missed it, haha 😄

math/pi/spigotpi.go
```diff
@@ -0,0 +1,59 @@
...
+// see spigotpi_test.go
+
+package pi
+
+import "strconv"
+
+func Spigot(n int) string {
+	pi := ""
+	boxes := n * 10 / 3
+	reminder := make([]int, boxes)
```
tjgurwara99, line 15: You are missing an `a` in the middle 😃 

```suggestion
	remainders := make([]int, boxes)
```

math/pi/spigotpi.go
```diff
@@ -0,0 +1,59 @@
...
+		case 10:
+			q = 0
+			for k := 1; k <= heldDigits; k++ {
+				replaced, _ := strconv.Atoi(pi[i-k : i-k+1])
+				if replaced == 9 {
+					replaced = 0
+				} else {
+					replaced++
+				}
+				pi = string(delChar(pi, i-k))
```
tjgurwara99, line 44: The string conversion is also something that can be handled inside the delChar function - sorry I missed it last time.

```suggestion
				pi = delChar(pi, i-k)
```

math/pi/spigotpi.go
```diff
@@ -0,0 +1,59 @@
...
+			heldDigits = 1
+		}
+		pi += strconv.Itoa(q)
+	}
+	return pi
+}
+
+func delChar(s string, index int) []rune {
+	tmp := []rune(s)
+	return append(tmp[0:index], tmp[index+1:]...)
```
tjgurwara99, line 58: Just do

```suggestion
	return string(append(tmp[0:index], tmp[index+1:]...))
```

math/power/powvialogarithm.go
```diff
@@ -0,0 +1,24 @@
...
+package power
+
+import (
+	"math"
+)
+
+func UsingLog(a float64, b float64) float64 {
+	var p float64
+	p = 1
+	if a < 0 && int(b)&1 != 0 {
```
tjgurwara99, line 17: Good use of bitwise operations. Nice!!

tjgurwara99 (changes requested):
Everything looks good other than some minor typo's. Very good, keep it up!

math/pi/spigotpi.go
i-redbyte: maybe then it's better like this:
func delChar(s string, index int) string{
	tmp := []rune(s)
	return string(append(tmp[0:index], tmp[index+1:]...))
}

math/pi/spigotpi.go
i-redbyte: Done

math/pi/spigotpi.go
i-redbyte: Done

math/pi/spigotpi.go
i-redbyte: Done

tjgurwara99 (dismissed):
All good except the `helfDigits` slice name. It should be `digitsHeld`. I'm approving now, but if you change it, I'll approve it again. Thanks for contributing.

tjgurwara99 (approved):
LGTM! Thanks for contributing!

PR #33, opened 2018-09-25
Backtracking algorithms added

Shivam010:

PR #349, opened 2021-09-19
feat: add Graph Coloring algorithms

Shivam010:
This will add 3 graph coloring algorithms
  * Using BFS
  * Using Backtracking
  * Using Greedy Approach

Ref #33

### DIFF
--- a/graph/coloring/backtracking.go
+++ b/graph/coloring/backtracking.go
@@ -1,8 +1,11 @@
+// This file contains the graph coloring implementation using backtracking
+// Author(s): [Shivam](https://github.com/Shivam010)
+
 package coloring
 
-// ColoringUsingBacktracking will return the Color of each vertex and the
+// ColorUsingBacktracking will return the Color of each vertex and the
 // total number of different colors used, using backtracking
-func (g *Graph) ColoringUsingBacktracking() (map[int]Color, int) {
+func (g *Graph) ColorUsingBacktracking() (map[int]Color, int) {
 	vertexColors := make(map[int]Color, g.vertices)
 	g.colorVertex(0, vertexColors)
 
--- a/graph/coloring/backtracking.go
+++ b/graph/coloring/backtracking.go
@@ -1,0 +1,47 @@
+package coloring
+
+// ColoringUsingBacktracking will return the Color of each vertex and the
+// total number of different colors used, using backtracking
+func (g *Graph) ColoringUsingBacktracking() (map[int]Color, int) {
+	vertexColors := make(map[int]Color, g.vertices)
+	g.colorVertex(0, vertexColors)
+
+	colorsUsed := 0
+	for _, cr := range vertexColors {
+		if colorsUsed < int(cr) {
+			colorsUsed = int(cr)
+		}
+	}
+	return vertexColors, colorsUsed
+}
+
+// colorVertex will try to color provided vertex, v
+func (g *Graph) colorVertex(v int, color map[int]Color) bool {
+	// If all vertices are colored, the colors store will be completely filled.
+	if len(color) == g.vertices {
+		return true
+	}
+
+	// As the upper bound of no. of colors is the no. of vertices in graph,
+	// try assigning each color to the vertex v
+	for cr := Color(1); cr <= Color(g.vertices); cr++ {
+		// Use the color, cr for vertex, v if it is safe to use, by
+		// checking its neighbours
+		safe := true
+		for nb := range g.edges[v] {
+			// cr, color is not safe if color of nb, crnb is not equal to cr
+			if crnb, ok := color[nb]; ok && crnb == cr {
+				safe = false
+				break
+			}
+		}
+		if safe {
+			color[v] = cr
+			if g.colorVertex(v+1, color) {
+				return true
+			}
+			delete(color, v)
+		}
+	}
+	return false
+}
--- a/graph/coloring/backtracking_test.go
+++ b/graph/coloring/backtracking_test.go
@@ -1,3 +1,6 @@
+// This file provides tests for coloring using backtracking.
+// Author(s): [Shivam](https://github.com/Shivam010)
+
 package coloring_test
 
 import (
@@ -5,16 +8,16 @@ import (
 	"testing"
 )
 
-func TestGraphColoringUsingBacktracking(t *testing.T) {
+func TestGraphColorUsingBacktracking(t *testing.T) {
 	for i, tt := range getTestGraphs() {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
-			colorsOfVertices, colors := tt.Graph.ColoringUsingBacktracking()
+			colorsOfVertices, colors := tt.Graph.ColorUsingBacktracking()
 			if colors != tt.ColorsUsed {
-				t.Errorf("ColoringUsingBacktracking() return more number of colors: %v, want %v colors", colors, tt.ColorsUsed)
+				t.Errorf("ColorUsingBacktracking() return more number of colors: %v, want %v colors", colors, tt.ColorsUsed)
 			}
 			// check colors
 			if err := tt.Graph.ValidateColorsOfVertex(colorsOfVertices); err != nil {
-				t.Errorf("ColoringUsingBacktracking() assigned colors are wrong, error = %v", err)
+				t.Errorf("ColorUsingBacktracking() assigned colors are wrong, error = %v", err)
 			}
 		})
 	}
--- a/graph/coloring/backtracking_test.go
+++ b/graph/coloring/backtracking_test.go
@@ -1,0 +1,21 @@
+package coloring_test
+
+import (
+	"strconv"
+	"testing"
+)
+
+func TestGraphColoringUsingBacktracking(t *testing.T) {
+	for i, tt := range getTestGraphs() {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			colorsOfVertices, colors := tt.Graph.ColoringUsingBacktracking()
+			if colors != tt.ColorsUsed {
+				t.Errorf("ColoringUsingBacktracking() return more number of colors: %v, want %v colors", colors, tt.ColorsUsed)
+			}
+			// check colors
+			if err := tt.Graph.ValidateColorsOfVertex(colorsOfVertices); err != nil {
+				t.Errorf("ColoringUsingBacktracking() assigned colors are wrong, error = %v", err)
+			}
+		})
+	}
+}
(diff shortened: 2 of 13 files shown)


graph/coloring/bfs.go
```diff
@@ -0,0 +1,53 @@
...
+
+	visited := make(map[int]struct{})
+	// Run BFS from each non-visited vertex
+	for i := 0; i < g.vertices; i++ {
+		if _, ok := visited[i]; ok {
+			continue
+		}
+		visited[i] = struct{}{}
+
+		queue := list.New()
```
tjgurwara99, line 22: Could it not be possible to use `TheAlgorithms/Go`repo's implementation of a queue? I think its almost same, or am I missing something?

graph/coloring/backtracking.go
```diff
@@ -0,0 +1,47 @@
+package coloring
```
tjgurwara99, line 1: Is adding a new `coloring` package necessary? I don't think so but if you can explain your reasoning, it might be ok. Let me know your thoughts.

graph/coloring/graph_test.go
```diff
@@ -0,0 +1,150 @@
...
+	g0.Graph.AddEdge(4, 3)
+	g0.Graph.AddEdge(3, 1)
+	g0.Graph.AddEdge(3, 2)
+	g0.Graph.AddEdge(1, 2)
+	g0.Graph.AddEdge(2, 0)
+
+	// Graph 1st:
+	// 		1---2
+	// 		| / |
+	//  4---3---0
```
siriak, line 40: Please here and in other places use spaces instead of tabs because the ascii pictures are distorted because of the latter

graph/coloring/backtracking.go
```diff
@@ -0,0 +1,47 @@
...
+
+// ColoringUsingBacktracking will return the Color of each vertex and the
+// total number of different colors used, using backtracking
+func (g *Graph) ColoringUsingBacktracking() (map[int]Color, int) {
```
tjgurwara99, line 5: Naming is not fluent, consider changing it to:

```suggestion
func (g *Graph) ColorUsingBacktracking() (map[int]Color, int) {
```

This naming would give an English friendly readability. For example,

```Go
g := colouring.Graph{...}

g.ColorUsingBacktracking()
...
```

graph/coloring/backtracking_test.go
```diff
@@ -0,0 +1,21 @@
+package coloring_test
+
+import (
+	"strconv"
+	"testing"
+)
+
+func TestGraphColoringUsingBacktracking(t *testing.T) {
```
tjgurwara99, line 8: Change according to the other naming changes :)

siriak (dismissed):
LGTM apart from the tabs :)

graph/coloring/bfs.go
```diff
@@ -0,0 +1,53 @@
+package coloring
+
+import "container/list"
+
+// ColoringUsingBFS will return the Color of each vertex and the
+// total number of different colors used, using BFS
+func (g *Graph) ColoringUsingBFS() (map[int]Color, int) {
```
tjgurwara99, line 7:
```suggestion
func (g *Graph) ColorUsingBFS() (map[int]Color, int) {
```

graph/coloring/graph.go
```diff
@@ -0,0 +1,53 @@
+package coloring
+
+import "errors"
+
+// Color provides a type for vertex color
+type Color int
+
+// Graph provides a structure to store an undirected graph.
+// It is safe to use its empty object.
+type Graph struct {
```
tjgurwara99, line 10: Is it not possible to put it in the `graph` package instead of adding it inside the `coloring` package?

graph/coloring/graph_test.go
```diff
@@ -0,0 +1,150 @@
...
+	for i := 0; i < 4; i++ {
+		for j := i + 1; j < 4; j++ {
+			g4.Graph.AddEdge(i, j)
+		}
+	}
+
+	return
+}
+
+func TestGraph_ValidateColorsOfVertex(t *testing.T) {
```
tjgurwara99, line 118: I don't like the underscores being used in Go. It kind of distracts me 😅
```suggestion
func TestGraphValidateColorsOfVertex(t *testing.T) {
```
But there is no hard naming conventions on tests in the Go community so it would be fine even if you leave it. I would say that I don't think your names would clash with other implementations so it is worth considering my suggestion above. But again no hard rule on this 😄

graph/coloring/graph_test.go
```diff
@@ -0,0 +1,150 @@
...
+		in := len(list[2].VertexColors) - v - 1
+		list[2].VertexColors[v] = list[2].VertexColors[in]
+	}
+	for v := range list[3].VertexColors {
+		list[3].VertexColors[v] = 1
+	}
+	return list[:4]
+}
+
+func TestGraph_ValidateColorsOfVertex_Negative(t *testing.T) {
```
tjgurwara99, line 142: Same suggestion as above ☝️ 
```suggestion
func TestGraphValidateColorsOfVertexNegative(t *testing.T) {
```

graph/doc.go
```diff
@@ -1,3 +1,3 @@
-// Package graph demonstates Graph search algorithms
+// Package graph demonstrates Graph search algorithms
```
tjgurwara99, line 1: It would also be nice to add your authorship in these files 😄. Look at our contribution guidelines, it has an example of this.

```suggestion
// filename
// description: Add one line description here
// details:
// This is a multi-line
// description containing links, references,
// math equations, etc.
// author(s) [Name](https://github.com/handle), [Name](https://github.com/handle)
// see relatedfile.go, anotherfile.go, file_test.go
 
// ** Is just an example of how to write description for package and function and other stuff ... **

// Package sort provides primitives for sorting slices and user-defined
// collections.

// Package graph demonstrates Graph search algorithms
```

tjgurwara99 (changes requested):
I really like your approach. I think overall its very robust (from a mathematical perspective) just minor styling issues (in my opinion). Consider the changes requested, however if you think those are unnecessary, we can discuss that here 😄

graph/coloring/graph_test.go
Shivam010: Yeah, sorry. I always forgot to use spaces instead of tabs in diagrams.

graph/coloring/bfs.go
Shivam010: Yeah, we can, I used the `container/list` because of the reflexes.

graph/coloring/backtracking.go
Shivam010: I wanted to have all the implementations collectively in a single place, that's why I created a separate package.

graph/coloring/backtracking.go
Shivam010: Yes, now I also see that. I'll fix that.

graph/coloring/backtracking.go
tjgurwara99: Yeah, fair point. I thought since you were using a linear undirected graph, it could still be considered part of the graph package.

graph/coloring/graph_test.go
Shivam010: > I don't like the underscores being used in Go. It kind of distracts me 😅

I also, don't like them, but the Go's test function generator keeps on using it

graph/doc.go
Shivam010: At first, I thought of adding it, but *(no disrespect)* I didn't like its format.

I'll add it no problem.

graph/coloring/graph_test.go
tjgurwara99: Which one are you referring to? I haven't seen a test generator supported by gotools - I'd like to know! 😃

graph/doc.go
tjgurwara99: Ah ok, if you'd rather not that's fine – just thought that I should mention it. If you have a suggestion for better format, let us know in the issues if it is something that can be considered.

graph/coloring/graph_test.go
Shivam010: Ah, no-no, I use GoLand IDE and it provides shortcuts for creating test functions. 
Even I haven't seen any test generator by gotools, I think it uses https://github.com/cweill/gotests, not sure.

graph/coloring/graph.go
Shivam010: On second thought, I think we should avoid this, doing so will make us convert all the functions from `Graph` methods to functions. And I think we should stick to the current implementation. 

We can copy the `graph.go` file to the `graph` package for others to use. 

GoProverb: "A little copying is better..."

Shivam010: What do you think?

tjgurwara99: Yeah, that sounds reasonable. In fact, if you add the implementation then you can use your methods as functions too - just pass it as first argument in your function call. Although I don't think that style should be promoted. Anyways, add the Linear undirected graph to graph package and mark this as resolved😄

graph/coloring/graph_test.go
tjgurwara99: Ah cool, then I think you should remove the underscore since the whole repo is using underscores. Thanks!

graph/doc.go
Shivam010: How about a simple description just like as we do with the package description, along with an authorship comment?

```go
// This file contains the simple structural implementation of undirected
// graph. Used in coloring algorithms [here](./coloring)
// Author(s): [Shivam](https://github.com/Shivam010)

package graph
```

tjgurwara99: As long as it is not the package level comment it should be fine. Package level comment (comment right above the package declaration without a '\n' character in between) is used to generate this repos automatic documentation in godocs so just keep that in mind😄.

Shivam010: Yes, that's why I kept the line in my suggestion as well.

tjgurwara99 (approved):
If possible could you add a package level comment describing and referencing the colouring problem in a separate `doc.go` file? Everything apart from that looks good to me!

graph/coloring/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package coloring provides implementation of different graph coloring
+// algorithms, e.g. coloring using BFS, using Backtracking, using greedy
+// approach.
+// Author(s): [Shivam](https://github.com/Shivam010)
+package coloring
```
Shivam010, line 5: I already added that here. 
Are referring to something else?

tjgurwara99: Oh no that's it. Sorry missed that. Thought there was a newline here 😅

tjgurwara99:
LGTM! Thanks for contributing.

raklaptudirm (approved):

PR #350, opened 2021-09-23
Find a missing number in a sequence

i-redbyte:
Given an array A containing n distinct numbers in the range [0, n], return the only number in the range that is missing from the array.


raklaptudirm (approved):

siriak (approved):
LGTM, thanks!

PR #353, opened 2021-09-24
optimized roman to int and added int to roman

brandenc40:
- Added an algorithm for converting an integer to a roman numeral string
- Optimized the existing roman numeral string to integer algorithm. I also made it to return an error if out of bounds, not sure if this is allowed as the old algorithm did not return an error. let me know if that part should be removed

Benchmarks:
Screen Shot 2021-09-23 at 10 10 11 PM

### DIFF
--- a/conversion/integertoroman.go
+++ b/conversion/integertoroman.go
@@ -1,0 +1,22 @@
+package conversion
+
+import "errors"
+
+var (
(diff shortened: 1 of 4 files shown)


raklaptudirm (approved):
No, there is no problem in adding the limit. In fact, I would consider it to be an improvement to the algorithm.

siriak (approved):
LGTM, thanks!

PR #354, opened 2021-09-24
Feat monte carlo pi

i-redbyte:
calculating pi by the Monte Carlo method

### DIFF
--- a/math/pi/montecarlopi.go
+++ b/math/pi/montecarlopi.go
@@ -1,0 +1,27 @@
+// montecarlopi.go
+// description: Calculating pi by the Monte Carlo method
+// details:
+// implementation of Monte Carlo Algorithm for the calculating of Pi - [Monte Carlo method](https://en.wikipedia.org/wiki/Monte_Carlo_method)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see montecarlopi_test.go
+
+package pi
+
+import (
+	"math/rand"
+	"time"
+)
+
+func MonteCarloPi(n int) float64 {
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+	pass := 0
+	for i := 0; i < n; i++ {
+		x := rnd.Float64()
+		y := rnd.Float64()
+		if x*x+y*y <= 1 {
+			pass += 1
+		}
+	}
+	pi := float64(pass) / float64(n) * float64(4)
+	return pi
+}
--- a/math/pi/montecarlopi.go
+++ b/math/pi/montecarlopi.go
@@ -12,16 +12,16 @@ import (
 	"time"
 )
 
-func MonteCarloPi(n int) float64 {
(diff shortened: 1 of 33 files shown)


raklaptudirm (approved):

math/pi/montecarlopi.go
```diff
@@ -0,0 +1,27 @@
...
+
+package pi
+
+import (
+	"math/rand"
+	"time"
+)
+
+func MonteCarloPi(n int) float64 {
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
```
tjgurwara99, line 16: Good - when I looked at it in the morning this was not clear. I didn't have time to write the comments but I like this change. 😄

math/pi/montecarlopi.go
```diff
@@ -0,0 +1,27 @@
...
+	pass := 0
```
tjgurwara99, line 17: On the other this is not the correct wording for this. I would call this variable something to signify that this is the inside of the unit circle.

Maybe it would be best to call this variable `circle` or `inside` not sure which one fits best. One thing I'm sure of is that `pass` is not a good name for this.

```suggestion
	circle := 0
```

math/pi/montecarlopi.go
```diff
@@ -0,0 +1,27 @@
...
+// see montecarlopi_test.go
+
+package pi
+
+import (
+	"math/rand"
+	"time"
+)
+
+func MonteCarloPi(n int) float64 {
```
tjgurwara99, line 15: Try to be descriptive when writing code for someone else (for example this repository) 😄 I learnt it the hard way 😅.

```suggestion
func MonteCarloPi(randomPoints int) float64 {
```

or

```suggestion
func MonteCarloPi(dots int) float64 {
```

math/pi/montecarlopi_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+package pi
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestMonteCarloPi(t *testing.T) {
+	t.Run("BruteForce Pi", func(t *testing.T) {
```
tjgurwara99, line 14: Why is this test named as BruteForce Pi?

tjgurwara99 (changes requested):
Consider the changes that I've requested. Once changes are made, I'll approve this PR 😄

math/pi/montecarlopi_test.go
i-redbyte: Fix

math/pi/montecarlopi.go
i-redbyte: Done

math/pi/montecarlopi.go
i-redbyte: I chose the name inside

tjgurwara99 (approved):
LGTM. Thanks for contributing!

siriak (approved):

PR #357, opened 2021-09-26
Center of the circle

i-redbyte:
Find the center of the circle.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Being a contributor at The Algorithms, we request you to follow the points menti
 
 - You did your own work.
   - No plagiarism is allowed. Any plagiarized work will not be merged.
-- Your work will be distributed under the [MIT License](https://github.com/TheAlgoritms/Go/blob/master/LICENSE) once your pull request has been merged.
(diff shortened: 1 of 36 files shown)


raklaptudirm:
In my opinion, this is a trivial algorithm.

i-redbyte:
> In my opinion, this is a trivial algorithm.

Simple algorithms are they also algorithms?
https://github.com/TheAlgorithms/Go/blob/master/math/pythagoras/pythagoras.go - there is also a simple algorithm here. And I also wanted to add that this is just a blank in which it is planned to add more complex algorithms.

raklaptudirm:
By trivial, I mean the following:
- It is extremely simple.
- It does not serve any important purpose.
- There is nothing to understand in it.

I will of course wait for the other maintainers opinions.

siriak (commented):
I wouldn't include that since it's too simple. Not that it's not an algorithm, but we cannot include everything because it'll be difficult to navigate the repository if we do so. Most people won't look for an implementation of this algorithm since they can write one by themselves in no time.

siriak:
Seems that the maintainers are on the same page on this issue, so I'm closing this PR. Sorry that this algorithm wasn't accepted, have a good day, and I hope to see contributions that align with our values from you!

PR #358, opened 2021-09-27
Is power of two

i-redbyte:
Checks if a number is a power of two

### DIFF
--- a/math/binary/checkisnumberpoweroftwo_test.go
+++ b/math/binary/checkisnumberpoweroftwo_test.go
@@ -61,11 +61,11 @@ func TestIsPowerOfTwoLeftShift(t *testing.T) {
 	}
 }
 
-func TestIsPowerOfTwoUseLog(t *testing.T) {
+func TestIsPowOfTwoUseLog(t *testing.T) {
 	tests := getTestsForPowerOfTwo()
 	for _, tv := range tests {
 		t.Run(tv.name, func(t *testing.T) {
-			result := math2.IsPowerOfTwoUseLog(float64(tv.a))
+			result := math2.IsPowOfTwoUseLog(float64(tv.a))
 			t.Log(tv.a, " ", result)
 			if result != tv.missing {
 				t.Errorf("Wrong result! Expected:%v, returned:%v ", tv.missing, result)
@@ -88,6 +88,6 @@ func BenchmarkIsPowerOfTwoUseCycleAndLeftShift(b *testing.B) {
 
 func BenchmarkIsPowerOfTwoUseLog(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		math2.IsPowerOfTwoUseLog(1024)
+		math2.IsPowOfTwoUseLog(1024)
 	}
 }
--- a/math/binary/checkisnumberpoweroftwo_test.go
+++ b/math/binary/checkisnumberpoweroftwo_test.go
@@ -6,6 +6,7 @@
 package binary
 
 import (
+	math2 "github.com/TheAlgorithms/Go/math"
 	"math"
 	"testing"
 )
@@ -64,7 +65,7 @@ func TestIsPowerOfTwoUseLog(t *testing.T) {
 	tests := getTestsForPowerOfTwo()
 	for _, tv := range tests {
 		t.Run(tv.name, func(t *testing.T) {
-			result := IsPowerOfTwoUseLog(float64(tv.a))
+			result := math2.IsPowerOfTwoUseLog(float64(tv.a))
 			t.Log(tv.a, " ", result)
 			if result != tv.missing {
 				t.Errorf("Wrong result! Expected:%v, returned:%v ", tv.missing, result)
@@ -87,6 +88,6 @@ func BenchmarkIsPowerOfTwoUseCycleAndLeftShift(b *testing.B) {
 
 func BenchmarkIsPowerOfTwoUseLog(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		IsPowerOfTwoUseLog(1024)
+		math2.IsPowerOfTwoUseLog(1024)
 	}
 }
--- a/math/binary/checkisnumberpoweroftwo_test.go
+++ b/math/binary/checkisnumberpoweroftwo_test.go
@@ -3,7 +3,7 @@
 // author(s) [red_byte](https://github.com/i-redbyte)
 // see checkisnumberpoweroftwo.go
 
-package binarymath
+package binary
 
 import (
 	"math"
@@ -34,11 +34,11 @@ func getTestsForPowerOfTwo() []struct {
 	return tests
 }
 
-func TestIsPowerOfTwoBinaryMethod(t *testing.T) {
+func TestIsPowerOfTwo(t *testing.T) {
 	tests := getTestsForPowerOfTwo()
 	for _, tv := range tests {
 		t.Run(tv.name, func(t *testing.T) {
-			result := IsPowerOfTwoBinaryMethod(tv.a)
+			result := IsPowerOfTwo(tv.a)
 			t.Log(tv.a, " ", result)
 			if result != tv.missing {
 				t.Errorf("Wrong result! Expected:%v, returned:%v ", tv.missing, result)
@@ -47,11 +47,11 @@ func TestIsPowerOfTwoBinaryMethod(t *testing.T) {
 	}
 }
 
-func TestIsPowerOfTwoUseCycleAndLeftShift(t *testing.T) {
+func TestIsPowerOfTwoLeftShift(t *testing.T) {
 	tests := getTestsForPowerOfTwo()
 	for _, tv := range tests {
 		t.Run(tv.name, func(t *testing.T) {
-			result := IsPowerOfTwoUseCycleAndLeftShift(uint(tv.a))
+			result := IsPowerOfTwoLeftShift(uint(tv.a))
 			t.Log(tv.a, " ", result)
 			if result != tv.missing {
 				t.Errorf("Wrong result! Expected:%v, returned:%v ", tv.missing, result)
@@ -75,13 +75,13 @@ func TestIsPowerOfTwoUseLog(t *testing.T) {
 
 func BenchmarkIsPowerOfTwoBinaryMethod(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		IsPowerOfTwoBinaryMethod(1024)
+		IsPowerOfTwo(1024)
 	}
 }
 
 func BenchmarkIsPowerOfTwoUseCycleAndLeftShift(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		IsPowerOfTwoUseCycleAndLeftShift(1024)
+		IsPowerOfTwoLeftShift(1024)
 	}
 }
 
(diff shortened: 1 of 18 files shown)


math/binarymath/checkisnumberpoweroftwo_test.go
```diff
@@ -0,0 +1,92 @@
...
+	name    string
+	a       int
+	missing bool
+} {
+	var tests = []struct {
+		name    string
+		a       int
+		missing bool
+	}{
+		{"64 is power of two? - YES", 64, true},
```
siriak, line 23: "Is 64 a power of 2?" would be grammatically better :)

siriak (approved):

math/binarymath/checkisnumberpoweroftwo_test.go
i-redbyte: Done

math/binarymath/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,37 @@
...
+	return false
+}
+
+func IsPowerOfTwoUseLog(number float64) bool {
+	if number == 0 || math.Round(number) == math.MaxInt64 {
+		return false
+	}
+	log := math.Log2(number)
+	pow := math.Pow(2, math.Round(log))
+	return pow == number
```
raklaptudirm, line 36: A better way will just be to check if the log is an integer or not:
```go
return log == math.Round(log)
```

siriak (approved):
Thanks!

math/binarymath/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,37 @@
...
+// see checkisnumberpoweroftwo_test.go
+
+package binarymath
+
+import (
+	"math"
+)
+
+func IsPowerOfTwoBinaryMethod(x int) bool {
+	return x > 0 && (x&(x-1)) == 0
```
raklaptudirm, line 15: Please add some explanatory comments as many people will not be familiar with how bit-wise operations work or how numbers are represented in binary.

An example of what you could write:
```go
// This function uses the fact that powers of 2 are represented
// like 10...0 in binary, and numbers one less than the power of 2
// are represented like 11...1.
// Therefore, using the and function:
//    10...0
//  & 01...1
//    00...0 -> 0
// This is also true for 0, which is not a power of 2, for which we
// have to add and extra condition.
```

raklaptudirm (changes requested):

math/binarymath/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,37 @@
+// checkisnumberpoweroftwo.go
+// description: Is the number a power of two
+// details:
+// Checks if a number is a power of two- [Power of two](https://en.wikipedia.org/wiki/Power_of_two)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see checkisnumberpoweroftwo_test.go
+
+package binarymath
```
tjgurwara99, line 8: This is me nitpicking but could you change the name of this package?

`math/binarymath` sounds odd - it would be much better to have `math/binary`. Don't you think?

math/binarymath/checkisnumberpoweroftwo.go
tjgurwara99: I also agree with @raklaptudirm on this 😄

math/binarymath/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"math"
+)
+
+func IsPowerOfTwoBinaryMethod(x int) bool {
```
tjgurwara99, line 23: Can you choose a more representative name? `IsPowerOfTwoBinaryMethod` has actually too many things for it to be easily understandable - to the likes of someone like me, 😅.

I think the a`*BinaryMethod` can be removed from suffix and `IsPowerOfTwo` is enough for it to be clear - note `binary` is in the package name.

math/binarymath/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	for p := uint(1); p > 0; p = p << 1 {
+		if number == p {
+			return true
+		}
+	}
+	return false
+}
+
+func IsPowerOfTwoUseLog(number float64) bool {
```
tjgurwara99, line 39: Should it really be in this package - this function doesn't have a single bitwise operation. Unless, by `binary` in the package name you mean the binary operations then that means you can include almost every algorithm.

math/binarymath/checkisnumberpoweroftwo.go
i-redbyte: Great note! Fixed, thanks!

math/binarymath/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"math"
+)
+
+func IsPowerOfTwoBinaryMethod(x int) bool {
+	return x > 0 && (x&(x-1)) == 0
+}
+
+func IsPowerOfTwoUseCycleAndLeftShift(number uint) bool {
```
tjgurwara99, line 27:
```suggestion
func IsPowerOfTwoLeftShift(number uint) bool {
```

math/binarymath/checkisnumberpoweroftwo_test.go
```diff
@@ -0,0 +1,92 @@
...
+// see checkisnumberpoweroftwo.go
+
+package binarymath
+
+import (
+	"math"
+	"testing"
+)
+
+func getTestsForPowerOfTwo() []struct {
```
tjgurwara99, line 13: The tests are immutable, so you don't really need to create a factory here. But it doesn't matter too much so even if you don't change it its fine.

tjgurwara99 (changes requested):
Very good use of bitwise operations. I would recommend you try to make the function names as succinct as possible without losing the readability.

math/binarymath/checkisnumberpoweroftwo.go
i-redbyte: good idea

math/binarymath/checkisnumberpoweroftwo.go
i-redbyte: This method was added by me to compare the performance of the binary approach and the conventional

i-redbyte: but if this method seems to be superfluous to you in this package, maybe it is worth leaving a comment that it was added for performance comparison? Or I can just delete it, it doesn't matter to me

math/binarymath/checkisnumberpoweroftwo.go
raklaptudirm: I forgot to mention this, but add an explanation for this too.

raklaptudirm (changes requested):

math/binarymath/checkisnumberpoweroftwo_test.go
i-redbyte: left it as it is

tjgurwara99 (approved):
LGTM 😄

tjgurwara99:
I've marked it as approved but I still think the log function should be moved to the parent package

siriak (dismissed):
LGTM

math/checkisnumberpoweroftwo.go
```diff
@@ -0,0 +1,16 @@
+package math
+
+import (
+	"math"
+)
+
+// IsPowerOfTwoUseLog This function checks if a number is a power of two using the logarithm.
+// The limiting degree can be from 0 to 63.
+// See alternatives in the binary package.
+func IsPowerOfTwoUseLog(number float64) bool {
```
tjgurwara99, line 10: `IsPowOfTwoUsingLog`

tjgurwara99 (changes requested):
Small change 😅 everything else is perfect 😄

i-redbyte:
> Wouldn't it be better if we created a new sub-package under the `math` package, and put the files their? What are your opinions?

New subpackage for one function? I think this is redundant

tjgurwara99:
> Wouldn't it be better if we created a new sub-package under the `math` package, and put the files their? What are your opinions?

Could you elaborate on what you mean? If you are talking about moving just one function, then I think its not worth it - creates a lot of clutter. But if you mean to move all power of two functions in a sub package then I can see the benefits. Either way, could you elaborate? @raklaptudirm

raklaptudirm:
I was saying that instead of moving the log based algorithm to another file as you had suggested, we could create a new sub-package under `math` and put all the functions there.
```
math/
 └─ powersOfTwo/
     └─ checkPowerOfTwo.go
     └─ checkPowerOfTwo_test.go
```

i-redbyte:
> I was saying that instead of moving the log based algorithm to another file as you had suggested, we could create a new sub-package under `math` and put all the functions there.
> 
> ```
> math/
>  └─ powersOfTwo/
>      └─ checkPowerOfTwo.go
>      └─ checkPowerOfTwo_test.go
> ```

I am generally satisfied with both approaches, so that as most people decide, so I will do

siriak (approved):
I'm fine with either approach too :)

tjgurwara99:
Ah ok, I guess this is a question of opinion so majority would rule. But I would say that the binary operation functions are good inside the binary directory but the algorithm that uses log doesn't necessitate a sub package in my opinion.

I guess my point would be that it depends on whether there would be other algorithms that would be added to the directory - personally I don't think there are any algorithms other than these algorithms that would go inside this subpackage - which is not a problem but it creates another layer of subdirectory for learners/explorers to browse through. We should try to keep the package structure as close to flat as possible unless there is a specific reason for creating separate package.

So I guess my vote would be a no to creating a subdirectory/subpackage for it 😅. But if @siriak and @raklaptudirm disagree then go ahead with a sub package approach 😄

i-redbyte:
> Ah ok, I guess this is a question of opinion so majority would rule. But I would say that the binary operation functions are good inside the binary directory but the algorithm that uses log doesn't necessitate a sub package in my opinion.
> 
> I guess my point would be that it depends on whether there would be other algorithms that would be added to the directory - personally I don't think there are any algorithms other than these algorithms that would go inside this subpackage - which is not a problem but it creates another layer of subdirectory for learners/explorers to browse through. We should try to keep the package structure as close to flat as possible unless there is a specific reason for creating separate package.
> 
> So I guess my vote would be a no to creating a subdirectory/subpackage for it 😅. But if @siriak and @raklaptudirm disagree then go ahead with a sub package approach 😄

Yes, it looks like the subdirectory will be redundant

math/binary/checkisnumberpoweroftwo_test.go
```diff
@@ -0,0 +1,93 @@
...
+	for i := 0; i < b.N; i++ {
+		IsPowerOfTwoLeftShift(1024)
+	}
+}
+
+func BenchmarkIsPowerOfTwoUseLog(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		math2.IsPowOfTwoUseLog(1024)
+	}
+}
```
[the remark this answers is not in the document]
i-redbyte: Done

i-redbyte:
> Could you move the test for the UseLog function to the parent math folder? do this dependent on whether other maintainers agree with my previous suggestion.

Done

i-redbyte:
So what do we decide in the end, gentlemen?

tjgurwara99 (approved):

siriak:
@raklaptudirm waiting on you :)

raklaptudirm (approved):
I was away for a bit, so could not reply earlier. All good. I am ok with the directory structure.

PR #360, opened 2021-09-29
Change binary tree traversal implementation and add the test

FemiNoviaLina:
Hi, I changed the binary tree traversal implementation so it return the traversal instead of print it. I also added the test. Please consider to review this change. Thank you.


siriak (approved):
LGTM, please resolve conflicts :)

raklaptudirm (approved):

PR #366, opened 2021-10-01
Math improvements

ivanvc:
Hi,

I was reading the code from the math algorithms and found some improvements:

1. In Moser–De Bruijn sequence, there's an if to check if the number is even, then another check if it was odd, for an integer (which is the variable type), that check seems rather unnecessary
2. In Monte Carlo Pi's method, there was an unnecessary cast to `float64` of the literal `4`, which is not required
3. As I was already editing files, I noticed that `bitwiseMax.go` had a wrong file name according to the guidelines, so I updated it too

It's my first time contributing to this project. I hope it meets the standards, else I can address any issues.


math/pi/montecarlopi.go
```diff
@@ -22,6 +22,6 @@ func MonteCarloPi(randomPoints int) float64 {
 			inside += 1
 		}
 	}
-	pi := float64(inside) / float64(randomPoints) * float64(4)
+	pi := float64(inside) / float64(randomPoints) * 4
```
raklaptudirm, line 25:
```suggestion
	pi := float64(inside) / randomPoints * 4
```
Dividing the `float64` number with `randomPoints` will automatically result in a `float64` number.

raklaptudirm (changes requested):

math/pi/montecarlopi.go
ivanvc: Oh yeah, you're right

raklaptudirm (approved):

math/pi/montecarlopi.go
raklaptudirm: Pardon me, I forgot golang does not have implicit type conversions, I have fixed that in the latest commit.

raklaptudirm (approved):

tjgurwara99 (approved):

PR #368, opened 2021-10-02
Recursion to find the exponent

thearkamitra:


tjgurwara99 (changes requested):
Please read our contribution guidelines before opening a PR.

recursion/exponent.go
```diff
@@ -0,0 +1,19 @@
+package main
```
tjgurwara99, line 1: Specifically the part where it says the functions should be self contained and not in a main package.

tjgurwara99 (changes requested):

recursion/exponent.go
tjgurwara99, line 1: Change the package name to something meaningful and not main

recursion/exponent.go
```diff
@@ -0,0 +1,19 @@
...
+
+import "fmt"
+
+func exponentialrecursion(x int, p int) int {
```
tjgurwara99, line 5: Export this function and write a test for this

tjgurwara99 (changes requested):

siriak (approved):
Agree with previous comments, but apart from that looks good

PR #369, opened 2021-10-02
Adjust instructions in CONTRIBUTING.md

### LINKED ISSUE #359: the "Code Formatter" section of CONTRIBUTING.md just contains go installation instructions
**Describe the bug**

In the file CONTRIBUTING.md, there is a section with the heading  "Code Formatter".  This section describes how to install go, and it has build instructions that may be out of date since they refer to the nonexistent file "main.go"

**To Reproduce**
1. View the file https://github.com/TheAlgorithms/Go/blob/master/CONTRIBUTING.md

**Expected behavior**

- I would expect this section to contain instructions something like "run the command go fmt".
- I would expect the existing content to be in an earlier section of the document with a header like "Installing Go"
- It might be helpful for beginners if there was a unified section of the document with sub-sections for installing go, running the existing code, formatting code, etc.  The section on preparing pull requests could then just reference the earlier section.

SignorMercurio:
Add a section about preparation for pull request in order to fix #359 .

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,6 +168,8 @@ Common prefixes:
   - Manual Installation: [Downloads - The Go Programming Language](https://golang.org/dl/)
 - Running (all platforms): `go run filexyz.go`
 
+> Note: New packages should not be `main`, and never implement `main` functions for any package.
+
 #### Code formatter
 To format your code, you can use the gofmt tool directly:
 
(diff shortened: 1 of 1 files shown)


tjgurwara99 (dismissed):
While you're at it can you also add an instruction saying that new packages should not be main and never to implement main functions for a package.

SignorMercurio:
> While you're at it can you also add an instruction saying that new packages should not be main and never to implement main functions for a package.

I added a note section. Could you please check if it's ok ?

raklaptudirm (approved):

tjgurwara99 (approved):
LGTM! Thanks for contributing.

PR #371, opened 2021-10-03
feat: reverse bits

i-redbyte:
Reverse the bits of an integer number


math/binary/reversebits.go
```diff
@@ -0,0 +1,19 @@
...
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see reversebits_test.go
+
+package binary
+
+func ReverseBits(number uint) uint {
+	result := uint(0)
+	powerTwo := 31
+	for number != 0 {
+		result += (number & 1) << powerTwo
```
raklaptudirm, line 14: Please add an explanation in the comments for the algorithm.

raklaptudirm (changes requested):

math/binary/reversebits.go
i-redbyte: Done

raklaptudirm (approved):

math/binary/reversebits.go
```diff
@@ -0,0 +1,23 @@
...
+
+package binary
+
+// ReverseBits This function initialized the result by 0 (all bits 0) and process the given number starting
+// from its least significant bit. If the current bit is 1, set the corresponding most significant bit in the result
+// and finally move on to the next bit in the input number.
+// Repeat this till all its bits are processed.
+func ReverseBits(number uint) uint {
+	result := uint(0)
+	powerTwo := 31
```
tjgurwara99, line 16: I don't like the name here 😅 

maybe this is more suitable
```suggestion
	intSize := 31
```

or perhaps

```suggestion
	intShift := 31
```

tjgurwara99 (dismissed):
Everything else is good. I'm approving it so its fine if you don't change the name.

tjgurwara99 (approved):

Issue #272, opened 2021-02-18
Algorithms that have no tests


cclauss:
From our GitHub Actions...
`go test $(go list ./... | grep -v /sbom | grep -v /rsa)`
* [ ] ciphers	[no test files]
* [ ] datastructures/binary-tree	[no test files]
* [ ] datastructures/dynamic-array	[no test files]
* [ ] datastructures/linkedlist/doublylinkedlist	[no test files]
* [ ] graphs/breathfirstsearch	[no test files]
* [ ] graphs/depthfirstsearch	[no test files]
* [ ] graphs/floydwarshall	[no test files]
* [ ] other/maxsubarraysum	[no test files]
* [ ] other/nestedbrackets	[no test files]
* [ ] other/passwordgenerator	[no test files]
* [ ] other/stringcombinations	[no test files]
* [ ] strings/single-string-matching/bom	[no test files]
* [ ] strings/single-string-matching/horspool	[no test files]

tjgurwara99:
This is an updated list of missing tests:

* [ ]       github.com/TheAlgorithms/Go/data_structures/binary_tree [no test files]
* [ ]       github.com/TheAlgorithms/Go/data_structures/set [no test files]
* [ ]       github.com/TheAlgorithms/Go/genetic_algorithm   [no test files]
* [ ]       github.com/TheAlgorithms/Go/graphs/depth_first_search   [no test files]
* [ ]       github.com/TheAlgorithms/Go/graphs/floyd_warshall       [no test files]
* [ ]       github.com/TheAlgorithms/Go/math/lcm    [no test files]
* [ ]       github.com/TheAlgorithms/Go/other/max_subarray_sum      [no test files]
* [ ]       github.com/TheAlgorithms/Go/other/nested_brackets       [no test files]
* [ ]       github.com/TheAlgorithms/Go/other/password_generator    [no test files]
* [ ]       github.com/TheAlgorithms/Go/other/string_combinations   [no test files]
* [ ]       github.com/TheAlgorithms/Go/strings/single_string_matching/bom  [no test files]
* [ ]       github.com/TheAlgorithms/Go/strings/single_string_matching/horspool     [no test files]

tjgurwara99:
We need tests for the following algorithms:

* [ ]     github.com/TheAlgorithms/Go/strings/bom [no test files] 
* [ ]     github.com/TheAlgorithms/Go/strings/combination [no test files]
[...]

PR #373, opened 2021-10-05
test: add TestFloydWarshall

calebeof:
As mentioned in Issue #272, there are no tests for FloydWarshall's algorithm, so this PR has the intention to add test cases for it. 

Besides the additional test cases, I've changed one check in original algorithm: the check of "matrix's width is different than the matrix's height" was only checking the first line. Now, it is checking all lines. 

Suggestions for additional edge cases are appreciated! :)

### DIFF
--- a/graph/floydwarshall_test.go
+++ b/graph/floydwarshall_test.go
@@ -1,0 +1,109 @@
+package graph
+
+import (
+	"math"
+	"testing"
+)
+
+//noConnection is a value (INF) which represents that there are no two edges that directly connects these two nodes
+var noConnection = math.Inf(1)
+
+func compareDist(a, b float64) bool {
+	const float64EqualityThreshold = 1e-9
+	return math.Abs(a-b) <= float64EqualityThreshold
+}
+
+func equalMatrices(a, b [][]float64) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i := range a {
+		if len(a[i]) != len(b[i]) {
+			return false
+		}
+
+		for j := range a[i] {
+			if a[i][j] == noConnection && b[i][j] == noConnection {
+				continue
+			}
+
+			if !compareDist(a[i][j], b[i][j]) {
+				return false
+			}
+		}
+	}
+
+	return true
+}
+
+func TestFloydWarshall(t *testing.T) {
+	var floydWarshallTestData = []struct {
+		description string
+		graph       [][]float64
+		expected    [][]float64
+	}{
+		{
+			description: "test empty graph",
+			graph:       nil,
+			expected:    nil,
+		},
+		{
+			description: "test graph with wrong dimensions",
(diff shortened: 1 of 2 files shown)


graph/floydwarshall_test.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"math"
+	"testing"
+)
+
+//noConnection is a value (INF) which represents that there are no two edges that directly connects these two nodes
+var noConnection = math.Inf(1)
+
+func compareDist(a, b float64) bool {
+	const float64EqualityThreshold = 1e-9
```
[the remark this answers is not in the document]
tjgurwara99: Plus this is taken verbatim – except for the function name – from stack overflow. Its only a few lines but if you take extract of source code from a source, cite it.

graph/floydwarshall_test.go
```diff
@@ -0,0 +1,109 @@
+package graph
+
+import (
+	"math"
+	"testing"
+)
+
+//noConnection is a value (INF) which represents that there are no two edges that directly connects these two nodes
+var noConnection = math.Inf(1)
```
tjgurwara99, line 9: This is commented out in the actual algorithm file - I think you can remove this and instead write it so that the one in the other file is exported in that package. Also, I would just call it `Inf` and not `noConnection` or `maxValue` as its written in the other file.

graph/floydwarshall_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func compareDist(a, b float64) bool {
+	const float64EqualityThreshold = 1e-9
+	return math.Abs(a-b) <= float64EqualityThreshold
+}
+
+func equalMatrices(a, b [][]float64) bool {
```
tjgurwara99, line 16: I think this should be turned into a method instead - this would ensure two things:
1. hard to misuse
2. easier to read and understand. 

My suggestion is to create a separate type

```Go
type WeightedGraph [][]float64
```

Then change the Floyd Warshall algorithm accordingly and finally create the method like so

```Go
func (w *WeightedGraph) IsAlmostEqualTo(b WeightedGraph) bool {
    ...
}
```

tjgurwara99 (changes requested):
Please consider the changes that I've requested.

graph/floydwarshall_test.go
calebeof: Thanks for that, removed both `maxValue` and `noConnection` and added `Inf`.

graph/floydwarshall_test.go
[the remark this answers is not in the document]
calebeof: Thanks a lot, created a reference :)

graph/floydwarshall_test.go
calebeof: Thanks!! Made this type and added this method. Can you please take a look? :)

PR #374, opened 2021-10-06
feat: kruskal's algorithm using DSU

viveknathani:
Hi guys! I saw that Kruskal's algorithm was missing from the list. So I wrote an implementation using the disjoint set union data structure. Also, it would be amazing if you could add the `hacktoberfest-accepted` label to it. 

References:
1. https://cp-algorithms.com/data_structures/disjoint_set_union.html
2. https://cp-algorithms.com/graph/mst_kruskal_with_dsu.html

### DIFF
--- a/graph/kruskal.go
+++ b/graph/kruskal.go
@@ -1,0 +1,115 @@
+package graph
+
+import (
+	"sort"
+)
+
+// edge describes the edge of a weighted graph
+type edge struct {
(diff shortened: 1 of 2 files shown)


raklaptudirm (changes requested):
Please add some links for reference.

raklaptudirm (approved):

PR #378, opened 2021-10-09
Create bitonicSort.go

chinifoxtrot:
Bitonic sort in golang


chinifoxtrot:
Thanks for the detailed explanation. Will keep these in mind.

PR #380, opened 2021-10-10
Feat: Add word removal to the Go/struct/trie

xavier268:
Added word removal, either lazily (default) or with tree restructuring (see Compact).
Added Capacity and Size.
Refactored and more detailled tests, examples and benchmarks
Fixed a bug in the previous Find method logic (not using the isLeaf value !)
Some minor comments and an example added to the Set go structure.

Xavier


xavier268:
File names changed.

PR #381, opened 2021-10-10
Add an example (and a warning) to Go/structures/Set

xavier268:
The Set, as it is currently implemented, cannot accept a Type that is not an acceptable Type for a map key.
Adding a slice, for instance, will panic.

I just added a (valid) example and a package level warning.

### DIFF
--- a/structure/set/set_example_test.go
+++ b/structure/set/set_example_test.go
@@ -1,0 +1,24 @@
+// package set implements a Set using a golang map.
+// This implies that only the types that are accepted as valid map keys can be used as set elements.
+// For instance, do not try to Add a slice, or the program will panic.
+//
+package set
+
+import (
+	"fmt"
+)
+
+func ExampleSet() {
+
+	set := New(1, 2, 3)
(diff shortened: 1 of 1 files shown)


xavier268:
Thanks Taj, your response. I initially tried to seperate the minor point on Go/structure/set from the more significant suggested points in Go/structures/trie by merging from a separate branches, but obviously, I messed up ... I'm not familiar with pull requests at this stage, so I will try to merge everything as a single PR.
Hope that works for you,

Xavier

xavier268:
Closed and merge (hopefully ...) with previous PR.

tjgurwara99:
Whenever you're working on two different things, it's better to create branches locally in your machine. Usually after a fresh clone of your report, if you do `git checkout master` which would open the master branch and then create a new branch from master by typing `git checkout -b <new-branch-name>` it would create and open a new branch which would be relative to master. Now, if you need to work on something else, checkout master branch again and create a new branch with the above command and work from there. This ensures that both your work don't interfere with each other.

PR #384, opened 2021-10-11
feat: Add prime factorization

ayoubc:


siriak:
Adding more algorithms for factorization would be nice, but probably need to do a different PR for that

PR #391, opened 2021-10-14
feat: Decimal to Binary Conversion

motasimmakki:
## Description
Algorithm to convert any Decimal number into its equivalent Binary representation.


motasimmakki:
@tjgurwara99 
Kindly have a look!

PR #397, opened 2021-10-18
test: eliminate duplication of test cases for sort

lostbearlabs:
The test framework and benchmark framework for sorts were both duplicated because `QuickSort` had a different function signature than the other sorting functions.

This PR modifies `QuickSort` to use the same signature as all the other sorts by introducing a new helper function `QuickSortRange` to handle the case of sorting only part of the input array.  With this change, the test and benchmark for `QuickSort` can use the same logic as all the other sorts.


lostbearlabs:
Would it be possible for a maintainer to approve the pending workflow for me?

PR #398, opened 2021-10-19
feat: Binary to Decimal Conversion

motasimmakki:
## Description:
Algorithm to convert Binary number into its equivalent Decimal number.


motasimmakki:
> I just noticed that the previous PR that you opened for decimal to binary I asked you to write the functions as convert and inverse (thinking that they were in their own package I made a mistake). Could you also correct those two functions. Rename Convert to `DecimalToBinary` and `Inverse` to `Reverse` in the strings package NOT in the convert package.

Kindly have a look!

PR #400, opened 2021-10-19
feat: bipartite checking implementation

ongspxm:


ongspxm:
right it does sound more elegant to reuse the Graph struct, i have made the necessary modifications

PR #411, opened 2021-10-21
test: improve coverage on hashmap package

mcaci:
Hello, I'm creating a separate PR to improve the coverage of the hashmap package in order to test the resize of the hashmap. Hope this can help!
Cheers,
Michele

### DIFF
--- a/sort/countingsort.go
+++ b/sort/countingsort.go
@@ -1,0 +1,25 @@
+// countingsort.go
+// description: Implementation of counting sort algorithm
+// details: A simple counting sort algorithm implementation
+// author [Phil](https://github.com/pschik)
+// see sort_test.go for a test implementation, test function TestQuickSort
+
+// package sort provides primitives for sorting slices and user-defined collections
+package sort
+
+func Count(data []int) []int {
+	var aMin, aMax = -1000, 1000
+	count := make([]int, aMax-aMin+1)
+	for _, x := range data {
+		count[x-aMin]++
+	}
+	z := 0
+	for i, c := range count {
+		for c > 0 {
+			data[z] = i + aMin
+			z++
+			c--
+		}
+	}
+	return data
+}
--- a/sort/sorts_test.go
+++ b/sort/sorts_test.go
@@ -23,28 +23,24 @@ func testFramework(t *testing.T, sortingFunction func([]int) []int) {
 			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
 			name:     "Reversed Unsigned",
 		},
-
 		//Sorted slice
 		{
 			input:    []int{-10, -9, -8, -7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
 			expected: []int{-10, -9, -8, -7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
 			name:     "Sorted Signed",
 		},
-
 		//Reversed slice
 		{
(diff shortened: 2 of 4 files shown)


raklaptudirm (dismissed):

mcaci:
Hi all, do you have any comments about my answer to the review? Does it make sense to you? I'm ready to revert my changes as in the review, but first I wanted to discuss about my reply to see in which direction we should go with the code

tjgurwara99:
@mcaci I don't see any replies

mcaci:
I updated the comment where i wrote my reply with your handle, maybe it helps finding it. If not I just copy/paste it below. Let me know if it worked.

tjgurwara99:
@mcaci I still don't see any comments on my feed

mcaci:
I see, no problem, I will paste it here just below:

Well, it depends much on the context where HashMap is created and used.
In the current situation, inside the same package, we agree that we can use both New and HashMap{...} to initialize its fields.
However, since these fields are unexported, if we want to call it from an external package we can't modify the fields; they too should be exported to be able to be modified from an external package (in the example of the Server struct in the net/http package this is the case).

So I added the NewWith to be able to customize the HashMap from outside the package, I changed the package inside hashmap_test.go to dynamic_test to test the code from outside the package.

I can revert these change to have test and code in the same package but it's good that we had this conversation around the NewWith. Let me know your feedbacks and preferences on what to do and I will change accordingly

tjgurwara99:
> However, since these fields are unexported, if we want to call it from an external package we can't modify the fields; they too should be exported to be able to be modified from an external package

I guess this has become a question of design choice. I think you were right in pointing out that the fields are not exported - I missed that previously so my earlier point about the `net/http` doesn't make sense in this context so I apologise 😅 

I think - after reading through the code thoroughly - you are trying to mimic what `make` keyword does in Go and I think you should replace `NewWith` to `Make` and change the function signature to return an instance of `Hashmap` not a pointer to a `Hashmap` i.e. `func Make(size uint64, capacity uint64) Hashmap` (or something along those lines). I make this distinction because the `new` keyword in the language returns a pointer to an object and `make` returns an instance of the object (map/slice) passed with the desired capacity and size. Thus I think it makes sense to follow this pattern - of course this is more of a design question and if you have a suggestion and want to argue against this, just tag me. I'm more than happy to discuss `Go` topics and idioms 😄

mcaci:
Hi @tjgurwara99, thanks for your feedback. It's good that we discussed about this, it was helpful to learn for many things. I see your point in my function `NewWith` to behave similarly to `make` and I agree with your suggestion to change it to `Make` and return the struct itself. I have done the update, thanks again for sharing your thoughts!

tjgurwara99:
I think you've overwritten everything that you did in the PR and rebased to a previous state. Try to revert your last commits and I'll see if it gets reverted back. This happens because you seem to rebase your source instead of committing line by line but try to be careful because you can loose a lot of information by doing that or try not to rebase them to previous states for your commits.

mcaci:
Hi, yes I didn't see it and I did a bad mistake on this PR. I created a new one (#435) to start fresh. I apologise for this mistake. I learned to be more careful next time to see that I don't overwrite anything.

tjgurwara99:
> Hi, yes I didn't see it and I did a bad mistake on this PR. I created a new one (#435) to start fresh. I apologise for this mistake. I learned to be more careful next time to see that I don't overwrite anything.
> 
> If you're ok I will close this one and keep the review going on #435.

That's fine - I'll look at it now. If this keeps happening to you, its worth looking at your `.gitconfig` in your home directory - this is your global configurations and you can configure it to however you use it 😄 there might be a setting that you might have set for `commit` flag for `git`.

PR #416, opened 2021-10-24
Implementation of Red Black Tree

VinWare:
Red Black Trees are a self-balancing variant of Binary Search Trees. They maintain the simple `O(log(n)` search times while also ensuring that the tree remains balanced, and never stretches out into a list, regardless of the order of input. This implementation is based on Introduction to Algorithms, Cormen et. al. 3rd edition.


raklaptudirm (dismissed):

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+	Root *Node
+}
+
+func (t *RBTree) Insert(Val int) {
+	t.insert(Val)
+	t.Root.isRed = false
+}
+
+// Internal insert function, refer Intro to Algorithms mentioned above
+func (t *RBTree) insert(Val int) {
```
tjgurwara99, line 46: Could you break this function down into separable pieces? Currently its responsibility is to insert a node but its finding the empty node to act as parent (which can be taken out of this function as a function itself) and doing so many things that its not readable (and indeed understandable) at all. Taking care of different cases can easily be handled by a strategy pattern instead of creating complicated if and else statements. Additionally, it seems that you have just re-written code from this book you mentioned and translated to Go. Keep in mind that some of the concepts in other languages make it difficult to write algorithms in Go and if you don't adapt the algorithm itself it will be very messy. Right now, the algorithm is so convoluted that it took me more than 50mins to properly understand this - which shouldn't be the case because Red Black tree is essentially a Binary search tree and they are supposed to be easy to understand.

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+}
+
+// Simply return new node
+func newNode(Val int, parent *Node) *Node {
+	return &Node{Val: Val, left: nil, right: nil, parent: parent, isRed: true}
+}
+
+// Public function to print preorder
+func (t *RBTree) PrintPreorder() {
```
tjgurwara99, line 175:
```suggestion
func (t *RBTree) PrintPreOrder() {
```

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+	var p = preorder(t.Root)
+	fmt.Printf("Length %d\n", len(p))
+	for _, a := range p {
+		fmt.Printf("%d %v\n", a.Val, a.isRed)
+	}
+}
+
+// Public function to print inorder
+func (t *RBTree) PrintInorder() {
```
tjgurwara99, line 184:
```suggestion
func (t *RBTree) PrintInOrder() {
```

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+	var p = inorder(t.Root)
+	fmt.Printf("Length %d\n", len(p))
+	for _, a := range p {
+		fmt.Printf("%d %v\n", a.Val, a.isRed)
+	}
+}
+
+// Returns inorder representation
+func inorder(currNode *Node) []*Node {
```
tjgurwara99, line 193:
```suggestion
func inOrder(currNode *Node) []*Node {
```

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	ret = append(ret, inorder(currNode.left)...)
+	ret = append(ret, currNode)
+	ret = append(ret, inorder(currNode.right)...)
+	return ret
+}
+
+// Returns preorder representation
+func preorder(currNode *Node) []*Node {
```
tjgurwara99, line 206:
```suggestion
func preOrder(currNode *Node) []*Node {
```

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+		return nil
+	}
+	if node.Val == val {
+		return node
+	}
+	if val < node.Val {
+		return search(node.left, val)
+	}
+	return search(node.right, val)
+}
```
tjgurwara99, line 234: Just put it inside the Search function. No need to create an intermediate function if there are no more than 1 search implementation. Creating an intermediate function should only be considered if there is more than one way to do one thing and you want to give the user choice of choosing which one to do. Right now, you only have one search and it is highly unlikely anyone is going to add more searching algorithms on this.

structure/redblacktree/rbtree.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	return curr
+}
+
+func (t *RBTree) fix(curr *Node) {
+	if curr == nil {
+		// Nothing to fix
+		return
+	}
+	for !t.isRoot(curr) && !curr.isRed {
```
tjgurwara99, line 331: Try to break this down into different bits because again, because of the sheer number of if statements it has taken me more than 40mins to understand this.

structure/redblacktree/node.go
```diff
@@ -0,0 +1,23 @@
+package redblacktree
+
+// A representation of a tree node
+// Consists of Val (currently only int allowed)
+// left, right and parent are pointers to the relevant nodes
+// isRed is a representation of the color, made as bool
+type Node struct {
```
tjgurwara99, line 7: Call it RBNode instead.
```suggestion
type RBNode struct {
```

tjgurwara99 (changes requested):
IMO, looking at the wikipedia - I personally think their pseudo code can easily be implemented in Go.

structure/redblacktree/rbtree.go
VinWare: Yeah I was on the fence about allowing Search on any Node.. might as well add it and allow subtree search

structure/redblacktree/rbtree.go
VinWare: Completed changes

structure/redblacktree/rbtree.go
VinWare: Completed changes

structure/redblacktree/rbtree.go
VinWare: Completed changes

structure/redblacktree/node.go
VinWare: Completed changes

structure/redblacktree/rbtree.go
VinWare: I'd still prefer to keep both a simple tree-wide search, that only takes int and another search that takes a node. If you want I'll expose `search` as `SearchNode` or `SearchSubTree`.

tjgurwara99: Yeah, if you have more than one implementation of search then it is feasible that you can create an intermediate one. Note that you can utilise the power of Go by using Function as an argument. Assuming you have two different search functions with the same signature (example `searchABC(node *Node, val int) *Node` and `searchDEF(node *Node, val int) *Node` then you can have your exposed Search implementation be such that it accepts the function as an argument but since they are simple and your Search function is doing nothing but calling searchABC or searchDEF then it is fine to just expose these two functions instead of a search function. Makes it clear what the intent of the function is.

VinWare: I have just made `search` public as `SearchFromNode` - basically allowing search within a subtree. As it uses the most basic implementation of search, I am not subdividing it further. Simple `Search` just runs this function from the root node

structure/redblacktree/rbtree.go
VinWare: Insert is now broken down into - finding a parent (`findNewParent`, the main logic (setting Root if needed else setting parent's children node values), and the tree rebalancing part (`rebalanceTree`). I have rewritten `rotate` to avoid duplicating code between left and right rotates, and because of this, I can remove the separate`if-else` statements for left and right. Now there are just one level of `if-else` statements in the `climbTreeColoring` method which I feel are simpler than a strategy pattern

siriak (dismissed):

structure/redblacktree/rbtree_test.go
```diff
@@ -0,0 +1,107 @@
+package redblacktree
```
tjgurwara99, line 1: Could I ask to make this one `redblacktree_test` instead. I want to see the implementation where only exposed api is tested because currently I think even your RBNode would not be working as expected when you make this change. Try it.

tjgurwara99: I can't even think of how a RBTree will be made outside of this package.

tjgurwara99 (changes requested):

structure/redblacktree/rbtree_test.go
VinWare: The only issue is that since more methods/functions are exposed, I will have to write tests for them, will add them in an upcoming commit

structure/redblacktree_test/rbtree_test.go
```diff
@@ -0,0 +1,108 @@
...
+	t.Run("Test 2: Delete existing item", func(t *testing.T) {
+		for _, val := range vals {
+			tree2.Insert(val)
+		}
+		tree2.Delete(delete2)
+		if !compare(tree2, deleteExpPreVal, deleteExpPreRed, deleteExpInVal, deleteExpInRed) {
+			t.Error("Insert did not work")
+		}
+	})
+}
```
tjgurwara99, line 108: What about the test for search algorithm?

tjgurwara99 (changes requested):

structure/redblacktree_test/rbtree_test.go
```diff
@@ -0,0 +1,108 @@
...
+	t.Run("Test 1: Insert values", func(t *testing.T) {
+		for _, val := range vals {
+			tree.Insert(val)
+		}
+		if !compare(tree, expPreVal, expPreRed, expInVal, expInRed) {
+			t.Error("Insert did not work")
+		}
+	})
+}
+func TestSearch(t *testing.T) {
```
VinWare, line 50: That's the second test, `TestSearch`. It's above the one you've linked here

structure/redblacktree_test/rbtree_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return true
+}
+func TestInsert(t *testing.T) {
+	var tree = &RBTree{}
+	var vals = []int{8, 18, 5, 15, 17, 25, 40, 80}
+	var expPreVal = []int{17, 8, 5, 15, 25, 18, 40, 80}
+	var expPreRed = []bool{false, true, false, false, true, false, false, true}
+	var expInVal = []int{5, 8, 15, 17, 18, 25, 40, 80}
+	var expInRed = []bool{false, true, false, false, false, true, false, true}
```
tjgurwara99, line 45: This is why I don't like this implementation. You have to keep track of more than 5 different indices just to compare the result!!

tjgurwara99: You have two data structures `RBNode` and `RBTree` but there is no need for `RBTree` right? You are not using root in any way which doesn't directly use the `RBNode`. In fact, you are just using `RBNode` in everything defined for the tree. In this implementation, the responsibility of RBTree struct depends directly on the implementation details of RBNode. The first principle in SOLID design pattern is to avoid such things. You can ask yourself a question, do you really need `RBTree`? In every method that you define for `RBTree` you are using the fields of `RBNode` continuously because these two are so linked together that you cannot decouple them - not good if you are trying to make production applications. Note that trees by nature are naturally recursive in nature but because of your `RBTree` implementation, you made this very difficult to do recursion on it. Anyhow, I recommend you to look at the implementation of Red Black Tree in our python repo - [here](https://github.com/TheAlgorithms/Python/blob/master/data_structures/binary_tree/red_black_tree.py)

Anyhow this is my suggestion as I recognise that you have spend quite a bit of time with this and if you think that my suggestion here is unreasonable, just comment here. I can then ask the other maintainers for their opinions on this and since it seems that they had previously approved it, they might consider your implementation to be fine if they think it can be merged, then I will agree with them. Hope these suggestions help.

structure/redblacktree_test/rbtree_test.go
tjgurwara99: Seems like I went blind here 😓

structure/redblacktree/rbtree.go
VinWare: Will look at this post-November

PR #424, opened 2021-10-31
Create README.md

kartikay23:
Created readme file for sorting

### DIFF
--- a/sort/README.md
+++ b/sort/README.md
@@ -1,0 +1,14 @@
+### Sorting
+ Sorting Algorithm is used to rearrange a given array or list elements according to a comparison operator on the elements.
+ The comparison operator is used to decide the new order of element in the respective data structure.
+ 
+ ## Time complexities:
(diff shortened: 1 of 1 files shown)


raklaptudirm:
Documentation is provided in comments in my opinion.

kartikay23:
I create readme file for sorting as an overview about it. People can see it as an overview of the topic if they do not want yo go side each file.

raklaptudirm:
I will wait for the opinion of the other maintainers before approving this.

tjgurwara99:
I also think this is unnecessary - people can just go to [this](https://pkg.go.dev/github.com/TheAlgorithms/Go@v0.0.2-alpha/sort) and get a much better overview of all the algorithms. I don't mind keeping this readme though either but overtime (from my experience I know that) this will get out of date very quickly.

siriak:
It seems that there are better options than adding readme, but thanks anyway!

PR #425, opened 2021-10-31
insertion sort using go

akshayprog:


raklaptudirm:
We already have an implementation [here](https://github.com/TheAlgorithms/Go/blob/master/sort/insertionsort.go).

PR #429, opened 2021-11-04
feat: Godocmd action added for updating readme automatically.

tjgurwara99:
Apologies for taking long - was quite busy.

I have made a few changes to the workflows to better segregate them when multistep actions fail. Gives us more information as to where they fail etc.

I tried to make the godocmd fast but it still takes 7/8s because of all the additional sorting and changing a map to slice that it ended up like this 😓. But the result is still good I think.


README.md
```diff
@@ -12,109 +12,831 @@ The repository is a collection of open-source implementation of a variety of alg
 Read our [Contribution Guidelines](CONTRIBUTING.md) before you contribute.
 
 ## List of Algorithms
+<!--- GODOCMD BEGIN --->
```
tjgurwara99, line 15: NOTE: This is necessary for my godocmd action to work otherwise it will append to the file.

README.md
```diff
@@ -12,109 +12,831 @@ The repository is a collection of open-source implementation of a variety of alg
...
+
+---
+##### Functions:
+
+1. [`Decrypt`](./cipher/xor/xor.go#L19):  Decrypt decrypts with Xor encryption
+2. [`Encrypt`](./cipher/xor/xor.go#L10):  Encrypt encrypts with Xor encryption after converting each character to byte The returned value might not be readable because there is no guarantee which is within the ASCII range If using other type such as string, []int, or some other types, add the statements for converting the type to []byte.
+
+---
+</details>
+<!--- GODOCMD END --->
```
tjgurwara99, line 842: same applies here otherwise it will overwrite everything after `<!--- GODOCMD BEGIN --->`

raklaptudirm (dismissed):

tjgurwara99:
@siriak @cclauss could you also please review this when you get some time?

siriak (dismissed):
Looks good, great job!

siriak (approved):

.github/workflows/golang_lint_and_test.yml
```diff
@@ -8,14 +8,17 @@ jobs:
...
         with:
           ignore_words_list: "actualy,nwe"
           skip: "go.mod,go.sum"
-      - uses: actions/setup-go@v2
-      - run: go version
-      - uses: golangci/golangci-lint-action@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+      - name: Run Golang CI Lint
+        uses: golangci/golangci-lint-action@v2
```
cclauss, line 19: I am not an active member of this Go repo so you guys decide.

But as a personal pet peeve, I tend not to `name:` job steps so that when they fail, I can quickly see the actual `uses:` or `run:` text rather than some vague English equivalent.

cclauss (approved):

.github/workflows/golang_lint_and_test.yml
tjgurwara99: Ah ok. I'm personally of the opinion that if the steps are orthogonal (or derived) then they should run under their own title to segregate between other steps and help with finding where things went wrong (and is beginner friendly imo)

cclauss: I do agree to small, discreet steps so that the contributor knows exactly which command failed.

raklaptudirm: Looks like @tjgurwara99 named each command separately, not a number of commands together. So I think naming them is ok.

PR #430, opened 2021-11-08
Feat/catalan number

i-redbyte:


math/catalan/catalannumber.go
```diff
@@ -0,0 +1,16 @@
+// catalannumber.go
```
raklaptudirm, line 1: Add a small explanation of what is a Catalan number.

raklaptudirm (changes requested):

math/catalan/catalannumber.go
i-redbyte: Done

raklaptudirm (dismissed):

math/catalan/catalannumber_test.go
```diff
@@ -0,0 +1,40 @@
...
+import "testing"
+
+func TestCatalanNumber(t *testing.T) {
+	type args struct {
+		n int
+	}
+	tests := []struct {
+		name string
+		args args
+		want int
```
tjgurwara99, line 17: There is only one argument right? Why did you make a struct to contain n when you can just have this?

```suggestion
	tests := []struct {
		name string
		arg  int
		want int
```

tjgurwara99 (changes requested):

math/catalan/catalannumber_test.go
i-redbyte: Done

tjgurwara99 (approved):

raklaptudirm (approved):

PR #437, opened 2021-11-15
feat: add Conversion/base64 algorithm

paul-leydier:
A basic implementation of the base64 encoding algorithm, based on [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648#section-4).

### DIFF
--- a/cipher/base64/base64.go
+++ b/cipher/base64/base64.go
@@ -7,12 +7,15 @@
 // see base64_test.go
 package base64
 
+import "strings" // Used for efficient string builder (more efficient than simply appending strings)
+
 const Alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
 
-// Encode encodes the received input bytes into a base64 string.
+// Encode encodes the received input bytes slice into a base64 string.
 // The implementation follows the RFC4648 standard, which is documented
 // at https://datatracker.ietf.org/doc/html/rfc4648#section-4
 func Encode(input []byte) (encoded string) {
+	var sb strings.Builder
 	// If not 24 bits (3 bytes) multiple, pad with 0 value bytes, and with "=" for the output
 	var padding string
 	for i := len(input) % 3; i > 0 && i < 3; i++ {
@@ -24,18 +27,21 @@ func Encode(input []byte) (encoded string) {
 	// encode 24 bits per 24 bits (3 bytes per 3 bytes)
 	for i := 0; i < len(input); i += 3 {
 		// select 3 8-bit input groups, and re-arrange them into 4 6-bit groups
+		// the literal 0x3F corresponds to the byte "0011 1111"
+		// the operation "byte & 0x3F" masks the two left-most bits
 		group := [4]byte{
 			input[i] >> 2,
-			(input[i]<<6)>>2 + input[i+1]>>4,
-			(input[i+1]<<4)>>2 + input[i+2]>>6,
-			(input[i+2] << 2) >> 2,
+			(input[i]<<4)&0x3F + input[i+1]>>4,
+			(input[i+1]<<2)&0x3F + input[i+2]>>6,
+			input[i+2] & 0x3F,
 		}
 
 		// translate each groups into a char using the static map
 		for _, b := range group {
-			encoded = encoded + string(Alphabet[int(b)])
+			sb.WriteString(string(Alphabet[int(b)]))
 		}
 	}
+	encoded = sb.String()
 
 	// Apply the output padding
 	encoded = encoded[:len(encoded)-len(padding)] + padding[:]
--- a/cipher/base64/base64.go
+++ b/cipher/base64/base64.go
@@ -17,7 +17,7 @@ const Alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789
 // Encode encodes the received input bytes slice into a base64 string.
 // The implementation follows the RFC4648 standard, which is documented
 // at https://datatracker.ietf.org/doc/html/rfc4648#section-4
-func Encode(input []byte) (encoded string) {
+func Encode(input []byte) string {
 	var sb strings.Builder
 	// If not 24 bits (3 bytes) multiple, pad with 0 value bytes, and with "=" for the output
 	var padding string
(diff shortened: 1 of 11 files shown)


paul-leydier:
Hi everyone!

You'll find in my pull request an implementation of base64 encoding/decoding according to RFC4648.

Here are the benchmark results on my computer:
```
goos: windows
goarch: amd64
pkg: github.com/TheAlgorithms/Go/cipher/base64
cpu: Intel(R) Core(TM) i5-9600K CPU @ 3.70GHz
BenchmarkEncode
BenchmarkEncode/Hello_World!
BenchmarkEncode/Hello_World!-6         	 7218757	       161.4 ns/op
BenchmarkEncode/Hello_World!a
BenchmarkEncode/Hello_World!a-6        	 4290879	       281.8 ns/op
BenchmarkEncode/Hello_World!ab
BenchmarkEncode/Hello_World!ab-6       	 4653878	       264.1 ns/op
BenchmarkEncode/Empty
BenchmarkEncode/Empty-6                	100000000	        10.60 ns/op
BenchmarkEncode/6
BenchmarkEncode/6-6                    	11310638	       104.9 ns/op
BenchmarkEncode/Lorem_ipsum
BenchmarkEncode/Lorem_ipsum-6          	  260875	      4933 ns/op
BenchmarkDecode
BenchmarkDecode/Hello_World!
BenchmarkDecode/Hello_World!-6         	10344987	       117.4 ns/op
BenchmarkDecode/Hello_World!a
BenchmarkDecode/Hello_World!a-6        	 8759992	       138.0 ns/op
BenchmarkDecode/Hello_World!ab
BenchmarkDecode/Hello_World!ab-6       	 8687486	       137.6 ns/op
BenchmarkDecode/Empty
BenchmarkDecode/Empty-6                	223022956	         5.307 ns/op
BenchmarkDecode/6
BenchmarkDecode/6-6                    	32433220	        38.15 ns/op
BenchmarkDecode/Lorem_ipsum
BenchmarkDecode/Lorem_ipsum-6          	  338247	      3498 ns/op
PASS
```

cipher/base64/base64.go
```diff
@@ -0,0 +1,80 @@
+// Package base64
+// base64.go
+// description: The base64 encodage algorithm as defined in the RFC4648 standard.
+// author: [Paul Leydier] (https://github.com/paul-leydier)
+// ref: https://datatracker.ietf.org/doc/html/rfc4648#section-4
+// ref: https://en.wikipedia.org/wiki/Base64
+// see base64_test.go
+package base64
```
tjgurwara99, line 8:
```suggestion
// base64.go
// description: The base64 encoding algorithm as defined in the RFC4648 standard.
// author: [Paul Leydier] (https://github.com/paul-leydier)
// ref: https://datatracker.ietf.org/doc/html/rfc4648#section-4
// ref: https://en.wikipedia.org/wiki/Base64
// see base64_test.go

// Package base64 encodes input as base64 as defined in the RFC4648 standard.
package base64
```

cipher/base64/base64.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"strings" // Used for efficient string builder (more efficient than simply appending strings)
+)
+
+const Alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+
+// Encode encodes the received input bytes slice into a base64 string.
+// The implementation follows the RFC4648 standard, which is documented
+// at https://datatracker.ietf.org/doc/html/rfc4648#section-4
+func Encode(input []byte) (encoded string) {
```
tjgurwara99, line 19:
```suggestion
func Encode(input []byte) string {
```

Try to avoid the predefined returns.

If you still want to use it then return the variable explicitly like so

```Go
func Something() (val string) {
...
    return val
}
```

tjgurwara99 (changes requested):
Other than that everything else seems to be okay.

cipher/base64/base64.go
paul-leydier: resolved in https://github.com/TheAlgorithms/Go/pull/437/commits/3ff3e7887f47a7f2e8793bde8d3df9550479dd7d

paul-leydier:
@tjgurwara99 Thanks for the review, useful feedback! I updated the pull request accordingly.

tjgurwara99 (changes requested):
Thinking about this, I don't think base64 is a cipher. Its a conversion. Can you move this to the conversion package instead.

paul-leydier:
> Thinking about this, I don't think base64 is a cipher. Its a conversion. Can you move this to the conversion package instead.

Indeed I thought about it but saw on the [The algorithms](https://the-algorithms.com/search?q=base64) website that it was categorized as a Cipher for both C++ and Python (although it is categorized as "conversions" for js), so I used the cipher package for consistency.

But I can definitely move it to the conversion package if you confirm?

tjgurwara99:
Yeah, move it to conversion package. Its more appropriate there.

paul-leydier:
I moved the code to the conversion package and renamed the function to be more precise, since the package name is not "base64" anymore.

raklaptudirm (approved):

raklaptudirm:
@tjgurwara99 The `README` generation is failing.

tjgurwara99:
Yeah, I'm waiting on the #434 to be merged and see if the changes work. If not we need to create a token specifically for the godocmd - I've asked @siriak for it but am waiting to hear back from him.

conversion/base64.go
```diff
@@ -0,0 +1,81 @@
+// base64.go
+// description: The base64 encoding algorithm as defined in the RFC4648 standard.
+// author: [Paul Leydier] (https://github.com/paul-leydier)
+// ref: https://datatracker.ietf.org/doc/html/rfc4648#section-4
+// ref: https://en.wikipedia.org/wiki/Base64
+// see base64_test.go
+
+package conversion
```
siriak, line 8:
```suggestion
package conversion1
```

conversion/base64.go
```diff
@@ -0,0 +1,81 @@
+// base64.go
+// description: The base64 encoding algorithm as defined in the RFC4648 standard.
+// author: [Paul Leydier] (https://github.com/paul-leydier)
+// ref: https://datatracker.ietf.org/doc/html/rfc4648#section-4
+// ref: https://en.wikipedia.org/wiki/Base64
+// see base64_test.go
+
+package conversion1
```
siriak, line 8:
```suggestion
package conversion
```

siriak (approved):

tjgurwara99 (approved):

PR #438, opened 2021-11-16
Optimize performance

iGwkang:
BenchmarkEncrypt

goos: windows
goarch: amd64
pkg: github.com/TheAlgorithms/Go/cipher/xor
cpu: Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
BenchmarkEncrypt
BenchmarkEncrypt-6   	 2252124	       526.5 ns/op
PASS

---------
BenchmarkOldEncrypt

goos: windows
goarch: amd64
pkg: github.com/TheAlgorithms/Go/cipher/xor
cpu: Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
BenchmarkOldEncrypt
BenchmarkOldEncrypt-6   	  905516	      1193 ns/op
PASS

### DIFF
--- a/cipher/xor/xor.go
+++ b/cipher/xor/xor.go
@@ -2,24 +2,58 @@
 // ref: https://en.wikipedia.org/wiki/XOR_cipher
 package xor
 
+import "unsafe"
+
 // Encrypt encrypts with Xor encryption after converting each character to byte
 // The returned value might not be readable because there is no guarantee
 // which is within the ASCII range
 // If using other type such as string, []int, or some other types,
 // add the statements for converting the type to []byte.
 func Encrypt(key byte, plaintext []byte) []byte {
-	cipherText := []byte{}
-	for _, ch := range plaintext {
-		cipherText = append(cipherText, key^ch)
+	cipherText := make([]byte, len(plaintext))
+	copy(cipherText, plaintext)
+
+	uint64Plaintext := *(*[]uint64)(unsafe.Pointer(&cipherText))
+
+	uint64Key := uint64(key) +
+		uint64(key)<<(8*1) +
+		uint64(key)<<(8*2) +
+		uint64(key)<<(8*3) +
+		uint64(key)<<(8*4) +
+		uint64(key)<<(8*5) +
+		uint64(key)<<(8*6) +
+		uint64(key)<<(8*7)
+
+	n := len(cipherText) / 8
+
+	// less than 64byte of data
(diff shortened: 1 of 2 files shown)


siriak (approved):

cipher/xor/xor.go
```diff
@@ -2,24 +2,58 @@
 // ref: https://en.wikipedia.org/wiki/XOR_cipher
 package xor
 
+import "unsafe"
```
tjgurwara99, line 5: I don't understand why you have to use unsafe here?

cipher/xor/xor.go
```diff
@@ -2,24 +2,58 @@
...
 // Encrypt encrypts with Xor encryption after converting each character to byte
 // The returned value might not be readable because there is no guarantee
 // which is within the ASCII range
 // If using other type such as string, []int, or some other types,
 // add the statements for converting the type to []byte.
 func Encrypt(key byte, plaintext []byte) []byte {
-	cipherText := []byte{}
-	for _, ch := range plaintext {
-		cipherText = append(cipherText, key^ch)
+	cipherText := make([]byte, len(plaintext))
```
tjgurwara99, line 13: Write your implementation in a separate function if it is a different algorithm from the one already implemented.

tjgurwara99 (changes requested):

cipher/xor/xor.go
iGwkang: The reason for using unsafe is to convert []byte to []uint64, byte xor byte is far less efficient than uin64 xor uint64.

cipher/xor/xor.go
iGwkang: The xor package is to xor each byte in the []byte with the key, and my implementation is designed to do the same.

cipher/xor/xor.go
tjgurwara99: I see that there is an open issue in Golang about this. I don't know why this is the case as byte is a type alias for uint8.

cipher/xor/xor.go
tjgurwara99: Yes but the previous implementation is also correct. I have no evidence to see why you implementation is better therefore keep the old implemenation and write your implementation in a new function instead of changing the already created (well working) function. Afterwards write benchmarks to test the optimised performances of both implementation that way anyone reading the code can see why your implementation is better/worse than the other implementation.

PR #442, opened 2021-11-16
Adding program to check if the number is an Armstrong number

kavithajm:
image

### DIFF
--- a/math/armstrong/isarmstrong.go
+++ b/math/armstrong/isarmstrong.go
@@ -19,18 +19,13 @@ func IsArmstrong(number int) bool {
 	// to get the number of digits in the number
 	length := float64(len(strconv.Itoa(number)))
 
-	// get the right most digit
-	for {
+	// get the right most digit and break the loop once all digits are iterated
+	for tempNum > 0 {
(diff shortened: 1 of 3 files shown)


siriak (approved):

math/armstrong/isarmstrong.go
```diff
@@ -0,0 +1,37 @@
...
+	// get the right most digit
+	for {
+		rightMost = tempNum % 10
+		sum += int(math.Pow(float64(rightMost), length))
+
+		// update the input digit minus the processed rightMost
+		tempNum /= 10
+
+		// break the loop if all numbers are iterated
+		if tempNum == 0 {
```
raklaptudirm, line 31: Just use a `for [condition] {}` loop instead of using a manual break.

raklaptudirm (changes requested):

raklaptudirm (approved):

math/armstrong/isarmstrong.go
kavithajm: Resolved

tjgurwara99 (approved):

PR #446, opened 2021-11-22
Fibonacci with matrix

i-redbyte:
Get Fibonacci number use matrix method


tjgurwara99 (dismissed):
LGTM.

This matrix method is essentially equivalent to

$ F_n =  \frac{\phi^n - \xi^n}{\phi - \xi} $

where $\phi =  \frac{1 + \sqrt{5}}{2} $ and $\xi = \frac{1 - \sqrt{5}}{2} $ (\xi is golden ratio).

Look up difference equations (not differential equations) and you will see how this is equivalent to what you have done.

siriak (approved):

tjgurwara99:
> @tjgurwara99 it's stuck again. Any ideas on how we can make it work?

Unless the PR I raised gets merged I can't do anything unfortunately 😓 Are you sure that the GoDocMD action has been disabled @siriak ?

siriak:
Hmm, I thought we've already merged #444

PR #451, opened 2021-12-05
Feat: Luhn algorithm

i-redbyte:


checksum/luhn.go
```diff
@@ -0,0 +1,26 @@
...
+		if i%2 == 0 {
+			number *= 2
+			if number > 9 {
+				number -= 9
+			}
+		}
+		result += number
+	}
+	return result%10 == 0
+}
```
tjgurwara99, line 26: Early returns are better than nested `if` statements and try to keep the conversion to the user - keep the algorithm as simple as possible.

```suggestion
func LuhnAlgorithm(s []rune) bool {
	n := len(s)
	var number rune
	var result rune
	for i := 0; i < n; i++ {
		number = s[i] - '0'
		if i%2 != 0 {
			result += number
			continue
		}
		number *= 2
		if number > 9 {
			number -= 9
		}
		result += number
	}
	return result%10 == 0
}
```

tjgurwara99 (changes requested):

checksum/luhn.go
i-redbyte: As for the early return, I agree, I corrected it, but I do not completely agree about the second part, IMHO we are operating in essence with numerical sequences and for this algorithm the array of runes, in my opinion, is redundant, the string is perfect here, in my opinion, but I'm ready to discuss this moment is with you.

tjgurwara99: It's up to you here, very much a design preference. However, from my experience, numerical sequences should remain numerical in nature, that is rune is numerical type whereas a string is not a numerical type and rudimentary operations are much better operated using rune and byte. Consider this same algorithm in C++ or Java, you would not be using strings there but rather the char type because of the extra overhead strings have, same applies here imo but again, up to you.

i-redbyte: > 

changed the string to an array of runes

tjgurwara99 (approved):

siriak (approved):
LGTM, thanks!

PR #452, opened 2021-12-08
Graphs articulation points

anirudnits:
- [x] Implement algorithm for identifying articulation points
- [x] Add comments
- [x] Add tests and ensure that the tests are passing
- [x] Passing lint checks


jyzkm:
邮件已收到，祝好！

siriak (approved):

graph/articulationpoints.go
```diff
@@ -0,0 +1,58 @@
...
+	dfs(graph, 0, -1)
+
+	if child_cnt[0] == 1 {
+		is_ap[0] = false // if the root has only one child, it is not an articulation point
+	}
+
+	return is_ap
+}
+
+func dfs(graph Graph, u int, parent int) {
```
tjgurwara99, line 36: There is already a depth first search algorithm within this repository - in fact within this package. I recommend that you stick to using the ones that are already there.

tjgurwara99 (changes requested):

graph/articulationpoints.go
anirudnits: @tjgurwara99 so the depth first search implemented here is a bit different from the dfs defined in the graphs folder. For instance, as part of the dfs I used for this use-case, I am assigning and updating the `discovery_time` of the vertices which is not part of the dfs already implemented, moreover manipulating other variables like `earliest_discovery`, `time`, `child` and `parent` are again not included in a typical dfs. Either I go and tweak the dfs implementation and include all these extraneous parameters or should I chnage the number of the function to something like `ArticulationPointHelper` to avoid confusion with a normal dfs?

tjgurwara99: > @tjgurwara99 so the depth first search implemented here is a bit different from the dfs defined in the graphs folder.

Exactly why I mentioned it here, calling it `dfs` is not ideal because dfs doesn't change any state in the graph.

> Either I go and tweak the dfs implementation and include all these extraneous parameters or should I change the name of the function to something like `ArticulationPointHelper` to avoid confusion with a normal dfs?

No you shouldn't tweak the already implemented dfs as your implementation is not purely dfs but rather an application of it. I suggest to call this `articulationPointHelper` and including a documentation comment of why its needed and its basis (based on dfs in this case).

anirudnits: @tjgurwara99 sure, I'll do that.

graph/articulationpoints.go
```diff
@@ -0,0 +1,62 @@
...
+	ArticulationPointHelper(graph, 0, -1)
+
+	if child_cnt[0] == 1 {
+		is_ap[0] = false // if the root has only one child, it is not an articulation point
+	}
+
+	return is_ap
+}
+
+func ArticulationPointHelper(graph Graph, u int, parent int) {
```
tjgurwara99, line 36: Make it unexported - unlikely to be used other than your own function. Exporting it pollutes the API.

```suggestion
func articulationPointHelper(graph Graph, u int, parent int) {
```

tjgurwara99 (changes requested):

graph/articulationpoints.go
```diff
@@ -0,0 +1,62 @@
...
+	earliest_discovery[u] = discovery_time[u] // set the earliest discovery time to the discovered time
+	time++                                    // increment the time
+
+	for v := range graph.edges[u] {
+		if v == parent {
+			continue
+		}
+		if !visited[v] {
+			child_cnt[u]++
+			ArticulationPointHelper(graph, v, u)
```
tjgurwara99, line 53: I don't think this compiles!

tjgurwara99 (changes requested):

graph/articulationpoints.go
```diff
@@ -0,0 +1,62 @@
...
+import "github.com/TheAlgorithms/Go/math/min"
+
+var (
+	time               int    // time variable to keep track of the time of discovery_time of a vertex
+	is_ap              []bool // boolean array to mark articulation points
+	visited            []bool // boolean array to mark visited vertices
+	child_cnt          []int  // array to store the number of children of a vertex
+	discovery_time     []int  // array to store the time of discovery of a vertex
+	earliest_discovery []int  // array to store the earliest discovered vertex reachable from a vertex
+)
```
tjgurwara99, line 16: Avoid global variables. Do you really have to use global variables??

tjgurwara99 (changes requested):

graph/articulationpoints.go
anirudnits: Apologies, I made the change in haste and didn't verify if its working, I'll make the necessary changes.

graph/articulationpoints.go
anirudnits: @tjgurwara99  So, I tried to do this using a nested function. Basically I wanted to make the helper function nested inside the main function, however because the helper function needs to be recursive I couldn't maneuver around this. The other option is to pass all these global variables as function parameters but that lead to a lot of function parameters. The workaround for this was making a struct and passing the struct in functions. Should I go ahead with the latter approach? Or is there a more subtle way to get this done?

~ Aniruddha

tjgurwara99: I don't see a problem with having a few extra arguments to the function. The struct already exists as well, you can embed and create your own struct but I think that is not ideal for a general purpose graph. So I recommend going with the argument approach but if I remember correctly, only your helper function would have to have all of these parameters and not your main function so, I don't see a problem with the argument approach as the helper function is not an exported function.

anirudnits: @tjgurwara99 I've made the changes, have a look and let me know.

~ Aniruddha

tjgurwara99: Hey! Sorry, been a bit busy at work. Will check it tonight and let you know. From a quick glance it seems that the implementation is good. The only issue I have is with your style of commenting - in Go there is an auto documentation friendly commenting style which is used. It seems you are not aware of it. I would suggest adding proper Go styled comments. [Read this for more information](https://go.dev/blog/godoc)

graph/articulationpoints.go
```diff
@@ -0,0 +1,53 @@
...
+	articulationPointHelper(graph, 0, -1, &time, is_ap, visited, child_cnt, discovery_time, earliest_discovery)
+
+	if child_cnt[0] == 1 {
+		is_ap[0] = false // if the root has only one child, it is not an articulation point
+	}
+
+	return is_ap
+}
+
+func articulationPointHelper(graph Graph, vertex int, parent int, time *int, is_ap []bool, visited []bool, child_cnt []int, discovery_time []int, earliest_discovery []int) {
```
tjgurwara99, line 27: This is actually very confusing haha.

tjgurwara99: Are you comfortable with pointers? If so, I recommend creating a package scoped (unexported) struct (with the first letter lowercase) with these fields like so


```Go

type apHelper struct {
	graph              Graph
	vertex             int
	parent             int
	time               *int
	is_ap              []bool
	visited            []bool
	child_cnt          []int
	discovery_time     []int
	earliest_discovery []int
}
```

This will make it easier to recursively call the function if you change the definition to something like this

```Go
func apHelperFunc(ap *apHelper) {
	...
	// do the operations to the ap.graph or ap.parent or ap.visited etc
	// and it will be reflected to the main one.
}
```

With the above change you will also need to change the `ArticulationPoint` function to something like this

```Go
func ArticulationPoint(graph Graph) []bool {
	//initialize all the variables
	ap := apHelper{
		time:               0,
		is_ap:              make([]bool, graph.vertices),
		visited:            make([]bool, graph.vertices),
		child_cnt:          make([]int, graph.vertices),
		discovery_time:     make([]int, graph.vertices),
		earliest_discovery: make([]int, graph.vertices),
	}
	
	articulationPointHelper(&ap)
```

and that should do it.

That is IMO much cleaner than having it as arguments.

Let me know what you think.

tjgurwara99 (changes requested):

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+func ArticulationPoint(graph Graph) []bool {
+	time := 0 // time variable to keep track of the time of discovery_time of a vertex
+
+	//initialize all the variables
+	apHelperInstance := &apHelper{
+		is_ap:              make([]bool, graph.vertices), // boolean array to mark articulation points
+		visited:            make([]bool, graph.vertices), // boolean array to mark visited vertices
+		child_cnt:          make([]int, graph.vertices),  // array to store the number of children of a vertex
+		discovery_time:     make([]int, graph.vertices),  // array to store the time of discovery of a vertex
+		earliest_discovery: make([]int, graph.vertices),  // array to store the earliest discovered vertex reachable from a vertex
```
tjgurwara99, line 26: Some of these comments are unhelpful. Comments should describe the behaviour not the action. Anyone who can read code would know that for example `is_ap` is a slice (array - you call it) of bool. That is not really helpful. Instead if you want to comment, you should comment on why instead of what/how.

graph/articulationpoints.go
```diff
@@ -0,0 +1,53 @@
+// Algorithm to identify articulation points in a connected graph
+// You can read more about articulation points here: https://en.wikipedia.org/wiki/Biconnected_component
+// and the implementation and explanation here: https://cptalks.quora.com/Cut-Vertex-Articulation-point
+
+package graph
+
+import "github.com/TheAlgorithms/Go/math/min"
+
+func ArticulationPoint(graph Graph) []bool {
```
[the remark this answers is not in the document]
tjgurwara99: Have you seen above ^ suggestion? That is how to comment go code - Go style comments for documenting code in places like go.dev or in our in house readme documentation tool. See [here](https://pkg.go.dev/github.com/TheAlgorithms/Go) for example.

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	return apHelperInstance.is_ap
+}
+
+func articulationPointHelper(apHelperInstance *apHelper, vertex int, parent int, time *int, graph Graph) {
+	// A recursive function to identify articulation points in a graph
+	// based on the depth first search transversal of the graph, however modified to keep track
+	// and update the child_cnt, discovery_time and earliest_discovery arrays defined above
+
```
tjgurwara99, line 41: This should go above the function (Go style comments) rather than inside the function.

tjgurwara99 (changes requested):

graph/articulationpoints.go
```diff
@@ -0,0 +1,53 @@
...
+			child_cnt[vertex]++
+			articulationPointHelper(graph, next_vertex, vertex, time, is_ap, visited, child_cnt, discovery_time, earliest_discovery)
+			earliest_discovery[vertex] = min.Int(earliest_discovery[vertex], earliest_discovery[next_vertex])
+			if earliest_discovery[next_vertex] >= discovery_time[vertex] {
+				is_ap[vertex] = true
+			}
+		} else {
+			earliest_discovery[vertex] = min.Int(earliest_discovery[vertex], discovery_time[next_vertex])
+		}
+	}
```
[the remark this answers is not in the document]
tjgurwara99: I don't think you have applied this either.

graph/articulationpoints.go
```diff
@@ -14,16 +10,21 @@ type apHelper struct {
 	earliest_discovery []int
 }
 
+// ArticulationPoint is a function to identify articulation points in a graph.
+// The function takes the graph as an argument and returns a boolean slice which indicates whether a vertex is an articulation point or not.
```
tjgurwara99, line 14:
```suggestion
// The function takes the graph as an argument and returns a boolean slice
// which indicates whether a vertex is an articulation point or not.
```

graph/articulationpoints.go
```diff
@@ -14,16 +10,21 @@ type apHelper struct {
...
 	apHelperInstance := &apHelper{
-		is_ap:              make([]bool, graph.vertices), // boolean array to mark articulation points
-		visited:            make([]bool, graph.vertices), // boolean array to mark visited vertices
-		child_cnt:          make([]int, graph.vertices),  // array to store the number of children of a vertex
-		discovery_time:     make([]int, graph.vertices),  // array to store the time of discovery of a vertex
-		earliest_discovery: make([]int, graph.vertices),  // array to store the earliest discovered vertex reachable from a vertex
+		is_ap:              make([]bool, graph.vertices),
+		visited:            make([]bool, graph.vertices),
+		child_cnt:          make([]int, graph.vertices),
+		discovery_time:     make([]int, graph.vertices), // integer slice to store the discovery time of a vertex as we traverse the graph in a depth first manner
```
tjgurwara99, line 26:
```suggestion
		// integer slice to store the discovery time of a vertex as we traverse
		// the graph in a depth first manner
		discovery_time:     make([]int, graph.vertices),
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+	earliest_discovery []int
+}
+
+// ArticulationPoint is a function to identify articulation points in a graph.
+// The function takes the graph as an argument and returns a boolean slice which indicates whether a vertex is an articulation point or not.
+// Worst Case Time Complexity: O(|V| + |E|)
+// Auxiliary Space: O(|V|)
+// reference: https://en.wikipedia.org/wiki/Biconnected_component and https://cptalks.quora.com/Cut-Vertex-Articulation-point
+func ArticulationPoint(graph Graph) []bool {
+	time := 0 // time variable to keep track of the time of discovery_time of a vertex
```
tjgurwara99, line 19:
```suggestion
	// time variable to keep track of the time of discovery_time of a vertex
	time := 0
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+
+	//initialize all the variables
+	apHelperInstance := &apHelper{
+		is_ap:              make([]bool, graph.vertices),
+		visited:            make([]bool, graph.vertices),
+		child_cnt:          make([]int, graph.vertices),
+		discovery_time:     make([]int, graph.vertices), // integer slice to store the discovery time of a vertex as we traverse the graph in a depth first manner
+		earliest_discovery: make([]int, graph.vertices), // integer slice to store the earliest discovered vertex reachable from a vertex
```
tjgurwara99, line 27:
```suggestion
		// integer slice to store the earliest discovered vertex reachable from a vertex
		earliest_discovery: make([]int, graph.vertices),
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+	if apHelperInstance.child_cnt[0] == 1 {
+		apHelperInstance.is_ap[0] = false // if the root has only one child, it is not an articulation point
+	}
+
+	return apHelperInstance.is_ap
+}
+
+// articulationPointHelper is a recursive function to traverse the graph and mark articulation points.
+// Based on the depth first search transversal of the graph, however modified to keep track
+// and update the child_cnt, discovery_time and earliest_discovery slices defined above
```
tjgurwara99, line 40: You keep lines very long, make it easier for the reader (and by extension the reviewers - us) by breaking them down.

```suggestion
// articulationPointHelper is a recursive function to traverse the graph
// and mark articulation points. Based on the depth first search transversal
// of the graph, however modified to keep track and update the
// `child_cnt`, `discovery_time` and `earliest_discovery` slices defined above.
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+func articulationPointHelper(apHelperInstance *apHelper, vertex int, parent int, time *int, graph Graph) {
+	apHelperInstance.visited[vertex] = true
```
tjgurwara99, line 42:
```suggestion
func articulationPointHelper(
	apHelperInstance *apHelper,
	vertex int,
	parent int,
	time *int,
	graph Graph
) {
	apHelperInstance.visited[vertex] = true
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+
+	apHelperInstance.discovery_time[vertex] = *time                                       // Mark the time of discovery of a vertex
+	apHelperInstance.earliest_discovery[vertex] = apHelperInstance.discovery_time[vertex] // set the earliest discovery time to the discovered time
+	*time++                                                                               // increment the time
```
tjgurwara99, line 46: Comment above not beside the expressions - again makes it easier to read.

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+
+	for next_vertex := range graph.edges[vertex] {
+		if next_vertex == parent {
+			continue
+		}
+
+		if apHelperInstance.visited[next_vertex] { // if the vertex is already visited, then it is a back edge and manipulate earliest discovery time accordingly
```
tjgurwara99, line 53: Not a useful comment - anyone who reads this code would be able to infer what is happening.

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+			apHelperInstance.earliest_discovery[vertex] = min.Int(apHelperInstance.earliest_discovery[vertex], apHelperInstance.discovery_time[next_vertex])
```
tjgurwara99, line 54:
```suggestion
			apHelperInstance.earliest_discovery[vertex] = min.Int(
				apHelperInstance.earliest_discovery[vertex],
				apHelperInstance.discovery_time[next_vertex]
			)
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+		} else {
+			apHelperInstance.child_cnt[vertex]++
+			articulationPointHelper(apHelperInstance, next_vertex, vertex, time, graph)
```
tjgurwara99, line 57:
```suggestion
			articulationPointHelper(
				apHelperInstance,
				next_vertex,
				vertex,
				time, graph
			)
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+			apHelperInstance.earliest_discovery[vertex] = min.Int(apHelperInstance.earliest_discovery[vertex], apHelperInstance.earliest_discovery[next_vertex])
```
tjgurwara99, line 58:
```suggestion
			apHelperInstance.earliest_discovery[vertex] = min.Int(
				apHelperInstance.earliest_discovery[vertex],
				apHelperInstance.earliest_discovery[next_vertex]
			)
```

graph/articulationpoints.go
```diff
@@ -0,0 +1,64 @@
...
+	*time++                                                                               // increment the time
+
+	for next_vertex := range graph.edges[vertex] {
+		if next_vertex == parent {
+			continue
+		}
+
+		if apHelperInstance.visited[next_vertex] { // if the vertex is already visited, then it is a back edge and manipulate earliest discovery time accordingly
+			apHelperInstance.earliest_discovery[vertex] = min.Int(apHelperInstance.earliest_discovery[vertex], apHelperInstance.discovery_time[next_vertex])
+		} else {
```
tjgurwara99, line 55: Do you really need this else? I remember I posted a better way to do this block previously.

tjgurwara99 (changes requested):

graph/articulationpoints.go
tjgurwara99: Here is the [link](https://github.com/TheAlgorithms/Go/pull/452#discussion_r773512443) to that recommendation.

graph/articulationpoints.go
anirudnits: I would disagree there. The back edge is a special case here and needs to be dealt separately, the earliest discovery time for a vertex is changed differently provided whether the edge is a back edge or not and that could cause some confusion on the part of the reader. That's why I explicitly mentioned this in the comment.

graph/articulationpoints.go
anirudnits: You referenced to use `continue` after the `if` block so that the "now" `else` block is not executed. However, I believe the `if-else` block makes more sense here, `if visited do something; else do someotherthing` and therefore kept it there.

graph/articulationpoints.go
anirudnits: Not sure, when I wrote the comments above they looked odd without the code. I'll do it however I'm skeptical whether it would increase readability.

graph/articulationpoints.go
tjgurwara99: You may think it makes sense and it is usually the case in other languages but in Go we usually return early. This is a style of writing that is seen everywhere in standard library and is preferred by all experienced devs (IMO). This repo is for Go learners and it should stick with Go's style of writing code and not just copy and/or transliterate other language code.

graph/articulationpoints.go
tjgurwara99: It does when people's screens are small (mine is for example)

graph/articulationpoints.go
tjgurwara99: The if condition is telling everyone that this is a special case and needs to be dealt with separately. Comment doesn't add anything new other than technical jargon (IMO)

graph/articulationpoints.go
tjgurwara99: [here](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) is a link to explain what I'm talking about above. You may disagree with this and that's your right but I would still argue that experienced go devs would frown at `else` statements (I have seen that happen a lot in interviews) and since this is the kind of repo that junior devs would be looking at for thier interview prep, I think we're should stick with idiomatic Go code in this repo.

Read the link for reference. Almost all go devs are aware of this "line of sight" and recommend it to everyone.

This is not a hard and fast take btw, in fact sometimes you can't get rid of the else statements but since in this case you can then I think you should.

graph/articulationpoints.go
anirudnits: Done

graph/articulationpoints.go
anirudnits: Done

graph/articulationpoints.go
anirudnits: Done.

graph/articulationpoints.go
anirudnits: Done.

graph/articulationpoints.go
anirudnits: Understood, thanks for explaining the rationale behind this, I was unaware. Implemented this in the latest commit.

graph/articulationpoints.go
anirudnits: Done.

graph/articulationpoints.go
anirudnits: Done.

graph/articulationpoints.go
anirudnits: Done.

graph/articulationpoints.go
anirudnits: Done.

tjgurwara99 (approved):
Nice! Now it looks very clean 😄 

Thank you for contributing!

graph/articulationpoints.go
```diff
@@ -0,0 +1,100 @@
...
+// articulationPointHelper is a recursive function to traverse the graph
+// and mark articulation points. Based on the depth first search transversal
+// of the graph, however modified to keep track and update the
+// `child_cnt`, `discovery_time`` and `earliest_discovery` slices defined above
+func articulationPointHelper(
+	apHelperInstance *apHelper,
+	vertex int,
+	parent int,
+	time *int,
+	graph Graph,
```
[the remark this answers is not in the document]
anirudnits: Yeah I did consider that. However, I decided against it simply because I thought this way its more clear about the usage of the function parameters. The "pure" parameters being passed as values while the "impure" parameters being passed as reference to be changed and tinkered inside the functions forming a "clearer" distinction within the parameters. I do understand that passing the graph variable by reference help squeeze out the last bit of execution optimization but as a normal paradigm I tried to keep the function as "pure" as possible. I don't know if has to do with standard golang conventions, in which case I can modify this to be passed as reference in my implementation.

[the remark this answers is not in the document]
tjgurwara99: No nothing to do with standard Golang convention, rather its the idea related to pointer operations. If you're not modifying the value of the `object` then pass the pointer instead of a copy. However, if you are modifying the value, you have to be very careful about the operations on the value.  In fact, if you're not changing the value then the function by definition is "pure" since there is no change in state even if you pass the value by reference. Purity of the function doesn't talk about pointers or copy but rather the notion of state. I hope this is clear.

Anyways, I don't mind you keeping it as is as well if the other maintainers agree @raklaptudirm @siriak ??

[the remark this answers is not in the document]
raklaptudirm: The go rule is "when in doubt, use a pointer". If the function is not making any changes to the input graph, a pointer is the better choice. But if it is, and we don't want it reflected in the input, we will be better off with a value.

Since this function does not modify the input graph, using a pointer is memory efficient, free from any side effects, and a good coding practice. Following the convention, I think it being a pointer is better.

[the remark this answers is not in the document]
anirudnits: @raklaptudirm Thanks for the explanation. I've changed the function so the graph is passed as reference.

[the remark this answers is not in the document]
tjgurwara99: That rule is for receivers only though, that is not generally true if we're talking about function arguments.

tjgurwara99 (approved):
All good from me 😄

raklaptudirm (approved):

graph/articulationpoints.go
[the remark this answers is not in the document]
raklaptudirm: @tjgurwara99  Yes, but I generally find that the rule can be used in function arguments too with good effect.

PR #453, opened 2021-12-20
Abs algo

i-redbyte:


math/binary/abs.go
```diff
@@ -0,0 +1,7 @@
+package binary
+
+// ABS - this function return absolute value use binary operation
+func ABS(base, n int) int {
+	m := n >> (base - 1)
```
raklaptudirm, line 5: Explain these bit-wise operations in comments.

raklaptudirm (changes requested):

math/binary/abs.go
i-redbyte: Done

math/binary/abs.go
```diff
@@ -0,0 +1,12 @@
+package binary
+
+// ABS - this function return absolute value use binary operation
+// Principle of operation:
+// 1) Get the mask by right shift by the bit width of the variable (int64, int32, int16, int8)
```
raklaptudirm, line 5: Since the method accepts an `int`, supporting other lengths is not required in my opinion. It will always be an `int`.

raklaptudirm (changes requested):

math/binary/abs.go
i-redbyte: > Since the method accepts an `int`, supporting other lengths is not required in my opinion. It will always be an `int`.

this is added only to describe the algorithm, to understand that the shift must be a multiple of the digit of your integer

raklaptudirm: In my opinion, it can be adequately described in the comments, and a separate variable is redundant.

i-redbyte: done

math/abs/abs.go
```diff
@@ -0,0 +1,16 @@
...
+// description: Absolute value
+// details:
+// In mathematics, the absolute value or modulus of a real number x, denoted |x|, is the non-negative value of x without regard to its sign. [Absolute value](https://en.wikipedia.org/wiki/Average#Arithmetic_mean)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see abs_test.go
+
+package abs
+
+// ABS - this function return absolute value
+func ABS(n int) int {
```
tjgurwara99, line 11: Why use `ABS` instead of `Abs`??

```suggestion
func Abs(n int) int {
```

tjgurwara99: Also, this should just exist in the math package not in a separate package.

math/binary/abs.go
```diff
@@ -0,0 +1,12 @@
...
+// 2) For negative numbers, above step sets mask as 1 1 1 1 1 1 1 1 and 0 0 0 0 0 0 0 0 for positive numbers.
+// 3) Add the mask to the given number.
+// 4) XOR of mask + n and mask gives the absolute value.
+func ABS(base, n int) int {
```
tjgurwara99, line 10:
```suggestion
func Abs(base, n int) int {
```

tjgurwara99 (changes requested):

math/abs/abs.go
i-redbyte: Done

math/binary/abs.go
i-redbyte: Done

math/binary/abs.go
```diff
@@ -0,0 +1,13 @@
+package binary
+
+// Abs - this function return absolute value use binary operation
```
tjgurwara99, line 3:
```suggestion
// Abs returns absolute value use binary operation
```

math/abs.go
```diff
@@ -0,0 +1,16 @@
+// abs.go
+// description: Absolute value
+// details:
+// In mathematics, the absolute value or modulus of a real number x, denoted |x|, is the non-negative value of x without regard to its sign. [Absolute value](https://en.wikipedia.org/wiki/Average#Arithmetic_mean)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see abs_test.go
+
+package math
+
+// Abs - this function return absolute value
```
tjgurwara99, line 10:
```suggestion
// Abs returns absolute value
```

tjgurwara99 (changes requested):
Small changes.

math/abs_test.go
```diff
@@ -0,0 +1,72 @@
+package math
+
+import (
+	"github.com/TheAlgorithms/Go/math/binary"
+	"math"
+	"testing"
+)
+
+func TestABS(t *testing.T) {
```
tjgurwara99, line 9:
```suggestion
func TestAbs(t *testing.T) {
```

math/abs_test.go
```diff
@@ -0,0 +1,72 @@
...
+func TestBinaryABS(t *testing.T) {
+	tests := getTests()
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if got := binary.Abs(64, test.n); got != test.want {
+				t.Errorf("Abs() = %v, want %v", got, test.want)
+			}
+		})
+	}
+}
```
tjgurwara99, line 29: Why did you move this here when `abs.Binary` is not in the package? This does not belong in this package. Move it inside the binary package.

tjgurwara99 (changes requested):

math/abs_test.go
i-redbyte: Done. is everything ok now?

i-redbyte: Done

math/abs_test.go
i-redbyte: Done. is everything ok now?

tjgurwara99 (approved):

raklaptudirm (approved):

PR #454, opened 2021-12-22
DO NOT MERGE Example of Generics in Sorting Function

ChiefMateStarbuck:
This PR serves as an example of how generics could look in in the sorting package. This code compiles with `golang1.18Beta`. Note that the `Numbers` constraint is not currently in the standard package but is being considered.

Link to full file: https://github.com/ChiefMateStarbuck/Go-1/blob/master/sort/bubblesort.go

### DIFF
--- a/sort/bubblesort.go
+++ b/sort/bubblesort.go
@@ -5,11 +5,7 @@ package sort
 
 import "constraints"
 
-type Numbers interface {
-	~contstraints.Integer | ~constraints.Float
-}
-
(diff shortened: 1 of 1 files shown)


tjgurwara99:
Make this a draft PR.

Also, the Numbers constraint is unnecessary for the sort – what you need is the `constraints.Ordered` for and that should be enough. 

[Here](https://gotipplay.golang.org/p/FUAZvFwNVb7) is an example of what I mean.

Remember `<` operator is defined for strings as well.

PR #456, opened 2021-12-31
add merge sort

1327253585:
Add non recursive merge sorting algorithm


sort/mergesort2.go
```diff
@@ -0,0 +1,37 @@
+package sort
+
+import "github.com/TheAlgorithms/Go/math/min"
+
+func Mergesort2(items []int) []int {
```
tjgurwara99, line 5:
```suggestion
func Merge2(items []int) []int {
```

sort/mergesort2.go
```diff
@@ -0,0 +1,37 @@
...
+	l1, r1, l2, r2 := l, mid, mid, r
+	var tmp []int
+	for l1 < r1 && l2 < r2 {
+		if items[l1] < items[l2] {
+			tmp = append(tmp, items[l1])
+			l1++
+		} else {
+			tmp = append(tmp, items[l2])
+			l2++
+		}
```
tjgurwara99, line 24:
```suggestion
		if items[l1] < items[l2] {
			tmp = append(tmp, items[l1])
			l1++
			continue
		}
		tmp = append(tmp, items[l2])
		l2++
```

tjgurwara99 (changes requested):

tjgurwara99 (approved):

sort/mergesort2.go
```diff
@@ -0,0 +1,37 @@
+package sort
+
+import "github.com/TheAlgorithms/Go/math/min"
+
+func Merge2(items []int) []int {
```
raklaptudirm, line 5: I believe `MergeIter` would be a better name for this function. `Merge2` does not convey that much information.

raklaptudirm (changes requested):

sort/mergesort2.go
tjgurwara99: In that case, same applies to `Merge` as well. Maybe a documentation comment here would be better. Not sure if you agree.

I think having a consistent way to do it would be better and we should come up with a naming convention or something for this repo.

1327253585: According to [Contribution Guidelines](https://github.com/TheAlgorithms/Go/blob/master/CONTRIBUTING.md) :
- If an implementation of the algorithm already exists and your version is different from that implemented, please use incremental numeric digit as a suffix. For example: if `binarysearch.go` already exists in the search folder, and you are contributing a new implementation, the filename should be `binarysearch2.go` and for a third implementation, `binarysearch3.go`.

Therefore,  if the exist recursive implementation of  merge sort  called `MegeRecursion` ,  `MergeIter` would be a better name for this function.

raklaptudirm: @tjgurwara99 we could rename the existing `sorts.Merge` to `sorts.MergeRecursive`

siriak (approved):

sort/mergesort2.go
tjgurwara99: I agree with this. We should update the contribution guidelines accordingly as well.

sort/mergesort.go
```diff
@@ -45,3 +47,13 @@ func Mergesort(items []int) []int {
 	return merge(a, b)
 
 }
+
+func MergesortIter(items []int) []int {
```
raklaptudirm, line 51: `MergeIter` not `MergesortIter`. Sort is in the package name.
https://go.dev/doc/effective_go#package-names

raklaptudirm (changes requested):

raklaptudirm (approved):

tjgurwara99 (approved):

PR #459, opened 2022-01-06
feat: add union find structure

1327253585:
feat: add union find structure


tjgurwara99:
This algorithm doesn't belong in a separate package. There is already a set package inside the structure package. You can see that your implementation would fit there.

structure/set/unionfind.go
```diff
@@ -0,0 +1,63 @@
...
+
+type UFS struct {
+	nodes map[Value]*node
+}
+
+func NewUFS() *UFS {
+	return &UFS{
+		nodes: make(map[Value]*node),
+	}
+}
```
tjgurwara99, line 20: Make `UFS` unexported since you're setting up the nodes through the `NewUFS` constructor and returning a pointer.

structure/set/unionfind.go
```diff
@@ -0,0 +1,63 @@
...
+	if p.value != value {
+		u.nodes[value] = u.find(p.value, init)
+	}
+	return p
+}
+
+func (u *UFS) Connected(x, y Value) bool {
+	return u.Find(x) == u.Find(y)
+}
+
```
tjgurwara99, line 43: I'm curious as to what you're doing here? The notion of sets have nothing to do with connectedness as far as my mathematical knowledge goes. Therefore this doesn't belong with this structure.

structure/set/unionfind.go
```diff
@@ -0,0 +1,63 @@
+package set
+
+// https://en.wikipedia.org/wiki/Disjoint-set_data_structure
+
+type Value interface{}
+
+type node struct {
+	rank  int
+	value Value
```
tjgurwara99, line 9: The usual way is to refer to this as `parent` rather than `value`. Keep it consistent with all the documentation that is available online for people to refer to. Even the Wikipedia article you refer to calls this parent. And instead of making this `Value` interface you should consider it to be a node. All your implementations would be simplified with that.

tjgurwara99 (changes requested):
Your implementation is fine, but there is already an implementation of this algorithm with this repository. My apologies that I asked for further change when there is already an implementation of this in this repo. [Here](https://github.com/TheAlgorithms/Go/blob/6134d6e383f0e2f42d9123c50d22d6e7333c1a3f/graph/kruskal.go#L21) is the link to the place where this structure sits.

Thanks for contributing but I don't think its wise to add duplicate code in this repo. @siriak @raklaptudirm what do you guys think? (The implementation is conceptually the same except for minute differences in approach)

raklaptudirm:
Duplication of algorithms should be avoided if the implementations are the same in essence imo.

1327253585:
sorry， i didn't see the implementations. I will close this pull request.

PR #461, opened 2022-01-15
Code to check whether two strings are Anagrams

Versatile-Venkat:
-  Added code to check wheather two given strings are anagrams
-  Added file names isanagram.go and isanagram_test.go and they  match File name guidelines
-  Added tests and examples.
-  Added documentation so that the program is self-explanatory and educational 
-  Relevant documentation/comments is changed or added
-  I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/strings/anagram/isanagram.go
+++ b/strings/anagram/isanagram.go
@@ -1,0 +1,41 @@
+// isanagram.go
+// description: Checks if two given strings are anagram or not
+// details: An anagram is a word or phrase that's formed by rearranging the letters of another word or phrase
+// Reference: https://www.geeksforgeeks.org/pangram-checking/
+// Author : Venkatasubramanian
+
+package anagram
+
+import (
+	"regexp"
+	"strings"
+)
+
+func cleanString(text string) string {
+	cleanText := strings.ToLower(text)                      // Convert to lowercase
+	cleanText = strings.Join(strings.Fields(cleanText), "") // Remove spaces
+	regex, _ := regexp.Compile(`[^\p{L}\p{N} ]+`)           // Regular expression for alphanumeric only characters
+	return regex.ReplaceAllString(cleanText, "")
+}
+
+func isanagram(text1 string, text2 string) bool {
+	cleanText1 := cleanString(text1)
+	cleanText2 := cleanString(text2)
+	if len(cleanText1) != len(cleanText2) {
+		return false
+	}
+	arr := make([]int, 256)
+	for _, char := range cleanText1 {
+		arr[char]++
+	}
+	for _, char := range cleanText2 {
+		arr[char]--
+	}
+	for i := 0; i < 256; i++ {
+		if arr[i] != 0 {
+			return false
+		}
+	}
+	return true
+
+}
--- a/strings/anagram/isanagram.go
+++ b/strings/anagram/isanagram.go
@@ -24,15 +24,15 @@ func isanagram(text1 string, text2 string) bool {
 	if len(cleanText1) != len(cleanText2) {
 		return false
 	}
-	arr := make([]int, 256)
+	charArray := make([]int, 256) //This array is used to represent the ASCII values.
 	for _, char := range cleanText1 {
-		arr[char]++
+		charArray[char]++ //Increments the index for the characters in cleanText1 Example: charArray['a']++ => charArray[97]++
 	}
 	for _, char := range cleanText2 {
-		arr[char]--
+		charArray[char]-- //Decrements the index for the characters in cleanText2  Example: charArray['a']-- => charArray[97]--
 	}
 	for i := 0; i < 256; i++ {
-		if arr[i] != 0 {
+		if charArray[i] != 0 { //If the characters in both texts are balanced, it is an anagram otherwise No
 			return false
 		}
 	}
--- a/strings/anagram/isanagram_test.go
+++ b/strings/anagram/isanagram_test.go
@@ -1,0 +1,60 @@
+package anagram
+
+import (
+	"testing"
+)
+
+var testCases = []struct {
+	name     string // test description
+	input1   string // user input1
+	input2   string //user input 2
+	expected bool   // expected return
+}{
+	{
+		"Anagram String 1",
+		"angel",
+		"glean",
+		true,
+	},
+	{
+		"Non-Anagram String 1",
+		"git",
+		"hub",
+		false,
+	},
+	{
+		"Anagram String 2",
+		"mango",
+		"goman",
+		true,
+	},
+	{
+		"Non-Anagram String 2",
+		"length",
+		"different",
+		false,
+	},
+	{
+		"Anagram String 3",
+		"distill231",
+		"23sidtill1",
+		true,
+	},
+	{
+		"Anagram String 4",
+		"I am happy",
+		"Happy am I",
+		true,
+	},
+}
+
+func TestIsAnagram(t *testing.T) {
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			func_result := isanagram(test.input1, test.input2)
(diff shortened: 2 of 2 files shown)


strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
+// isanagram.go
```
raklaptudirm, line 1: Use a go package comment instead of this comment style.

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+	return regex.ReplaceAllString(cleanText, "")
+}
+
+func isanagram(text1 string, text2 string) bool {
+	cleanText1 := cleanString(text1)
+	cleanText2 := cleanString(text2)
+	if len(cleanText1) != len(cleanText2) {
+		return false
+	}
+	charArray := make([]int, 256) //This array is used to represent the ASCII values.
```
raklaptudirm, line 27: Why a hard-coded upper bound? Slices can be reallocated with [`append()`](https://golangbyexample.com/append-function-in-golang/).

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, char := range cleanText1 {
+		charArray[char]++ //Increments the index for the characters in cleanText1 Example: charArray['a']++ => charArray[97]++
+	}
+	for _, char := range cleanText2 {
+		charArray[char]-- //Decrements the index for the characters in cleanText2  Example: charArray['a']-- => charArray[97]--
+	}
+	for i := 0; i < 256; i++ {
```
raklaptudirm, line 34: Same as above.

raklaptudirm (changes requested):

strings/anagram/isanagram.go
Versatile-Venkat: Upper bound is hard coded because all the characters that exists has ASCII value between 0 and 255

raklaptudirm:
Is there any reason for keeping it in the ascii range?

Versatile-Venkat:
> Is there any reason for keeping it in the ascii range?

As we are dealing with strings, and strings obviously contains characters that are within this range. This is the reason for having a constant space

raklaptudirm:
go strings are `rune` arrays, and use the unicode encoding, which supports a much wider range than ascii.

strings/anagram/isanagram.go
tjgurwara99: Agreed

tjgurwara99 (changes requested):
Your algorithm is only catered to ASCII characters. Go is a modern language and supports utf-8 so I think you should modify your algorithm to also work on utf-8 characters since it takes in strings and would fail on utf-8 characters - meaning this is not a production quality algorithm and should be reconsidered.

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"regexp"
+	"strings"
+)
+
+func cleanString(text string) string {
+	cleanText := strings.ToLower(text)                      // Convert to lowercase
+	cleanText = strings.Join(strings.Fields(cleanText), "") // Remove spaces
+	regex, _ := regexp.Compile(`[^\p{L}\p{N} ]+`)           // Regular expression for alphanumeric only characters
```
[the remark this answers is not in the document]
raklaptudirm: The current implementation without the `cleanString()` should work, since the `range` keyword automatically breaks the string into valid utf-8 runes. Removal of `cleanString()` and changing the `[]int` to a `map[rune]int` should work. Also, anagrams exclude whitespace, so you can either still include a different `cleanString()` implementation, or run a check in the range loop with `unicode.IsSpace()`.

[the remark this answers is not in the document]
Versatile-Venkat: Sure will resolve it within EOD

strings/anagram/isanagram.go
raklaptudirm: Using a multi-line comment is not what I meant. See how to write go documentation comments: https://go.dev/doc/effective_go#commentary

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+	return regex.ReplaceAllString(cleanText, "")
+}
+
+func isanagram(text1 string, text2 string) bool {
```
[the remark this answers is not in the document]
tjgurwara99: This has still not been resolved

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
+/*isanagram.go
+description: Checks if two given strings are anagram or not
+details: An anagram is a word or phrase that's formed by rearranging the letters of another word or phrase
+Reference: https://www.geeksforgeeks.org/pangram-checking/
+Author : Venkatasubramanian */
+
+package anagram
```
tjgurwara99, line 7:
```suggestion
// isanagram.go
// description: Checks if two given strings are anagram or not
// details: An anagram is a word or phrase that's formed by rearranging the letters of another word or phrase
// Reference: https://www.geeksforgeeks.org/pangram-checking/
// Author : Venkatasubramanian

// Package anagram does something <--- put your explanation here right on top of the package declaration
package anagram
```

tjgurwara99: The multi line comment was not what @raklaptudirm was referring to in the previous review.

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+		return false
+	}
+	charMap := make(map[rune]int)
+
+	for i := 0; i < len(cleanText1); i++ {
+		if charMap[cleanText1[i]] != 0 {
+			charMap[cleanText1[i]]++
+		} else {
+			charMap[cleanText1[i]] = 1
+		}
```
tjgurwara99, line 29:
```suggestion
		if charMap[cleanText1[i]] != 0 {
			charMap[cleanText1[i]]++
			continue
		}
		charMap[cleanText1[i]] = 1
```

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	for i := 0; i < len(cleanText2); i++ {
+		if charMap[cleanText2[i]] > 0 {
+			charMap[cleanText2[i]]--
+		} else {
+			return false
+		}
```
tjgurwara99, line 36:
```suggestion
		if charMap[cleanText2[i] <= 0 {
			return false
		}
		charMap[cleanText2[i]]--
```

strings/anagram/isanagram.go
tjgurwara99: Also, you referenced `pangram-checking` from geeksforgeeks but that is a completely different algorithm. Please check your references properly before committing them.

strings/anagram/isanagram.go
```diff
@@ -0,0 +1,41 @@
...
+
+func isanagram(text1 string, text2 string) bool {
+	cleanText1 := cleanString(text1)
+	cleanText2 := cleanString(text2)
+	if len(cleanText1) != len(cleanText2) {
+		return false
+	}
+	charMap := make(map[rune]int)
+
+	for i := 0; i < len(cleanText1); i++ {
```
tjgurwara99, line 24: You could've just as easily used the `range` operator to look over the string runes. That would have prevented you from keeping track of `i`  over and over.

IMO, its much better to use range here as it makes it easier to understand what you're trying to achieve.

tjgurwara99 (changes requested):
Just reading a different languages code and transliterating to a different language is not a good strategy for learning the said language. It just seems that you took the geeksforgeeks hash map implementation and rewrote it in Go without much thought (its not plagiarism as you reference it, albeit to a different article - possible human error).

PR #465, opened 2022-02-02
Simplify BreadthFirstSearch.go

Primital:
Remove redundant if clause which condition is always met at line of execution

The condition of the for loop and the if clause are identical, so the condition of the if clause will always be met

I would suggest refactoring line 15,16 to
`v, queue := queue[0], queue[1:]` 
as well, but perhaps it is better to keep it simpler.


siriak (approved):
LGTM, thanks!

tjgurwara99 (approved):

Issue #460, opened 2022-01-11
Optimized nlogn approach for longest increasing sub-sequence


anirudnits:
**Description**
The dynamic programming approach for finding the longest increasing sub-sequence in `O(n^2)` within an integer array is defined here: https://github.com/TheAlgorithms/Go/blob/master/dynamic/longestincreasingsubsequence.go. However, there's a greedy approach which solves the problem in `O(nlogn)`. 

For enhancement:
   - Greedy logic for finding the longest increasing sub-sequence in lesser asymptotic time.

tjgurwara99:
I don't think you need to open issues when you're thinking of adding an algorithm to the repository. Just open a PR and the maintainers will take a look.

siriak:
On the other hand, it gives a leg up to other folks who might be interested in contributing, doesn't it? There's nothing wrong with having a backlog of up-for-grabs issues IMO.

PR #466, opened 2022-02-03
Longest increasing subsequence greedy

anirudnits:
PR in reference to issue #460.


raklaptudirm (changes requested):
We already have a non-greedy implementation [here](https://github.com/TheAlgorithms/Go/blob/master/dynamic/longestincreasingsubsequence.go). Since the algorithm is similar, move it to the `github.com/TheAlgorithms/Go/dynamic/` package

anirudnits:
Hello @raklaptudirm, I didn't want this greedy implementation under the dynamic-programming directory as it causes confusion and also is not in-line with how the repository is organized. Should I create a new directory for greedy approaches and put this in there?

raklaptudirm:
There are no sub-packages under `dynamic`, so I don't think we should introduce one. @tjgurwara99 @siriak What are your opinions?

other/longestincreasingsubsequencegreedy/longestincreasingsubsequencegreedy.go
```diff
@@ -0,0 +1,43 @@
+package longestincreasingsubsequencegreedy
```
tjgurwara99, line 1: This doesn't need a separate package. There is a `dynamic` package in the repository which already has a longest increasing subsequence algorithm. Just move this algorithm there along with the old implementation.

tjgurwara99 (changes requested):
Once moved, you can also use the tests there.

other/longestincreasingsubsequencegreedy/longestincreasingsubsequencegreedy.go
anirudnits: Done.

tjgurwara99 (changes requested):

dynamic/longestincreasingsubsequence_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/TheAlgorithms/Go/dynamic"
 )
 
-func TestLongestIncreasingSubsequence(t *testing.T) {
+func LongestIncreasingSubsequenceTest(t *testing.T, algorithm func(nums []int) int) {
```
tjgurwara99, line 10: This shouldn't be exported

tjgurwara99 (changes requested):

tjgurwara99 (approved):

raklaptudirm (approved):

PR #470, opened 2022-02-13
Added backtracking algorithm

sixwaaaay:
Signed-off-by: sixway <lw.2675@qq.com>

### DIFF
--- a/backtrack/permutation.go
+++ b/backtrack/permutation.go
@@ -20,22 +20,22 @@ func Permutation(nums []int) [][]int {
 	perm := make([]int, n)
 	vis := make([]bool, n)
 	var permutations [][]int
-	var backtrackPerm func(int)
-	backtrackPerm = func(depth int) {
-		if depth == n {
-			permutations = append(permutations, append([]int{}, perm...))
-			return
-		}
-		for i := range nums {
-			if vis[i] || (i > 0 && !vis[i-1] && nums[i] == nums[i-1]) {
-				continue
-			} // get first element that never used/visited
-			perm[depth] = nums[i]
-			vis[i] = true
-			backtrackPerm(depth + 1)
-			vis[i] = false
-		}
-	}
-	backtrackPerm(0)
+	backtrackPerm(&permutations, nums, 0, n, perm, vis)
 	return permutations
+}
+
+func backtrackPerm(permutations *[][]int, nums []int, depth int, n int, perm []int, vis []bool) {
+	if depth == n {
+		*permutations = append(*permutations, append([]int{}, perm...))
+		return
+	}
+	for i := range nums {
+		if vis[i] || (i > 0 && !vis[i-1] && nums[i] == nums[i-1]) {
+			continue
+		} // get first element that never used/visited
+		perm[depth] = nums[i]
+		vis[i] = true
+		backtrackPerm(permutations, nums, depth+1, n, perm, vis)
+		vis[i] = false
+	}
 }
--- a/backtrack/permutation.go
+++ b/backtrack/permutation.go
@@ -1,0 +1,26 @@
+// permutation.go
+// description: generate permutations of a given set
+// details:
+// In mathematics, a permutation of a set is, loosely speaking, an arrangement of its members into a sequence or linear order, or if the set is already ordered, a rearrangement of its elements. The word "permutation" also refers to the act or process of changing the linear order of an ordered set.
+// author: [sixwaaaay](https://github.com/sixwaaaay)
+// see permutation_test.go
+
+package backtrack
+
+func Permutation(nums []int) (set [][]int) {
+	n := len(nums)
+	var backtrack func(int)
+	backtrack = func(depth int) {
+		if depth == n {
+			set = append(set, append([]int{}, nums...))
+			return
+		}
+		for i := depth; i < len(nums); i++ {
+			nums[depth], nums[i] = nums[i], nums[depth]
+			backtrack(depth + 1)
+			nums[depth], nums[i] = nums[i], nums[depth]
+		}
+	}
+	backtrack(0)
+	return
+}
--- a/backtrack/permutation.go
+++ b/backtrack/permutation.go
@@ -1,26 +1,41 @@
 // permutation.go
-// description: generate permutations of a given set
+// description: generate subsets use backtrack algorithm
 // details:
-// In mathematics, a permutation of a set is, loosely speaking, an arrangement of its members into a sequence or linear order, or if the set is already ordered, a rearrangement of its elements. The word "permutation" also refers to the act or process of changing the linear order of an ordered set.
+// In mathematics, a permutation of a set is, loosely speaking,
+// an arrangement of its members into a sequence or linear order,
+// or if the set is already ordered, a rearrangement of its elements.
+// The word "permutation" also refers to the act or process of changing
+// the linear order of an ordered set.
 // author: [sixwaaaay](https://github.com/sixwaaaay)
 // see permutation_test.go
 
 package backtrack
 
-func Permutation(nums []int) (set [][]int) {
+import "sort"
+
+// Permutation is a function to generate all permutations of a given array
+func Permutation(nums []int) [][]int {
+	sort.Ints(nums)
 	n := len(nums)
-	var backtrack func(int)
-	backtrack = func(depth int) {
+	perm := make([]int, n)
+	vis := make([]bool, n)
+	var permutations [][]int
+	var backtrackPerm func(int)
+	backtrackPerm = func(depth int) {
 		if depth == n {
-			set = append(set, append([]int{}, nums...))
+			permutations = append(permutations, append([]int{}, perm...))
 			return
 		}
-		for i := depth; i < len(nums); i++ {
-			nums[depth], nums[i] = nums[i], nums[depth]
-			backtrack(depth + 1)
-			nums[depth], nums[i] = nums[i], nums[depth]
+		for i := range nums {
+			if vis[i] || (i > 0 && !vis[i-1] && nums[i] == nums[i-1]) {
+				continue
+			} // get first element that never used/visited
+			perm[depth] = nums[i]
+			vis[i] = true
+			backtrackPerm(depth + 1)
+			vis[i] = false
 		}
 	}
-	backtrack(0)
-	return
+	backtrackPerm(0)
+	return permutations
 }
--- a/backtrack/permutation_test.go
+++ b/backtrack/permutation_test.go
@@ -10,13 +10,73 @@ import (
 	"testing"
 )
 
+var testCasesForPermutation = []struct {
+	name         string  // test case name
+	nums         []int   // nums array
+	permutations [][]int // expected subsets
+}{
+	{
+		name: "array without duplicate elements",
+		nums: []int{
+			2, 3, 4,
+		},
+		permutations: [][]int{
+			{2, 3, 4},
+			{2, 4, 3},
+			{3, 2, 4},
+			{3, 4, 2},
+			{4, 2, 3},
+			{4, 3, 2},
+		},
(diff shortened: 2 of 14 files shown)


backtrack/permutation.go
```diff
@@ -0,0 +1,26 @@
+// permutation.go
+// description: generate permutations of a given set
+// details:
+// In mathematics, a permutation of a set is, loosely speaking, an arrangement of its members into a sequence or linear order, or if the set is already ordered, a rearrangement of its elements. The word "permutation" also refers to the act or process of changing the linear order of an ordered set.
+// author: [sixwaaaay](https://github.com/sixwaaaay)
+// see permutation_test.go
+
+package backtrack
+
+func Permutation(nums []int) (set [][]int) {
```
tjgurwara99, line 10: The naming doesn't make sense - there is no indication that this is a permutation of a set rather than an elementary permutation. You've written very good code, the only thing I feel could be improved is the naming you have chosen.

backtrack/subset.go
```diff
@@ -0,0 +1,38 @@
...
+
+func Subset(nums []int) (set [][]int) {
+	n := len(nums)
+	// init index array for record the index of each element selected
+	index := make([]bool, n)
+	for i := 0; i < n; i++ {
+		index[i] = false
+	}
+	var backtrack func(int)
+	backtrack = func(depth int) {
```
tjgurwara99, line 19: Can you not call this something else? There could be infinitely many things that could be called backtracking - make it more specific to the problem at hand.

backtrack/subset.go
```diff
@@ -0,0 +1,38 @@
+// subset.go
+// description: subset of a given set
+// details:
+//In mathematics, set A is a subset of a set B if all elements of A are also elements of B; B is then a superset of A. It is possible for A and B to be equal; if they are unequal, then A is a proper subset of B. The relationship of one set being a subset of another is called inclusion (or sometimes containment). A is a subset of B may also be expressed as B includes (or contains) A or A is included (or contained) in B.
```
tjgurwara99, line 4: Split this into multiple lines.

backtrack/permutation_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+// test permutation function
+func TestPermutation(t *testing.T) {
+	t.Run("generate permutations for given array", func(t *testing.T) {
+		excepted := [][]int{{1, 2, 3}, {1, 3, 2}, {2, 1, 3}, {2, 3, 1}, {3, 2, 1}, {3, 1, 2}}
```
tjgurwara99, line 16: Add more tests here too 🙂

backtrack/permutation.go
```diff
@@ -0,0 +1,26 @@
...
+			return
+		}
+		for i := depth; i < len(nums); i++ {
+			nums[depth], nums[i] = nums[i], nums[depth]
+			backtrack(depth + 1)
+			nums[depth], nums[i] = nums[i], nums[depth]
+		}
+	}
+	backtrack(0)
+	return
```
tjgurwara99, line 25: no need for this return

tjgurwara99 (changes requested):
Good implementation overall! Just few minor changes that I think are worth considering. For example, the naming of your functions or using the set implementation already in the repository etc. I've commented on the parts that I think you should modify or think about more.

backtrack/subset.go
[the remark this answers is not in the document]
sixwaaaay: the backtrack function is recursive, so when I use closure, it must be this form of definition

backtrack/permutation.go
sixwaaaay: when remove this return: .\permutation.go:26:1: missing return at end of function. so this return can't be removed.

backtrack/subset.go
```diff
@@ -0,0 +1,38 @@
...
+//  - https://en.wikipedia.org/wiki/Subset
+// author: [sixwaaaay](https://github.com/sixwaaaay)
+// see subset_test.go
+
+package backtrack
+
+func Subset(nums []int) (set [][]int) {
```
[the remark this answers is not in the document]
sixwaaaay: Do I directly change this to a good one with complex branching and bounds, or is it better to implement a simple (n int) function first, or to limit the currently written function to an array nums of distinct integers
For example, permutation(n int), I can assume that this time the array is only distinct integers

backtrack/permutation.go
tjgurwara99: Oh, I didn't see the return `set` in the function signature. I think you should remove the `set` from the function return because it is very easy to miss that.

Having an explicit is not going to hurt the implementation and explicit is better than implicit.

backtrack/subset.go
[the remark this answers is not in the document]
tjgurwara99: Just directly change this to the end goal - no need to go through the intermediate steps since conceptually they will be the same and it would require us maintainers to review more implementations of the same algorithm.

backtrack/subset.go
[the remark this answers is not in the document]
tjgurwara99: Hmm, then I don't think you need closure in this case - most of the things that you are using through closure are easily passable to the function as arguments by reference. Personally I prefer to use closures only when dealing with iterative changes rather than recursive changes.

tjgurwara99 (changes requested):

backtrack/subset.go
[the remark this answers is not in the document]
tjgurwara99: There is one thing I think is worth considering - haven't really thought about this but maybe this behaviour is not completely in the spec:

```Go
var someFunc func(int)

someFunc = func(i int) { someFunc (i-1) }

go func() {
	someFunc = func(i int) {...}
}()

go func() {
	someFunc = func(i int) {...}
}()

someFunc = somethingElse

```
I don't think it is very clear what will happen to the first someFunc call since one of the go routine is changing the value of the function to `someOtherFunc`.

You can argue that you are not doing that in this code and you'd be right. But this is one of those Gotchas that are very difficult to debug in a large application.

Up to you to change this but this pattern of using closure recursion has a big issues that should not be used in production code - since this is a learning repo, this issue cannot be overlooked because people see this magic happening and decide to use this in interesting ways that can potentially result in weird results.

backtrack/permutation.go
```diff
@@ -0,0 +1,41 @@
...
+			if vis[i] || (i > 0 && !vis[i-1] && nums[i] == nums[i-1]) {
+				continue
+			} // get first element that never used/visited
+			perm[depth] = nums[i]
+			vis[i] = true
+			backtrackPerm(depth + 1)
+			vis[i] = false
+		}
+	}
+	backtrackPerm(0)
```
tjgurwara99, line 39: Why not extract this function as a unexported package function - just pass the values by reference.

tjgurwara99: Something like this should work I think

```Go
func btPerm(permutations [][]int, depth, length int, vis []bool, perm []bool) {
  	if depth == length {
  		permutations = append(permutations, append([]int{}, perm...))
 		return
 	}
  	for i := range nums {
  		if vis[i] || (i > 0 && !vis[i-1] && nums[i] == nums[i-1]) {
  			continue
  		} // get first element that never used/visited
  		perm[depth] = nums[i]
  		vis[i] = true
  		backtrackPerm(permutations, depth + 1, length, vis, perm)
  		vis[i] = false
  	}
  }
```
and inside the `Permutation` function, you would have something like this

```Go
func Permutation(nums []int) [][]int {
 	sort.Ints(nums)
 	perm := make([]int, n)
 	vis := make([]bool, n)
 	var permutations [][]int
 	btPerm(permutations, 0, len(nums), vis, perm)
 	return permutations
}
```
Would work since all the slices are references and copying integers is inexpensive.

tjgurwara99 (changes requested):
Consider these changes - I know you've marked this PR as work in progress but thought I should still look over this.

Also, just mark this PR as `draft`.

sixwaaaay (commented):
I am confused, that which way is more readable.

backtrack/permutation.go
```diff
@@ -20,22 +20,22 @@ func Permutation(nums []int) [][]int {
...
-			backtrackPerm(depth + 1)
-			vis[i] = false
-		}
-	}
-	backtrackPerm(0)
+	backtrackPerm(&permutations, nums, 0, n, perm, vis)
 	return permutations
 }
+
+func backtrackPerm(permutations *[][]int, nums []int, depth int, n int, perm []int, vis []bool) {
```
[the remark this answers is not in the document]
tjgurwara99: I don't think you need a pointer to append... I'll open a gitpod Saturday and see if in-fact we do indeed need it but `[][]int` is a slice (internally a pointer) so I'm under the impression that we don't need the pointer to it. But will let you know on Friday as soon as I get some time 😄

backtrack/permutation.go
```diff
@@ -20,22 +20,22 @@ func Permutation(nums []int) [][]int {
...
+	if depth == n {
```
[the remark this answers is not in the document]
tjgurwara99: If you're not passing the `nums` like this `nums[a:b]` for some a, b then yes it can be done that way, and the Len result is also constant so it doesn't change much.

backtrack/tsp.go
```diff
@@ -0,0 +1,46 @@
...
+func TSP(graph [][]int) (int, []int) {
+	n := len(graph)
+	// x is the current solution, bestX is the best solution found so far
+	x, bestX := make([]int, n), make([]int, n)
+	for i := 0; i < n; i++ {
+		x[i] = i
+	} // initialize x as the permutation of the numbers from 0 to n-1
+	bestCost := math.MaxInt32 // best cost
+	var tsp func(int, int)
+	tsp = func(d int, cost int) { // depth, cost so far
```
tjgurwara99, line 26: Same problem with concurrent access applies here... Not sure if this is ideal. Can be solved similarly by passing values by reference?

This pattern exists somewhere I read recently but I personally don't think its a good way since the closure is passing references anyways so having a separate function to receive these values by reference would be best and in fact quite clear as to what is happening (in my opinion)

backtrack/subset.go
```diff
@@ -0,0 +1,48 @@
...
+import "sort"
+
+// Subset is a function to generate all subsets of a given array
+func Subset(nums []int) [][]int {
+	sort.Ints(nums)
+	var subsets [][]int
+	n := len(nums)
+	index := make([]bool, n)
+	var backtrackSubset func(int)
+	backtrackSubset = func(depth int) {
```
tjgurwara99, line 27: Same

tjgurwara99 (changes requested):

backtrack/permutation.go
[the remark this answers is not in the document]
raklaptudirm: Aren't slices passed by reference? Because they themselves are pointers to an underlying array, so I think we do not need to take a pointer.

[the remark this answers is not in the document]
sixwaaaay: Its length cannot be changed, otherwise you will get a new slice.
Pre-calculation may not be a good solution as well, because it is probably a waste of space.
In some languages such as c, a top-level variable may be used here.
And the idea of some languages like java is to use classes here to organize the code, then it puts some variables that need to be passed as fields of the class.
Both of these can easily cause confusion for users.
Closures, on the other hand, achieve the effect of having the above inside a function, which is why I used it in the first place.But it has the disadvantage of not being clear enough.
The two function cases are somewhat complicated by the characteristics of go itself.
go's [][]int doesn't come with a resize method like c++'s vector. It is done through an external method.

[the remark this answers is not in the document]
raklaptudirm: Ah, I see the problem. The function itself is modifying the slice. 👍🏽

PR #471, opened 2022-03-01
Added Square root algorithm using fast inverse square root

i-redbyte:
Computing the square root with binary operations


raklaptudirm (changes requested):
Add an explanation of the algorithm in comments.

raklaptudirm (dismissed):
Ah, you are using the quake III algorithm. It will be quite complicated to describe it in comments, so let it be. You can still add the original comments though XD

math/binary/sqrt.go
```diff
@@ -0,0 +1,26 @@
...
+
+const threeHalves = 1.5
+
+func Sqrt(number float32) float64 {
+	var halfHumber, y float32
+	halfHumber = number * 0.5
+	z := math.Float32bits(number) // floating point bit level hacking
+	z = 0x5f3759df - (z >> 1)     // Newton's approximation
+	y = math.Float32frombits(z)
+	for i := 0; i < 3; i++ {
```
tjgurwara99, line 22: Feel like one is enough here - specially for 32 bit floats

math/binary/sqrt.go
```diff
@@ -0,0 +1,26 @@
...
+		y = y * (threeHalves - (halfHumber * y * y))
+	}
+	return math.RoundToEven(float64(1 / y))
```
tjgurwara99, line 25: You are converting a 32 bit to a 64 bit? Shouldn't it adhere to what the input numbers type?

tjgurwara99: If you're using the float64 then compare it with the `math` package's sqrt as well.

tjgurwara99, line 25: Why `RoundToEven`?

math/binary/sqrt.go
```diff
@@ -0,0 +1,26 @@
...
+
+package binary
+
+import (
+	"math"
+)
+
+const threeHalves = 1.5
+
+func Sqrt(number float32) float64 {
```
tjgurwara99, line 16: I'm curious to see which is faster, the assembly sqrt function vs this one. Assembly one is usually faster in C/C++ but sometime the speed for assembly doesn't translate in Go. I saw that you added the benchmark comparing it to the `math` sqrt - will check it out when I get time.

tjgurwara99 (changes requested):

math/binary/sqrt.go
```diff
@@ -16,8 +16,8 @@ const threeHalves = 1.5
 func Sqrt(number float32) float64 {
 	var halfHumber, y float32
 	halfHumber = number * 0.5
-	z := math.Float32bits(number)
-	z = 0x5f3759df - (z >> 1)
+	z := math.Float32bits(number) // floating point bit level hacking
+	z = 0x5f3759df - (z >> 1)     // Newton's approximation
```
raklaptudirm, line 20: This is the bit hack to get the best approximation from the function, not the newton's approximation.

math/binary/sqrt.go
```diff
@@ -16,8 +16,8 @@ const threeHalves = 1.5
...
 	y = math.Float32frombits(z)
 	for i := 0; i < 3; i++ {
```
raklaptudirm, line 22: This is the newton's approximation, and a single iteration is enough.

math/binary/sqrt.go
raklaptudirm, line 25: I think this really breaks the deal. The Fast inverse square root alogrithm was created precisely to avoid expensive operations like division. I think you should add it as the inverse square root algorithm only.

raklaptudirm (changes requested):

math/binary/sqrt.go
i-redbyte: very good point (corrected)

math/binary/sqrt.go
i-redbyte: accepts only float64 without rounding problems begin.

i-redbyte: math.RoundToEven  accepts only float64 and without rounding, problems begin. I don't know how to solve this problem nicely

tests have BenchmarkMathSqrt measurement

i-redbyte: and if you use the float64 input parameter, then the cast
z := math.Float64bits(number)
z = 0x5f3759df - (z >> 1)

it works incorrectly

math/binary/sqrt.go
i-redbyte: in my PC(MacBook Pro (16-inch, 2019) Intel Core i9 2,3 DDR4 32 Gb):

BenchmarkSquareRootCalculation-16    	1000000000	         0.2363 ns/op
BenchmarkMathSqrt-16    	                         1000000000	         0.2466 ns/op

math/binary/sqrt.go
i-redbyte: sorry, comment moved, fixed

math/binary/sqrt.go
i-redbyte: Done

math/binary/sqrt.go
i-redbyte: > Why `RoundToEven`?

you can use a regular round, but then the benchmarks are worse than when using RoundToEven

raklaptudirm: I don't see how that is relevant?

i-redbyte: if you do not use rounding, then due to the peculiarities of float, we will get results like 1.00000001 or 0.9999999999 and it seems that this is not very convenient for writing tests and checking values

raklaptudirm: Yes, but I don't how that has anything to do with my comment.

i-redbyte: > Yes, but I don't how that has anything to do with my comment.

then apparently I did not understand your comment, sorry. I just wanted to give a nice trick of getting a square root without using library functions sqrt

raklaptudirm: The quake III algorithm is a genius thing, so I wanted to add it as is. @tjgurwara99 @siriak What do you guys think?

tjgurwara99: > math.RoundToEven accepts only float64 and without rounding, problems begin. I don't know how to solve this problem nicely

I think people looking at this algorithm would know the peculiarity of `float`s, so rounding is not necessary I think. The equality of floats is almost always done like so `Abs(actualValue - calculatedValue) < epsilon` for some small enough epsilon (usually depends on the threshold of float size (32 bit or 64 bit) so it makes sense to not have to round the number.

tjgurwara99: See my comment on the other block.

math/binary/sqrt.go
tjgurwara99: Very interesting, I feel like the conversion [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/math/sqrt.go;drc=1d20a362d0ca4898d77865e314ef6f73582daef0;l=92) to assembly instruction that happens in the `math.Sqrt` function is taking the most time but I think I will try to do some more profiling to figure out why. Will be busy this whole coming week so probably won't get it done anytime soon 😓

math/binary/sqrt.go
tjgurwara99: Personally don't think this is bad though. The Quake III game algorithm was built because of the efficiency of calculation of reflections angles that this algorithm provides (if I remember correctly - don't quote me on that though, my physics is bad so... 😂) so if we needed `sqrt` instead of the inverse square root then it makes sense but personally I think this is a valid approach to get an `sqrt` directly.  Also from the benchmarks, it seems to be faster still than the standard library `Sqrt` which in most new architectures amd64, arm64 etc) uses the assembly instructions which should in theory be faster than this so I think it's a good approach.

tjgurwara99: Also, regarding the benchmark, can you make sure that your architecture is one of these mentioned [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/math/sqrt_asm.go;drc=refs%2Ftags%2Fgo1.17.8;l=5)?

i-redbyte: Done

i-redbyte: I tested on two devices(arm64), this architecture is on the list
1)
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz
BenchmarkSquareRootCalculation
BenchmarkSquareRootCalculation-4   	1000000000	         0.3576 ns/op
PASS

2)

goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkSquareRootCalculation
BenchmarkSquareRootCalculation-16      1000000000           0.2569 ns/op
PASS

siriak (approved):

tjgurwara99 (approved):
LGTM

raklaptudirm (approved):

Issue #472, opened 2022-03-08
feat: implementation of find kth element in array


jsrdxzw:
**Description**
An implementation of finding kth largest or smallest element in random array by using quick sort thinking

tjgurwara99:
Closed with #473

PR #473, opened 2022-03-08
feat: implementation of find kth element in array (#472)

jsrdxzw:
An implementation of finding kth largest or smallest element in random array by using quick sort algorithm


sixwaaaay:
Should we randomize this method? Although on average the complexity is linear, however, there is very much data that is close to ordered (or a significant portion of it is ordered), meaning that the complexity is closer to the worst case.

math/kth/kthnumber.go
```diff
@@ -0,0 +1,47 @@
...
+		} else if k > pivot {
+			start = pivot + 1
+		} else {
+			end = pivot
+		}
+	}
+	return -1
+}
+
+func partition(nums []int, start, end int) int {
```
tjgurwara99, line 36: We already have a partition function, export it from the sort package and use that instead of creating a new one.

math/kth/kthnumber.go
```diff
@@ -0,0 +1,47 @@
...
+	end := len(nums)
+	for start < end {
+		pivot := partition(nums, start, end)
+		if k == pivot {
+			return nums[pivot]
+		} else if k > pivot {
+			start = pivot + 1
+		} else {
+			end = pivot
+		}
```
tjgurwara99, line 31: This style of writing conditions is not very "Go-like".

Consider changing it to something like this

```suggestion
		if k == pivot {
			return nums[pivot]
		}
		if k > pivot {
			start = pivot + 1
			continue
		}
		end = pivot
```

math/kth/kthnumber_test.go
```diff
@@ -0,0 +1,91 @@
...
+		k        int
+		expected int
+		name     string
+	}{
+		{
+			input:    []int{6, 7, 0, -1, 10, 70, 8, 22, 3, 9},
+			k:        3,
+			expected: 10,
+			name:     "3th largest number",
+		},
```
tjgurwara99, line 19: This edge case will fail (I think) since there is no difference between what is returned vs the supposed error value

```suggestion
		{
			input:    []int{6, 7, 0, -1, 10, 70, 8, 22, 3, 9},
			k:        3,
			expected: 10,
			name:     "3th largest number",
		},
		{
			input:    []int{-1, -1, -1, -1, -1, -1, -1, -1, -1, -1},
			k:        3,
			expected: -1,
			name:     "3th largest number",
		},
		{
			input:    []int{-1, -1, -1, -1, -1, -1},
			k:        7,
			expected: -1,
			name:     "This should be an error",
		},
```

math/kth/kthnumber.go
```diff
@@ -0,0 +1,47 @@
...
+// FindKthMax use quick sort algorithm to get kth large element in given array
+// return -1 when not found
+func FindKthMax(nums []int, k int) int {
+	index := len(nums) - k
+	return kthNumber(nums, index)
+}
+
+// FindKthMin use quick sort algorithm to get kth small element in given array
+// return -1 when not found
+func FindKthMin(nums []int, k int) int {
```
tjgurwara99, line 12:
```suggestion
func FindKthMin(nums []int, k int) (int, error) {
```

tjgurwara99 (changes requested):
Consider the following suggestions

jsrdxzw:
> Should we randomize this method? Although on average the complexity is linear, however, there is very much data that is close to ordered (or a significant portion of it is ordered), meaning that the complexity is closer to the worst case.

I have just modified my code following the comments, may be you can have a look at `sort.partition` and improve it

math/kthnumber.go
```diff
@@ -0,0 +1,40 @@
+package math
+
+import (
+	"github.com/TheAlgorithms/Go/search"
+	"github.com/TheAlgorithms/Go/sort"
+)
+
+// FindKthMax use quick sort algorithm to get kth large element in given array
+// return -1 when not found
+func FindKthMax(nums []int, k int) (int, error) {
```
tjgurwara99, line 10:
```suggestion
// FindKthMax returns the kth large element in given an integer slice
// with nil `error` if found and returns -1 with `error` `search.ErrNotFound`
// if not found. NOTE: The `nums` slice gets mutated in the process.
func FindKthMax(nums []int, k int) (int, error) {
```

math/kthnumber.go
```diff
@@ -0,0 +1,40 @@
...
+	index := len(nums) - k
+	return kthNumber(nums, index)
+}
+
+// FindKthMin use quick sort algorithm to get kth small element in given array
+// return -1 when not found
+func FindKthMin(nums []int, k int) (int, error) {
```
tjgurwara99, line 17:
```suggestion
// FindKthMin returns kth smallest element in given an integer slice
// with nil `error` if found and returns -1 with `error` `search.ErrNotFound`
// if not found. NOTE: The `nums` slice gets mutated in the process.
func FindKthMin(nums []int, k int) (int, error) {
```

math/kthnumber.go
```diff
@@ -0,0 +1,40 @@
...
+	index := k - 1
+	return kthNumber(nums, index)
+}
+
+func kthNumber(nums []int, k int) (int, error) {
```
tjgurwara99, line 22: Add a simple explanation about how it connects with quick sort.

tjgurwara99 (changes requested):

tjgurwara99:
Also could you enable GitHub actions in your fork? There are some auto documentation tools that would write to the README file.

tjgurwara99 (approved):
Looks good! Thank you for contributing to this project 😄

raklaptudirm (approved):

PR #474, opened 2022-03-16
package name geometry

i-redbyte:
add package description for further filling with geometric algorithms


math/geometry/straightlines.go
```diff
@@ -27,27 +28,27 @@ func Section(p1, p2 *Point, r float64) Point {
...
 
-// Calculates the slope (gradient) of a line.
+// Slope Calculates the slope (gradient) of a line.
 func Slope(l *Line) float64 {
 	return (l.P2.Y - l.P1.Y) / (l.P2.X - l.P1.X)
 }
 
-// Calculates the Y-Intercept of a line from a specific Point.
+// Intercept Calculates the Y-Intercept of a line from a specific Point.
 func Intercept(p *Point, slope float64) float64 {
```
tjgurwara99, line 37: Can you call this `YIntercept` instead of just `Intercept`? Because we can also have `XIntercept`s and this would be confusing.

tjgurwara99 (changes requested):

math/geometry/straightlines.go
i-redbyte: Done

math/geometry/straightlines.go
```diff
@@ -1,10 +1,11 @@
+// Package geometry contains geometric algorithms
 package geometry
 
 import (
 	"math"
 )
 
-// Defines a point with x and y coordinates.
+// Point Defines a point with x and y coordinates.
```
raklaptudirm, line 8:
```suggestion
// Point defines a point with x and y coordinates.
```
Do this for all the places you have added the name.

raklaptudirm (changes requested):

math/geometry/straightlines.go
i-redbyte: Done

raklaptudirm (approved):

tjgurwara99 (approved):

tjgurwara99:
It seems the workflow is frozen again - @i-redbyte could you maybe try to add an empty commit?

`git commit --allow-empty -m <message>` That would launch the action and should theoretically work

i-redbyte:
> It seems the workflow is frozen again - @i-redbyte could you maybe try to add an empty commit?
> 
> `git commit --allow-empty -m <message>` That would launch the action and should theoretically work

done

PR #475, opened 2022-03-18
Feat/cos

i-redbyte:
add cosine function algorithm


math/geometry/trigan/main_function.go
```diff
@@ -0,0 +1,17 @@
+// Package trigan for trigonometric functions and algorithms
+package trigan
+
+import "math"
+
+// Cos  returns the cosine of the radian argument x. [See more](https://en.wikipedia.org/wiki/Sine_and_cosine)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see sqrt_test.go
+
+func Cos(x float64) float64 {
```
raklaptudirm, line 10: Add comments explaining your code.

math/geometry/straightlines.go
```diff
@@ -1,10 +1,11 @@
+// Package geometry contains geometric algorithms
```
raklaptudirm, line 1: I believe these changes were already merged.

raklaptudirm (approved):

math/geometry/trigan/main_function.go
```diff
@@ -0,0 +1,17 @@
+// Package trigan for trigonometric functions and algorithms
+package trigan
+
+import "math"
+
+// Cos  returns the cosine of the radian argument x. [See more](https://en.wikipedia.org/wiki/Sine_and_cosine)
+// author(s) [red_byte](https://github.com/i-redbyte)
+// see sqrt_test.go
+
```
tjgurwara99, line 9: Remove this line

math/geometry/trigan/main_function.go
```diff
@@ -0,0 +1,17 @@
...
+func Cos(x float64) float64 {
+	tp := 1. / (2. * math.Pi)
+	x *= tp
+	x -= .25 + math.Floor(x+.25)
+	x *= 16. * (math.Abs(x) - .5)
+	x += .225 * x * (math.Abs(x) - 1.) //Extra precision
+	return x
```
tjgurwara99, line 16: Good algorithm. Just a note though - I think you should make your `float`s explicit. Like `1.0` instead of `1.` Everyone  who has coded would know about this but its better to be explicit still. Adding a `0` pre/post a number doesn't hurt.

Second, I think we already have the `math.Abs` and `math.Floor` in the `math` package - kind of a redundant thing if you already have `Cos` in the standard library.

Also, `Cos` is not only relevant to geometry but rather the whole math package - so should live in the math package.

tjgurwara99 (changes requested):

math/geometry/trigan/main_function.go
i-redbyte: Done

math/geometry/trigan/main_function.go
i-redbyte: Added a link to the idea that I took as a basis

math/geometry/trigan/main_function.go
i-redbyte: Good! Done

math/geometry/trigan/main_function.go
i-redbyte: Done

math/geometry/trigan/main_function.go
tjgurwara99: > Also, Cos is not only relevant to geometry but rather the whole math package - so should live in the math package.

I think you missed this part of the comment...

tjgurwara99 (changes requested):

i-redbyte:
I'm sorry, my commit with edits, for some reason, didn't pass right away and I didn't notice it

math/geometry/trigan/main_function.go
i-redbyte: I'm sorry, my commit with edits, for some reason, didn't pass right away and I didn't notice it

math/cos_test.go
```diff
@@ -0,0 +1,44 @@
+package math
+
+import (
+	"math"
+	"testing"
+)
+
+const accuracy = 0.001
```
tjgurwara99, line 8: call this `epsilon` instead

math/cos_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func BenchmarkCos(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		Cos(180)
+	}
+}
+
+func BenchmarkMathCos(b *testing.B) {
```
tjgurwara99, line 40: Just to point out that the reason why the Math Cos is slow is because the epsilon tolerance there is quite high. So I think you should add a comment here to make sure people are aware of this fact.

tjgurwara99 (changes requested):
Other than these two comments, everything else seems to be good 😄

math/cos_test.go
i-redbyte: Done

math/cos_test.go
i-redbyte: Done

math/cos_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func BenchmarkCos(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		Cos(180)
+	}
+}
+
+// A function math.Cos is slower because it calculates a more accurate value
```
tjgurwara99, line 40:
```suggestion
// BenchmarkMathCos is slower because the standard library `math.Cos` calculates a more accurate value.
```

tjgurwara99 (approved):
All good now.

siriak (approved):

PR #476, opened 2022-03-20
Feat/sin

i-redbyte:
added sin function


math/cos_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 )
 
-const epsilon = 0.001
+const epsilonCos = 0.001
```
tjgurwara99, line 8: Why different? Doesn't this mean that this is not a good algorithm that it changes the precision?

i-redbyte: I'm sorry, this is a poorly written test by me

i-redbyte: Fix

math/sin_test.go
```diff
@@ -0,0 +1,44 @@
+package math
+
+import (
+	"math"
+	"testing"
+)
+
+const epsilonSin = 0.001
```
tjgurwara99, line 8: Call it `epsilon` and remove the other `epsilonCos` because they are both the same.

math/sin_test.go
```diff
@@ -0,0 +1,44 @@
+package math
```
tjgurwara99, line 1:
```suggestion
package math_test
```

math/cos_test.go
tjgurwara99, line 8: Change the package name to `math_test` in this file as well.

tjgurwara99 (changes requested):

math/sin_test.go
i-redbyte: create new package math_test?

math/sin_test.go
i-redbyte: Done

math/cos_test.go
i-redbyte: > Change the package name to `math_test` in this file as well.

create new package math_test?

tjgurwara99: Are you not aware of the `_test` feature of Go tests?

The following is an explanation for why there is a distinction - ignore it if you already know the difference between `package a` and `package a_test`. 

There are a few things to consider when writing tests (at least for me). First thing to consider is whether the test is an integration test or unit test. Unit tests usually live inside the package they were developed (in this case `math`). Integration tests however are a bit more technical, where you have the option to choose the same package or a different package (Go allows you to have another package in the same directory with `_test` suffix). Now the tests that you have are integration tests because they are being performed on exported function (think public methods in Java).

In practice, your methods are simple so it doesn't matter what kind of tests you do but I have been thinking recently to change a few things in the repo (whenever I get some time). But having a separate package (in the same folder with `_test` suffix) makes it easy to test API behaviour.


Consider as an example:


```Go
package something

func someInternalFunction(input int) error { ... }

func PublicApiFunction() error {
    input := <calculated_input>
    return someInternalFunction(input)
}

```

Now to test the `someInternalFunction`, you would need to write test inside the `something` package because its only available to the members of this package - the tests are members of this package since we have `package something` in the beginning of our test files. Note that you can also test the `PubliApiFunction` directly inside this package as well however, from a Public API perspective, we don't care about what the internal behaviour of the `PublicApiFunction` is and so there might be situations where the behaviour of our `PublicApiFunction` is different outside the package. This is where integration tests come in, and you can use the `package something_test` declaration in the beginning of a file in the same directory as you package and import the `something` package how you would in your code outside the package. This kind of test ensures that there is no leak in our Public API logic and would behave as we expect would it to behave.

i-redbyte: > 

Thank you very much! Done

math/sin_test.go
i-redbyte: Thank you ! Done

math/cos_test.go
```diff
@@ -1,11 +1,12 @@
...
+package math_test
 
 import (
+	. "github.com/TheAlgorithms/Go/math"
 	"math"
 	"testing"
 )
 
-const epsilon = 0.001
+const Epsilon = 0.001
```
tjgurwara99, line 9: This should be lowercase.

math/cos_test.go
```diff
@@ -1,11 +1,12 @@
-package math
+package math_test
 
 import (
+	. "github.com/TheAlgorithms/Go/math"
```
tjgurwara99, line 4: This is a bad practice. Just use the math package instead of importing everything in the `_test` space - call it `algmath` instead or something. Better yet change the standard library import name to `stdmath "math"` instead. Importing without a namespace is a bad practice in general whether its Go or some other language.

tjgurwara99 (changes requested):

math/cos_test.go
i-redbyte: > This is a bad practice. Just use the math package instead of importing everything in the `_test` space - call it `algmath` instead or something. Better yet change the standard library import name to `stdmath "math"` instead. Importing without a namespace is a bad practice in general whether its Go or some other language.

Good! poke me right?! I do not understand

tjgurwara99:
```suggestion
	algmath "github.com/TheAlgorithms/Go/math"
```

and when calling it in this package, you use `algmath` because its confined to the package scoped `algmath`
 like so

```Go
algmath.Cos(...)
```

The concept of namespacing was added to C++ in 1995 and the idea for this was to prevent name collisions and grouping similar logical code into blocks (This `std::vector` in the C++ standard library - `std` is the namespace for standard library utilities). The idea is the same as packages in Go or modules in Python. The reason is that now, if you were to define a function `Sin` or `Cos` in the `math_test` package then you would have errors. I also understand that this is a small package and thus you are unlikely to come across these issues but in big packages it is actually an issue (at least in  code bases I've worked on).

For more information, I prefer a lot – *not all* – of the Go styling conventions from [this](https://github.com/uber-go/guide/blob/master/style.md#import-aliasing)

math/cos_test.go
i-redbyte: Done

math/cos_test.go
i-redbyte: Yes, great, now I understand that you meant the namespace, this is very correct. Corrected it

tjgurwara99 (approved):

raklaptudirm (approved):

PR #477, opened 2022-03-23
Change `x/2` to `x>>1` for better performance

JustZyx:


raklaptudirm (approved):

tjgurwara99:
TBH I like this optimisation but if you run the benchmark on a function (call it `midPoint` or something) you will see that there is actually no difference between them - at least on my hardware there isn't. However, maybe there is a better way to show both the old way and the new way without claiming that either one is faster than the other. Proof matters when there is a claim that one approach is faster/better than the other. I have a way in which you can get a good proof of what I'm trying to say. Consider this method - for the sake of learning

```Go

type midFunc func(int, int) int

var midSimple := func(left, right int) int {
	return left + (right - left)/2
}

var midBitShift := func(left, right int) int {
	return left + (right-left) >> 1
}

var mid midFunc = midBitShift // this makes it easy to change in tests + its not public so changing it is safe.

func Binary(array []int, target int, lowIndex int, highIndex int) (int, error) {
	...
	m := mid(lowIndex, highIndex)
	...
}
```
and use this function to get the mid in the functions. The benefit of this method is that we can actually benchmark both ways (old and new) without issues.

To show the effectiveness of this method, we can then have two more benchmarks like this

```Go

func benchMarkBinary(b *testing.B, bin func([]int, int, int, int) (int, error)) {
	testCase := generateBenchmarkTestCase()
	b.ResetTimer() // this is important because the generateBenchmarkTestCase() is expensive
	for i := 0; i < b.N; i++ {
		_, _ = bin(testCase, i, 0, len(testCase)-1)
	}
}

func BenchmarkBinaryBitShift(b *testing.B) {
	benchMarkBinary(b, Binary)
}

func BenchmakrBinarySimple(b *testing.B {
	mid = midSimple
	defer func() {
		mid = midBitShift
	}() // to return this back to the original state
	benchMarkBinary(b, Binary) // this will test with 
}
```

This would show that both approaches are almost the same but at the same time, I would like to keep both approaches.

raklaptudirm:
[This](https://stackoverflow.com/questions/10681375/which-is-better-option-to-use-for-dividing-an-integer-number-by-2) may be relevant.

JustZyx:
> TBH I like this optimisation but if you run the benchmark on a function (call it `midPoint` or something) you will see that there is actually no difference between them - at least on my hardware there isn't. However, maybe there is a better way to show both the old way and the new way without claiming that either one is faster than the other. Proof matters when there is a claim that one approach is faster/better than the other. I have a way in which you can get a good proof of what I'm trying to say. Consider this method - for the sake of learning
> 
> ```go
> type midFunc func(int, int) int
> 
> var midSimple := func(left, right int) int {
> 	return left + (right - left)/2
> }
> 
> var midBitShift := func(left, right int) int {
> 	return left + (right-left) >> 1
> }
> 
> var mid midFunc = midBitShift // this makes it easy to change in tests + its not public so changing it is safe.
> 
> func Binary(array []int, target int, lowIndex int, highIndex int) (int, error) {
> 	...
> 	m := mid(lowIndex, highIndex)
> 	...
> }
> ```
> 
> and use this function to get the mid in the functions. The benefit of this method is that we can actually benchmark both ways (old and new) without issues.
> 
> To show the effectiveness of this method, we can then have two more benchmarks like this
> 
> ```go
> func benchMarkBinary(b *testing.B, bin func([]int, int, int, int) (int, error)) {
> 	testCase := generateBenchmarkTestCase()
> 	b.ResetTimer() // this is important because the generateBenchmarkTestCase() is expensive
> 	for i := 0; i < b.N; i++ {
> 		_, _ = bin(testCase, i, 0, len(testCase)-1)
> 	}
> }
> 
> func BenchmarkBinaryBitShift(b *testing.B) {
> 	benchMarkBinary(b, Binary)
> }
> 
> func BenchmakrBinarySimple(b *testing.B {
> 	mid = midSimple
> 	defer func() {
> 		mid = midBitShift
> 	}() // to return this back to the original state
> 	benchMarkBinary(b, Binary) // this will test with 
> }
> ```
> 
> This would show that both approaches are almost the same but at the same time, I would like to keep both approaches. So makes it possible to keep both approaches.

You are right.  From the below command, we can see assembly of the new way is same to the old way.
`
go tool compile -N -l -S filename.go
`
It seems go compiler eliminates the difference between dividing and bitshift by optimization.
Now that performance is no difference,  bitshift is no need because of it's poor readability.

PR #479, opened 2022-03-23
Maintenance: Package restructure to prepare for generics

tjgurwara99:
Generics are out and I thought we should restructure the repository in a better way making it easier to navigate. I've made a few changes and would appreciate your input on the next restructuring that I will be doing so wanted to get the easy bits out of the way in this PR.


cipher/caesar_test.go
```diff
@@ -1,11 +1,13 @@
-package caesar
+package cipher_test
```
raklaptudirm, line 1: Maybe add the new fuzzing tests for the ciphers, by encrypting and then decrypting and checking equality with the cleartext?

cipher/diffiehellman.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// GenerateShareKey : generates a key using client private key , generator and primeNumber
+// this key can be made public
+// shareKey = (g^key)%primeNumber
+func GenerateShareKey(prvKey int64) (int64, error) {
+	return modular.Exponentiation(generator, prvKey, primeNumber)
+}
+
+// GenerateMutualKey : generates a mutual key that can be used by only alice and bob
```
raklaptudirm, line 17: Probably add a wiki link. The reference to Alice and Bob may not be understood by everyone (inside jokes XD)

cipher/errors.go
```diff
@@ -0,0 +1,10 @@
+package cipher
+
+import "errors"
+
+// ErrorFailedToEncrypt Raised when a cipher function fails to encrypt the message
+var ErrorFailedToEncrypt = errors.New("failed to Encrypt")
+
+var NoTextToEncryptError = errors.New("no text to cipher")
```
raklaptudirm, line 8:
```suggestion
var ErrorNoText = errors.New("cipher: no text to encrypt")
```

cipher/errors.go
```diff
@@ -0,0 +1,10 @@
+package cipher
+
+import "errors"
+
+// ErrorFailedToEncrypt Raised when a cipher function fails to encrypt the message
+var ErrorFailedToEncrypt = errors.New("failed to Encrypt")
```
raklaptudirm, line 6:
```suggestion
var ErrorEncrypt = errors.New("cipher: failed to encrypt")
```
Maybe even change this to a struct with the necessary data instead of a simple string.

cipher/errors.go
```diff
@@ -0,0 +1,10 @@
...
+
+var NoTextToEncryptError = errors.New("no text to cipher")
+
+var KeyMissingError = errors.New("no key provided")
```
raklaptudirm, line 10:
```suggestion
var ErrorMissingKey = errors.New("cipher: no key provided")
```

raklaptudirm (commented):
General comments:
- Explanation links were removed from some cipher's comments. Add them.
- Use the new fuzzing tests to test both the encryption and decryption of the ciphers.

.github/workflows/golang_lint_and_test.yml
```diff
@@ -4,14 +4,12 @@ on: [push, pull_request]
...
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
       - uses: actions/checkout@v2
       - name: Codespell
         uses: codespell-project/actions-codespell@master
         with:
-          ignore_words_list: "actualy,nwe"
+          ignore_words_list: "actualy,nwe,caesar"
```
siriak, line 12: Why are actualy and nwe ignored?

siriak (approved):

.github/workflows/golang_lint_and_test.yml
tjgurwara99: The `actualy` is in fact `actualY` in the code🤷🏼‍♂️ the code spell has some issues like that.

cipher/caesar_test.go
tjgurwara99: I am in favour of using `FuzzXxx` but from what I have seen - they are a bit slow, plus I won't have time to do any significant improvement for a while other than small refactoring 😓.

Maybe we should open a separate issue so that contributors can add Fuzz tests.

raklaptudirm: Sure. Though I also can't work on them right now with exams going on. But they will be useful in the future.

tjgurwara99: Yeah, don't worry - it would be open to all contributors not just maintainers with a `help-wanted` tag and hopefully if someone sees it then they can contribute if not, we – maintainers – can add them whenever we get some time.

cclauss:
This looks like a lot of changes for a single PR.  Could we make a main PR that sets us up for generics and then separate PRs for individual algorithms?

tjgurwara99:
> This looks like a lot of changes for a single PR. Could we make a main PR that sets us up for generics and then separate PRs for individual algorithms?

Sure, I'll create some interfaces to get us started with Generics, and slowly raise small PR's for improving the repo.

PR #482, opened 2022-03-25
Add LowerBound and UpperBound algorithms

JustZyx:
Inspired by c++ stl in `<algorithm>`: [lower_bound](http://www.cplusplus.com/reference/algorithm/lower_bound/) and [upper_bound](http://www.cplusplus.com/reference/algorithm/upper_bound/)


search/binary.go
```diff
@@ -36,3 +36,45 @@ func BinaryIterative(array []int, target int, lowIndex int, highIndex int) (int,
 	}
 	return -1, ErrNotFound
 }
+
+//Returns index to the first element in the range [lowIndex, highIndex] that is not less than (i.e. greater or equal to) target, or return -1 and ErrNotFound if no such element is found.
+func LowerBound(array []int, target int, lowIndex int, highIndex int) (int, error) {
```
tjgurwara99, line 41: In Go, we have access to the `len` function therefore it is unnecessary to have the low and high index in the function call. If a user wishes to do a subarray search, then they can pass the sub array like so `arr[:someIdx]` which changes the `len` I believe - have to check to make sure actually. But in any case, this could be simplified to this 😄 

 ```suggestion
func LowerBound(array []int, target int) (int, error) {
```

and use `highIndex := len(array) - 1` inside the function

File: search/binary.go [hunk 2]
```diff
@@ -36,3 +36,45 @@ func BinaryIterative(array []int, target int, lowIndex int, highIndex int) (int,
...
+
+	//when target greater than every element in array, startIndex will out of bounds
+	if startIndex > highIndex {
+		return -1, ErrNotFound
+	}
+	return startIndex, nil
+}
+
+//Returns index to the first element in the range [lowIndex, highIndex] that is greater than target, or return -1 and ErrNotFound if no such element is found.
+func UpperBound(array []int, target int, lowIndex int, highIndex int) (int, error) {
```
tjgurwara99, line 62: Same applies here

tjgurwara99 (changes requested):
Nice!

Please consider the changes that I've suggested. Do let me know if you disagree with anything that I mentioned here 😄

tjgurwara99:
You also need to format your code correctly - look at the autogenerated comment in the PR for more information. Let us know if you have any troubles 😄

search/binary.go
JustZyx: Actually, I agree with your suggestion. 
I write by this way because for following before code style as below.
`func Binary(array []int, target int, lowIndex int, highIndex int) (int, error)`
`func BinaryIterative(array []int, target int, lowIndex int, highIndex int) (int, error)`
Change all or keep this style. I'm a little confused, which dou you think is better.

tjgurwara99: You can change the previous one as well. I think its best to be consistent.

JustZyx: Except `func Binary`, all changed.  Pass the sub array like so `arr[:someIdx]`, that lowIndex and highIndex it's blong to `arr[:someIdx]` not `arr []int`. For example:
array := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
target := 10
the lowIndex is still `0` and highIndex is `len(arr[x:y])-1`, so the mid is not index in array []int.

tjgurwara99 (approved):

raklaptudirm (approved):

PR #484, opened 2022-03-27
Add Deterministic Miller-Rabin Primality Test

raklaptudirm:
### Tasks
- [x] Added deterministic version of Miller-Rabin test for valid `int64` numbers. Note the algorithm can still fail due to overflow while modular exponentiation.
- [x] Refactor the existing code.
- [x] Comment code and fix some old comments.


math/prime/millerrabinprimalitytest.go
```diff
@@ -14,25 +15,48 @@ import (
...
 
-// MillerTest This is the intermediate step that repeats within the
-// miller rabin primality test for better probabilitic chances of
-// receiving the correct result.
-func MillerTest(d, num int64) (bool, error) {
-	random := rand.Int63n(num-1) + 2
+// isTrivial checks if num's primality is easy to determine.
+// If it is, it returns true and num's primality. Otherwise
+// it returns false and false.
+func isTrivial(num int64) (trivial bool, prime bool) {
```
tjgurwara99, line 33:
```suggestion
func isTrivial(num int64) (bool, bool) {
```

This function is very confusing in its current design - why do we care whether this is trivial or not? We should only care about whether this is a prime or not.

Furthermore, you are using named return types - there is no readability benefit for them usually (in fact I think you're function signature is not right for its purpose). The main benefit of using named return types is pre-allocation (a compile time optimisation) but you are neither assigning any values to these or using them in any specific way making this redundant.

math/prime/millerrabinprimalitytest_test.go
```diff
@@ -23,7 +18,17 @@ func TestMillerRabinTest(t *testing.T) {
...
+			if err != test.err {
+				t.Errorf("For input: %d, unexpected error: %v, expected error: %v", test.input, err, test.err)
+			}
+			if output != test.expected {
+				t.Errorf("For input: %d, expected: %v", test.input, output)
+			}
+		})
+
+		t.Run(test.name, func(t *testing.T) {
+			output, err := MillerRabinDeterministic(test.input)
```
tjgurwara99, line 31: separate out the tests - they should be independent

math/prime/millerrabinprimalitytest_test.go
```diff
@@ -36,6 +41,10 @@ func TestMillerRabinTest(t *testing.T) {
 
 func BenchmarkMillerRabinPrimalityTest(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		_, _ = MillerRabinTest(23, 5)
+		_, _ = MillerRabinProbabilistic(23, 5)
+	}
+
+	for i := 0; i < b.N; i++ {
```
tjgurwara99, line 47: Separate out benchmarks please - they skew the actual results.

tjgurwara99 (changes requested):

math/prime/millerrabinprimalitytest.go
```diff
@@ -14,25 +15,48 @@ import (
...
-func findRD(num int64) (int64, int64) {
-	r := int64(0)
+// formatNum accepts a number and returns the
+// odd number d such that num = 2^s * d + 1
+func formatNum(num int64) (int64, int64) {
+	s := int64(0)
 	d := num - 1
 	for num%2 == 0 {
 		d /= 2
-		r++
```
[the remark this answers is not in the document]
raklaptudirm: Wikipedia uses the `s` notation for this, so I changed it since we link to Wikipedia as a reference.

math/prime/millerrabinprimalitytest.go
raklaptudirm: > This function is very confusing in its current design - why do we care whether this is trivial or not? We should only care about whether this is a prime or not.

`isTrivial` checks if a number's primality can be trivially determined. For example, all multiples of 2, other than 2 itself, are not primes, so we do not have to run the Miller-Rabin test in that case (also you can't write even numbers as `2^s * d + 1`). Also, the primality of 2 and 3 can be determined trivially as prime. If we can determine the primality without running the tests, the function will return true and the primality of the number. Otherwise, we have to run the test since we can't determine it's primality, so it returns false and false in that case.

> Furthermore, you are using named return types - there is no readability benefit for them usually (in fact I think you're function signature is not right for its purpose). The main benefit of using named return types is pre-allocation (a compile time optimisation) but you are neither assigning any values to these or using them in any specific way making this redundant.

Named returned types appear in the docs, so they are useful as documentation.

math/prime/millerrabinprimalitytest_test.go
raklaptudirm: Done.

math/prime/millerrabinprimalitytest_test.go
raklaptudirm: Done.

zhrgithub:
done


| |
zhr
|
|
***@***.***
|
签名由网易邮箱大师定制


在2022年03月28日 13:15，Rak ***@***.***> 写道：

@raklaptudirm commented on this pull request.

In math/prime/millerrabinprimalitytest_test.go:

> @@ -23,7 +18,17 @@ func TestMillerRabinTest(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			output, err := MillerRabinTest(test.input, test.rounds)
+			output, err := MillerRabinProbabilistic(test.input, test.rounds)
+			if err != test.err {
+				t.Errorf("For input: %d, unexpected error: %v, expected error: %v", test.input, err, test.err)
+			}
+			if output != test.expected {
+				t.Errorf("For input: %d, expected: %v", test.input, output)
+			}
+		})
+
+			output, err := MillerRabinDeterministic(test.input)


Done.

tjgurwara99 (changes requested):

raklaptudirm:
I am a modular approach guy so since we were repeating those tests, I made a function out of them.

siriak (approved):

tjgurwara99:
> I am a modular approach guy so since we were repeating those tests, I made a function out of them.

Where?

raklaptudirm:
I was talking about the `isTrivial` function. Since it is required in both the deterministic and the probabilistic tests.

tjgurwara99:
> I was talking about the isTrivial function.

I disagree with that statement actually. Your approach is concise, not clean imo (there is a huge difference between these two). I think you might disagree with this but I really think you have a different understanding of what Modularity is.

Modular in the context of software engineering is vague to begin with, so before that I would like to define what I mean by it. Googling the terms gives us a few definitions of Modularity - one of which is here which corresponds to my understanding `In software engineering, modularity refers to the extent to which a software/Web application may be divided into smaller modules.` This definition was taken from [here](https://www.techopedia.com/definition/24772/modularity).

Now, with this in mind, I would argue that your code is less modular than the suggestion I made. The division of the functions into two functions distributed with respect their responsibility, then I think you will agree. With that said however, there are still things that we should consider that goes hand in hand with this principle of modularity. Its called the single responsibility principle, the idea is quite straight forward but very powerful when it comes to software engineering - I really think you should study it. It's the corner stone of SOLID principles defined by Uncle Bob.

Anyways, I leave it up to you whether you want to change that or not, I think you should.

raklaptudirm:
Well, we have 3 maintainers for a reason :). @siriak  What do you think?

math/prime/millerrabinprimalitytest.go
siriak: I'd argue that the SRP is not violated here because the purpose of this function is not to determine primality, but only **try** to do that. And it returns the result and the flag if it has succeeded or not. This is a common pattern in other languages for operations that can fail (e.g. parsing a string to a number)

siriak (approved):
Overall I like this design, the `isTrivial` function could be split further as suggested, but I think it's fine as-is because the operations performed are trivial.

math/prime/millerrabinprimalitytest.go
tjgurwara99: Fair enough, if you disagree then majority rules. If you're talking about the pattern of result and a flag, then it still doesn't follow idiomatic Go calling structure (`res, ok` rather than `ok, res`).

Anyways, I feel like this is a redundant thing since majority agrees but I still think SRP is violated. The `ok` flag is valid pattern in Go as well but trivial flag is by its nature testing something else rather than testing whether this is a prime (just thinking of the naming of the function itself points that out) so the pattern doesn't really apply here - note that there is separate condition testing for completely separate thing - which I think violates the principle but... I concede.

tjgurwara99 (dismissed):
Majority agrees with this change so here's my approval 😄

math/prime/millerrabinprimalitytest.go
raklaptudirm: Changing it to follow the comma ok idiom is a good idea, will make the changes.

raklaptudirm: @tjgurwara99 [Here](https://github.com/golovers/effective-go#named-result-parameters) is an excerpt from a document I quite like, which explains my usage of named returns.

raklaptudirm: Also, I made the code more concise using named returns.

tjgurwara99: > @tjgurwara99 [Here](https://github.com/golovers/effective-go#named-result-parameters) is an excerpt from a document I quite like, which explains my usage of named returns.

Will take a look 😄

raklaptudirm:
@tjgurwara99 @siriak ping

tjgurwara99 (approved):

siriak (approved):

PR #485, opened 2022-03-27
Fix: restructure checksum package

tjgurwara99:
Checksum package restructuring. Will slowly raise small PRs for each package to do some cleanup that we need to do to make it easier to work with 😄


checksum/crc8.go
```diff
@@ -8,7 +8,7 @@
 // see crc8_test.go
 
 // Package crc describes algorithms for finding various CRC checksums
```
raklaptudirm, line 10: Change/Remove this package comment.

checksum/crc8.go
```diff
@@ -22,8 +22,8 @@ type CRCModel struct {
 	Name   string
 }
 
-// CalculateCRC8 This function calculate CRC8 checksum.
-func CalculateCRC8(data []byte, model CRCModel) uint8 {
+// CRC8 This function calculate CRC8 checksum.
```
raklaptudirm, line 25:
```suggestion
// CRC8 calculates CRC8 checksum of the given data.
```

checksum/luhn.go
```diff
@@ -7,8 +7,8 @@
 // Package checksum describes algorithms for finding various checksums
 package checksum
 
-// LuhnAlgorithm This function calculates the checksum using the Luna algorithm
-func LuhnAlgorithm(s []rune) bool {
+// Luhn This function calculates the checksum using the Luhn algorithm
```
raklaptudirm, line 10:
```suggestion
// Luhn validates the provided data using the Luhn algorithm.
```

checksum/luhn.go
```diff
@@ -7,8 +7,8 @@
...
+func Luhn(s []rune) bool {
```
raklaptudirm, line 11:
```suggestion
func Luhn(s []byte) bool {
```
More space efficient and useful.

raklaptudirm (changes requested):

checksum/luhn.go
tjgurwara99: I don't understand what you mean by this?

tjgurwara99: I think your comment about the more used type may or may not be valid (I actually don't know - it also depends on which codebase you're taking about) but that has nothing to do with you initial comment which was about `[]byte` being

> More space efficient and useful.

I don't see the relevance of the first comment and the next comment?

Having said that, `[]rune` is equally as efficient, I can show you benchmarks for it if you'd like. If by space, you mean the memory taken up by it, it would take up the same amount of memory because size of a string is fixed so memory taken up by converting them to bytes or runes will be the same (from what I can tell, unless I'm mistaken - if you have evidence to the contrary, I would like to know 😄)

Also, something never being used doesn't mean that it is better or worse than other things that's used often. If we're talking about idiomatic coding then I personally am not aware of any design decisions directing people to use `[]byte` or `[]rune` - again, if you have evidence to the contrary, then please let me know, I prefer to keep my code idiomatic 😄 but as far as I'm aware there is no idiom prohibiting the use of rune slice.

raklaptudirm: > If by space, you mean the memory taken up by it, it would take up the same amount of memory because size of a string is fixed so memory taken up by converting them to bytes or runes will be the same (from what I can tell, unless I'm mistaken - if you have evidence to the contrary, I would like to know 😄)

The comment about space efficiency was due to the fact that `byte` is an alias for `uint8` while `rune` is one for `int32`, so each "character" occupies more space when stored as a rune.

tjgurwara99: > Also, since each character in our string is supposed to be a decimal digit, using a rune is redundant as all the decimal numbers are in the ascii range(48-57) which can be represented by a byte.

The above sentence explains your whole point now - I understand what you're trying to say, and I agree with the fact that we don't need `rune` handling. Thanks for noticing! Will make my changes soon.

On the note of efficiency - I would just put this here for you to note that there will be no difference in size for both `[]byte` and `[]rune` for ascii characters - [here](https://go.dev/play/p/3yxkhCaeegF) you can see that in terms of memory there would be no difference, the compiler does its job of managing the memory well.

raklaptudirm: > On the note of efficiency - I would just put this [here](https://go.dev/play/p/3yxkhCaeegF) for you to note that there will be no difference in size for both []byte and []rune for ascii characters - here you can see that in terms of memory there would be no difference, the compiler does its job of managing the memory quite well 😄.

Quoting from the `unsafe` package:
> The size does not include any memory possibly referenced by x. For instance, if x is a slice, Sizeof returns the size of the slice descriptor, not the size of the memory referenced by the slice.

The slice descriptor contains the slice length, capacity, and a pointer to an underlying array. So this comparison is actually wrong since the actual memory occupied by the underlying array is not taken into consideration. [Here](https://go.dev/play/p/SaiSKbufNWj), I have used arrays for comparison and the results are completely different.

> if I sound difficult its not that I am against making changes or combative towards other ideas but rather I really would like to know whether what I know is incorrect and try to correct myself and my understanding to improve 😄

Not at all, these discussions are usually more productive than just going with what each of us thinks is right :D

tjgurwara99: Oh wow this makes so much sense! Thanks for pointing it out, I never knew that about the unsafe package - I always got slice sizes to be 24 🤣

Anyways, will make the change now

raklaptudirm (approved):

raklaptudirm:
@tjgurwara99 make an empty commit, the tests are stuck again

siriak (approved):

PR #492, opened 2022-04-12
adding parenthesis checking algorithm 

89453728:
I have code parenthesis checking algorithm in **strings/search/pharentesis.go** with self testing in **strings/search/pharentesis_test.go** 

The objective of the algorithm is to check if every opened parenthesis is closed well.


strings/search/pharentesis.go
```diff
@@ -0,0 +1,22 @@
+package search
+
+func Pharentesis(text string) bool {
```
raklaptudirm, line 3:
```suggestion
func Parenthesis(text string) bool {
```

strings/search/pharentesis.go
```diff
@@ -0,0 +1,22 @@
...
+	var phar_counter int
```
raklaptudirm, line 4: Use lowerCamelCase for variable names. https://go.dev/doc/effective_go#names

strings/search/pharentesis_test.go
```diff
@@ -0,0 +1,67 @@
+package search
+
+import (
+	"reflect"
+	"testing"
+	"fmt"
+)
+
+type testCase struct {
```
raklaptudirm, line 9: No need to declare a separate type as this will only be used once. You can encode the struct directly in the struct literal.

strings/search/pharentesis_test.go
```diff
@@ -0,0 +1,67 @@
...
+		true,
+	},
+	{
+		"<inverted pharentesis>",
+		")()()()()(",
+		false,
+	},
+}
+
+func testPharentesis (t *testing.T){
```
raklaptudirm, line 48:
```suggestion
func TestParenthesis (t *testing.T){
```

raklaptudirm (changes requested):
Looks like you are not familiar with go conventions. I will suggest reading [this](https://go.dev/doc/effective_go) to familiarize yourself with them. Also, format your code with `gofmt`.

strings/search/pharentesis_test.go
89453728: yes, sorry for not knowing about the agreement. I just made the necessary fixes, I don't know if it's suitable now but I'll make a pull request

Thanks!

strings/search/parenthesis_test.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func TestParenthesis(t *testing.T) {
+	errors := int(0)
+	testok := true
+
+	for _, tc := range parenthesisTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := Parenthesis(tc.text)
+			if !reflect.DeepEqual(actual, tc.expected) {
```
raklaptudirm, line 53: Since `Parenthesis` is returning a `bool`, `reflect.DeepEqual` is not needed here. You can just use the not-equal to `!=` operator.

strings/search/parenthesis_test.go
```diff
@@ -0,0 +1,65 @@
...
+				testok = false
+				errors++
+				t.Errorf("Expected %t value from test %s with input %s\n", tc.expected, tc.name, tc.text)
+			}
+		})
+	}
+	if testok {
```
raklaptudirm, line 60: This notice is not required. Go will automatically report if there are any errors.

strings/search/parenthesis_test.go
```diff
@@ -0,0 +1,65 @@
...
+		fmt.Println("Pharentesis test is done without errors")
+		return
+	}
+	fmt.Printf("Pharentesis test finished with %d errors\n", errors)
```
raklaptudirm, line 64: Same as above.

strings/search/parenthesis.go
```diff
@@ -0,0 +1,21 @@
+package search
+
+func Parenthesis(text string) bool {
+	parcounter := int(0)
+
+	for i := 0; i < len(text); i++ {
```
raklaptudirm, line 6: Use a range expression here to simplify the code.

strings/search/parenthesis.go
```diff
@@ -0,0 +1,21 @@
...
+		case '(':
+			parcounter++
+		case ')':
+			parcounter--
+		}
+		if parcounter < 0 {
+			return false
+		}
+	}
+	if parcounter == 0 {
```
raklaptudirm, line 17: This if statement can be reduced to a single return.

raklaptudirm (changes requested):

89453728:
I have completed the mentioned fixes

Thanks for reviewing my proposal!

raklaptudirm (changes requested):
Left some comments, otherwise lgtm.

tjgurwara99:
Haven't looked at it properly yet - will give it a review tonight - but at first glance this algorithm is not a string searching algorithm so it doesn't belong in that directory.

raklaptudirm:
I would suggest creating a new package under `strings/`.

89453728:
at the time I doubted if it should be in search or in strings/search , seeing what there was in search focused mainly on search in lists, I decided to put it in strings/search since it is a search in a string

raklaptudirm:
@89453728 As I mentioned, it will be best to create a new package under `strings/`, something like `parenthesis`.

raklaptudirm (approved):

siriak (approved):
LGTM, thanks!

PR #494, opened 2022-04-16
add coinchange implementation

nursyah21:


dynamic/coinchange.go
```diff
@@ -0,0 +1,29 @@
+package dynamic
+/*
```
raklaptudirm, line 2: Looks like you documented the code using jsdoc. Go uses a much simpler way to document code. https://go.dev/doc/effective_go#commentary

dynamic/coinchange.go
```diff
@@ -0,0 +1,29 @@
...
+@params 
+coins: combination of coin 
+amount: target coins
+
+@return
+num of combination
+ */
+
+
+func coinChange(coins []int32, amount int32) int32 {
```
raklaptudirm, line 12: Export this function:
```suggestion
func CoinChange(coins []int32, amount int32) int32 {
```

dynamic/coinchange_test.go
```diff
@@ -0,0 +1,30 @@
+package dynamic
```
raklaptudirm, line 1: Use black-box testing:
```suggestion
package dynamic_test
```

raklaptudirm (changes requested):
Format your code with `gofmt`

dynamic/coinchange_test.go
nursyah21: if i use dynamic_test, the output is
./coinchange_test.go:23:14: undefined: CoinChange
FAIL	github.com/TheAlgorithms/Go/dynamic [build failed]

how do i run code, if i use black-box testing?

raklaptudirm: You first need to change the function name to export it.

PR #495, opened 2022-04-16
Added coinchange implementation

nursyah21:


dynamic/coinchange.go
```diff
@@ -0,0 +1,21 @@
+package dynamic
+
+// return combination of coin
```
raklaptudirm, line 3: Make this comment to be more thorough, and start with the function name.
```go
// CoinChange ...
// ...
```

dynamic/coinchange.go
```diff
@@ -0,0 +1,21 @@
...
+func CoinChange(coins []int32, amount int32) int32 {
+	combination := make([]int32, amount)
+	combination[0] = 1
+
+	//Time Complexity O(mn)
```
raklaptudirm, line 8: These comments are unnecessary.

raklaptudirm (approved):
Looks good to me.

dynamic/coinchange.go
nursyah21: can I change the code directly here, without having to open a pull request again?.

raklaptudirm: Yeah of course.

dynamic/coinchange.go
```diff
@@ -1,15 +1,12 @@
 package dynamic
 
-// return combination of coin
+// CoinChange
```
raklaptudirm, line 3: I think you are not getting the idea of go doc comments. A few examples may help:
https://github.com/TheAlgorithms/Go/blob/master/math/prime/millerrabinprimalitytest.go#L99
https://github.com/TheAlgorithms/Go/blob/master/math/prime/millerrabinprimalitytest.go#L119

nursyah21: Can you give me suggestion, what should i write. i still can't understand

is this correct?
```
// CoinChange find number of combination coins to get amount target
```

raklaptudirm: I think the language can be improved upon.

nursyah21: i only can think to write like this, make a simple comment, i look another example they make comment like this too just simple and describe what does the function do
```
// CoinChange find number of combination coins to get amount target
```

raklaptudirm:
```go
// CoinChange finds the number of possible combinations of coins
// of different values which can get to the target amount.
```

raklaptudirm (changes requested):
Please remove the changes made to the `go.mod` and the `go.sum` files.

nursyah21:
> Please remove the changes made to the `go.mod` and the `go.sum` files.
how do to remove changes commmit

raklaptudirm:
Just edit those files.

raklaptudirm:
Why did you close the pr?

PR #497, opened 2022-04-19
Fix: segmenttree algorithm bug

hugokung:
Lazy node value should be accumulated during propagation, because an interval can be updated multiple times.

### DIFF
--- a/structure/segmenttree/segmenttree.go
+++ b/structure/segmenttree/segmenttree.go
@@ -11,7 +11,7 @@ import (
 	"github.com/TheAlgorithms/Go/math/min"
 )
 
-const emptyLazyNode = -1
+const emptyLazyNode = 0
 
 //SegmentTree with original Array and the Segment Tree Array
 type SegmentTree struct {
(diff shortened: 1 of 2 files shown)


structure/segmenttree/segmenttree.go
```diff
@@ -81,7 +82,8 @@ func (s *SegmentTree) Update(node int, leftNode int, rightNode int, firstIndex i
 
 	if (leftNode >= firstIndex) && (rightNode <= lastIndex) {
 		//inside the interval
-		s.LazyTree[node] = value
+		//acumulate the lazy node value
```
siriak, line 85:
```suggestion
		//accumulate the lazy node value
```

siriak (approved):
LGTM, the linter issues seem to be unrelated to your code, so I'm merging

raklaptudirm:
@siriak Seems like the checks started failing after this merge.

raklaptudirm:
Yup, the checks failed on the test file.

PR #501, opened 2022-04-24
segmenttree: format test file with `gofmt`

raklaptudirm:
Linter tests are failing as the `segmenttree_test.go` (#497) file was not properly formatted. Fixed.


siriak (approved):
Thanks for finding that, I thought that it had something to do with a newer version of some package because it's what was in the logs!

tjgurwara99 (approved):

PR #504, opened 2022-04-29
Consolidate Primality Algorithm testing into single suite

raklaptudirm:
Tests for Primality testing algorithms were essentially duplicates of each other. Consolidated them into a single helper function which can be called with any primality testing algorithm, much like the `sort` package.


raklaptudirm:
@tjgurwara99 Tests for `MillerRabidProbabilistic` are failing.

tjgurwara99:
That is not a suitable test for a probabilistic result... The probabilistic implementation is probabilistic because of the chance of giving false positives. So the result would be be correct ("almost always" - its a mathematical term, might actually be always in fact) if its prime, but if the number is not a prime, there is no guarantee that it would not think that it is prime. Which is when it fails in the tests I think - haven't checked it yet, I'm confident that it would only fail on the non-primes.

raklaptudirm:
But it is failing on all the primes, and I don't understand why.
```
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(5) (0.00s)
        prime_test.go:27: 5: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(7) (0.00s)
        prime_test.go:27: 7: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(11) (0.00s)
        prime_test.go:27: 11: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(13) (0.00s)
        prime_test.go:27: 13: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(17) (0.00s)
        prime_test.go:27: 17: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(19) (0.00s)
        prime_test.go:27: 19: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(23) (0.00s)
        prime_test.go:27: 23: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(29) (0.00s)
        prime_test.go:27: 29: Miller-Rabin Probabilistic function returned false
    --- FAIL: TestMillerRabinProbabilistic/Miller-Rabin_Probabilistic(31) (0.00s)
        prime_test.go:27: 31: Miller-Rabin Probabilistic function returned false
```
It is not supposed to, but somehow it is failing.

tjgurwara99:
> It is not supposed to, but somehow it is failing.

Yep, that is definitely not supposed to happen 😓. I'll take a look later as soon as I have access to a laptop - though might take a few days, away on holiday. Will keep you posted!

tjgurwara99:
This might be an issue `rand.Int63n(num-1) + 2`. Could you maybe check, if you change that to something like `rand.Int63n(num-3) + 2` it might work again... Not sure why I did it that way 😅 but maybe that fixes it.

raklaptudirm:
Yes, that fixed it.

math/prime/prime_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+var primeList = []int{2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127}
+var testLimit = 127
+
+type primalityTest func(int64) bool
+
+func primalityTestTestingHelper(t *testing.T, name string, f primalityTest) {
+	arrayIndex := 0
+	for i := 1; i <= testLimit; i++ {
+		isPrime := i == primeList[arrayIndex]
```
tjgurwara99, line 16: The non primes still might be an issue in the probabilistic test (unlikely but possible) - so it might be best to keep the prime vs non-prime numbers separate... Maybe write a separate test for the non-primes only for the deterministic version and the one that is being shared should only be testing the algorithm with prime numbers.

tjgurwara99 (changes requested):
Just that one note... Let me know if you disagree though! I will approve.

math/prime/prime_test.go
raklaptudirm: We are currently using 40 rounds of Miller-Rabin tests. The max composite which we are testing is 125 (126 is even so will be discarded by the `isTrivial` function. A maximum of ¼ numbers below the number being tested can be liars. 125/4 = 31.25. So I think we are safe. The probability of that occurring is really low, and even if that happens, a re-run of the tests must fix it.

siriak (approved):

tjgurwara99 (approved):
If you don't want to separate the tests, here's my approval!

PR #512, opened 2022-07-06
feat: Lowest Common Ancestor (LCA) algorithm.

datbeohbbh:
Hi guys! I want to add an implementation of LCA algorihm.


siriak (approved):
Looks good, thanks!

yanglbme (approved):

tjgurwara99:
I feel like the graph implementation that exists in the graph package could have been adapted instead of creating a new graph data structure.

datbeohbbh:
@tjgurwara99 Thank you for your response. You are right, the graph implementation in the graph package can be adapted. But `struct Graph` saves the `edges` using `map`, I think with a large graph up to 100,000 vertices, it is not a good choice. Instead, I use `slice` to save the edges in an adjacent list. Again, thank you for spending your time!

datbeohbbh:
@tjgurwara99  Hi! As I understand, `map` is a hash table, so to access an element with a key, it must do some checking on the presence of that element and it costs some `loadFactor`, so complexity is approx `O(loadFactor)`. With `slice`, access with an index is more efficient, and complexity is `O(1)`. Back to this algorithm context, for example if I do a `dfs()` on a large tree, I need to access the list of edges of each vertex frequently, so using `slice` will be more efficient. Please correct me if I'm wrong.

PR #514, opened 2022-07-15
fix CONTRIBUTING.md

AFMahmuda:
fix https://github.com/TheAlgorithms/Go/issues/513

PR #517, opened 2022-08-04
add patience sorting algorithm

guuzaa:
feat: add patience sorting algorithm
test: add test for patience sorting algorithm

### DIFF
--- a/sort/patiencesort.go
+++ b/sort/patiencesort.go
@@ -17,14 +17,10 @@ func Patience[T constraints.Ordered](arr []T) []T {
 
 	var piles [][]T
 
-	// Traverse the given array
 	for _, card := range arr {
-		// Iterate over all the piles
-		// and check the top of the stack of each pile
-		// is less than the current element or not.
 		left, right := 0, len(piles)
 		for left < right {
-			mid := left + (right-left)>>1
+			mid := left + (right-left)/2 
 			if piles[mid][len(piles[mid])-1] >= card {
 				right = mid
 			} else {
@@ -32,14 +28,9 @@ func Patience[T constraints.Ordered](arr []T) []T {
 			}
 		}
 
-		// If all the top of the stack of each pile is less than than the current element
 		if left == len(piles) {
-			// Create a new pile with the current element
-			// and insert the pile into all the piles
 			piles = append(piles, []T{card})
 		} else {
-			// The current element is placed on the leftmost existing pile 
-			// whose top card has a value greater than or equal to the new card's value
 			piles[left] = append(piles[left], card)
 		}
 	}
@@ -50,39 +41,25 @@ func Patience[T constraints.Ordered](arr []T) []T {
 func mergePiles[T constraints.Ordered](piles [][]T) []T {
 	var ret []T
 
-	// Find the smallest element of top of pile
-	// and remove it from the piles
-	// and store it into the final array , in every iteration,
-	// util all piles are empty
 	for len(piles) > 0 {
+		minID := 0
+		minValue := piles[minID][len(piles[minID])-1]
 
-		// Store the index and the value of the smallest element
-		// of the top of the piles
-		minId := 0
-		minV := piles[minId][len(piles[minId])-1]
-
-		// Calculate the smallest element of the top of the every pile
 		for i := 1; i < len(piles); i++ {
-
-			// if minV > the top of the current pile
-			if minV > piles[i][len(piles[i])-1] {
-				// Update minV
-				minV = piles[i][len(piles[i])-1]
-				// Update index
-				minId = i
+			if minValue <= piles[i][len(piles[i])-1] {
+				continue
 			}
+
+			minValue = piles[i][len(piles[i])-1]
+			minID = i
 		}
 
-		// Insert the smallest element of the top of the pile
-		ret = append(ret, minV)
+		ret = append(ret, minValue)
 
-		// Remove the top element from the current pile
-		piles[minId] = piles[minId][:len(piles[minId])-1]
+		piles[minID] = piles[minID][:len(piles[minID])-1]
 
-		// If current pile is empty
-		if len(piles[minId]) == 0 {
-			// Remove current pile form all piles
-			piles = append(piles[:minId], piles[minId+1:]...)
+		if len(piles[minID]) == 0 {
+			piles = append(piles[:minID], piles[minID+1:]...)
 		}
 	}
 
--- a/sort/patiencesort.go
+++ b/sort/patiencesort.go
@@ -1,0 +1,90 @@
+// Package sort
+// Patience sorting is a sorting algorithm inspired by the card game patience.
+//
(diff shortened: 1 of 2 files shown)


sort/patiencesort.go
```diff
@@ -0,0 +1,90 @@
...
+	var piles [][]T
+
+	// Traverse the given array
+	for _, card := range arr {
+		// Iterate over all the piles
+		// and check the top of the stack of each pile
+		// is less than the current element or not.
+		left, right := 0, len(piles)
+		for left < right {
+			mid := left + (right-left)>>1
```
tjgurwara99, line 27:
```suggestion
			mid := (left + (right-left))>>1
```

might make it more clear what you're trying to do.

But I also suggest to keep this simple, no need for fancy bit shifting when we already know that the compiler is quick enough in division by 2. So
```suggestion
			mid := (left + (right-left))/2
```
might be more suitable.

sort/patiencesort.go
```diff
@@ -0,0 +1,90 @@
...
+		// Calculate the smallest element of the top of the every pile
+		for i := 1; i < len(piles); i++ {
+
+			// if minV > the top of the current pile
+			if minV > piles[i][len(piles[i])-1] {
+				// Update minV
+				minV = piles[i][len(piles[i])-1]
+				// Update index
+				minId = i
+			}
```
tjgurwara99, line 73: Flip this into a guard statement maybe...

```suggestion
			if minV <= piles[i][len(piles[i])-1] {
				continue
			}
			// Update minV
			minV = piles[i][len(piles[i])-1]
			// Update index
			minId = i
```

sort/patiencesort.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Find the smallest element of top of pile
+	// and remove it from the piles
+	// and store it into the final array , in every iteration,
+	// util all piles are empty
+	for len(piles) > 0 {
+
+		// Store the index and the value of the smallest element
+		// of the top of the piles
+		minId := 0
```
tjgurwara99, line 61: whenever you're using `Id` in Go, you should use it with `ID` instead of capitalising just `I`, so in this case
```suggestion
		minID := 0
```

sort/patiencesort.go
```diff
@@ -0,0 +1,90 @@
...
+	return mergePiles(piles)
+}
+
+func mergePiles[T constraints.Ordered](piles [][]T) []T {
+	var ret []T
+
+	// Find the smallest element of top of pile
+	// and remove it from the piles
+	// and store it into the final array , in every iteration,
+	// util all piles are empty
```
tjgurwara99, line 56: Not necessary to explain the code. Your code is the explanation of what is happening, comment should be used only when you have to explain why you're doing something that might not be quite straight forward.

Comments should explain "why" not "what" (IMO)

sort/patiencesort.go
```diff
@@ -0,0 +1,90 @@
...
+		// of the top of the piles
+		minId := 0
+		minV := piles[minId][len(piles[minId])-1]
+
+		// Calculate the smallest element of the top of the every pile
+		for i := 1; i < len(piles); i++ {
+
+			// if minV > the top of the current pile
+			if minV > piles[i][len(piles[i])-1] {
+				// Update minV
```
tjgurwara99, line 69: Again, no need for these kinds of comments.

sort/patiencesort.go
```diff
@@ -0,0 +1,90 @@
...
+import "github.com/TheAlgorithms/Go/constraints"
+
+func Patience[T constraints.Ordered](arr []T) []T {
+	if len(arr) <= 1 {
+		return arr
+	}
+
+	var piles [][]T
+
+	// Traverse the given array
```
tjgurwara99, line 20: This comment doesn't provide value to anyone reading the code.

tjgurwara99 (changes requested):

siriak (dismissed):
The code looks good, but a formatting check is failing

tjgurwara99:
Please run `gofmt` on your code

guuzaa:
Sorry, I ran `gofmt` on my code. I don't know why this error happened.
I will double check it as soon as possible.

guuzaa:
It is a little weird. These formatting errors occur on the files I didn't change.

guuzaa:
Hi @tjgurwara99 , I have run `golangci-lint` on my code. There are indeed some formatting errors. Then, I try to run `gofmt -w` on these files. However, there is no change. Due to these files are not written by me, I cannot change it manually. As a reult, I'm writing this comment to ask for your help.

The following is my machine :
OS `Ubuntu 20.04.3 LTS`
Go `1.18.3`
golangcli-lint  `1.48.0 built from 2d8fea81 on 2022-08-04T18:44:38Z`

guuzaa:
I checked these files again. There isn't a leading space in some comments. Does it matter?

guuzaa:
I might find the reason.
The[ last PR](https://github.com/TheAlgorithms/Go/pull/514) tested  with `golangci-lint` 1.47.
image
Recently `golangci-lint`  was upgraded to 1.48, and the old code no longer pass the formatting check.
image

tjgurwara99:
> I checked these files again. There isn't a leading space in some comments. Does it matter?

If it did, it wouldn't have passed earlier either. I also tried to look into it and couldn't wrap my head around it. Thanks for pointing out the version difference, I was going to inspect that too. I will take a look at `golangci-lint` change logs  and see what they changed. Might be related to the new Go 1.19 and the new doc styling.

In the meantime, if you'd like to merge this PR for now, I don't see any harm there since the issue is not related to your changes. cc @siriak @raklaptudirm what do you think? If you don't mind waiting however, I would prefer to get this sorted and can add changes to your PR directly.

raklaptudirm:
If the required changes are unrelated to this pr, we should merge this and make those changes ourselves imo.

siriak:
Agree with @raklaptudirm, but fixing them here also works if done quickly

tjgurwara99:
Ok, I looked over it and this is a bigger issue on the `golangci-lint` side. It seems that the issue is essentially related to the `//` and `// ` and other space formatting issues. There is a newly opened [issue](https://github.com/golangci/golangci-lint/issues/3069) on `golangci-lint` repo, which is unlikely to be quickly fixed so we can merge this PR now I guess.

tjgurwara99:
#518 fixes the issues cc @siriak @raklaptudirm @yanglbme - try to merge that before merging this 😄 

Also, the issue was indeed related to the spacing. The changes are mostly adding space or removing tabs or both in some cases. It seems that the golangci-lint 1.48 made it a bit stricter, which is quite good but not backwards compatible since this is not enforced at the `gofmt` level.

raklaptudirm (approved):

guuzaa:
I think I should merge the master first and push the merged code to this PR.

tjgurwara99 (approved):

PR #520, opened 2022-08-06
Simple showcase of Roundoff operation in GoLang

jo3zeph:
Hi I am new to GoLang, hope my work is up to your standards. Thank you for your time

feat: added roundoff.go
test: added roundoff_test.go
docs: added comments on both roundoff.go and roundoff_test.go


raklaptudirm:
I don't think this qualifies as an algorithm as this function just calls a standard library function.

PR #521, opened 2022-08-06
Added mean algorithm and test code

Jason07311:
Hello I would like to contribute to an algorithm that computes mean related in Math (Statistic). Thank you very much!
feat: added mean.go
test: added mean_test.go with 4 test cases
docs: added comment and explanation to mean.go & mean_test.go


math/mean/mean.go
```diff
@@ -0,0 +1,27 @@
+package mean
```
tjgurwara99, line 1: You do not need to create a separate package just for one function. This clearly belongs in the `math` package.

math/mean/mean.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+)
+
+func mean(values []float64) float64{
+
+	if len(values) <= 0 {
+		fmt.Println("No values detected")
```
tjgurwara99, line 10: No need for this.

math/mean/mean.go
```diff
@@ -0,0 +1,27 @@
+package mean
+
+import (
+	"fmt"
+)
+
+func mean(values []float64) float64{
```
tjgurwara99, line 7: Please capitalise this function, otherwise how would you expect it to work outside this package.

math/mean/mean_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{
+			name: "With negative values",
+			testValues: []float64{-1,2,-3,4,5},
+			average: 1.4,
+		},
+		{
+			name: "With floating values",
+			testValues: []float64{1.1,2.2,3.3,4.4,5.5},
+			average: 3.3,
+		},
```
tjgurwara99, line 32: A another test case with no values inside.

```
testValues: []float64{},
```

to make sure you cover this edge case.

tjgurwara99, line 32:
```suggestion
		},
		{
 			name: "With no values",
 			testValues: []float64{},
 			average: 0,
 		},
```

math/mean/mean.go
tjgurwara99, line 7: Please also consider using generics here, this is clearly something that would benefit from generics.

tjgurwara99 (changes requested):
Please consider the suggestions provided.

math/mean/mean.go
Jason07311: Sorry to ask, is the meaning of "No need" referring to removing the whole validating condition of empty array ? Or just the line of code printing the message ?

One reason of not suggesting to remove the validating condition above is to consider the case where  empty array is passed into the function.

math/mean/mean.go
Jason07311: Just added the generics. Please review.

math/mean/mean_test.go
Jason07311: Just added the new test case. Please review.

math/mean/mean.go
Jason07311: Just edited the function name. Please review.

math/mean/mean.go
Jason07311: Just moved the mean.go and the mean_test.go under the directory of maths

math/mean/mean.go
tjgurwara99: Just the line printing the message.

math/mean/mean.go
tjgurwara99: I can still see the file present. You need to remove this file and commit the removal to git and push your changes. Otherwise it will remain in the source control.

math/mean.go
```diff
@@ -0,0 +1,31 @@
...
+	"golang.org/x/exp/constraints"
+)
+
+type Digit interface {
+	constraints.Float | constraints.Integer
+}
+
+func Mean[T Digit](values []T) float64 {
+
+	// Validating the empty array
```
tjgurwara99, line 13: No need for these kind of comments. They usually don't provide value.

When adding comments to code, there are usually two things that you should consider:

1. are you doing something that is very complex that will not be understood by someone reading your code.
2. are you explain why you're doing something rather than what you're doing.

If none of the above cases apply then it is (usually) just considered noise.

math/mean/mean.go
Jason07311: > Just the line printing the message.

Got it, just removed it in my latest committed work

math/mean.go
```diff
@@ -0,0 +1,31 @@
...
+	if len(values) == 0 {
+		return 0
+	}
+
+	var summation float64 = 0
+
+	// Summation of all values in the slice
```
tjgurwara99, line 20: Same here.

math/mean.go
```diff
@@ -0,0 +1,31 @@
...
+	for _, singleValue := range values {
+		summation += float64(singleValue)
+	}
+
+	// Casting to float64 done for the len()
```
tjgurwara99, line 25: same here.

math/mean.go
```diff
@@ -0,0 +1,31 @@
...
+	var average = summation / float64(len(values))
+
+	// Return of average value
```
tjgurwara99, line 28: Same here.

math/mean/mean_test.go
```diff
@@ -0,0 +1,47 @@
+package mean
```
tjgurwara99, line 1: You need to remove this file too.

tjgurwara99 (changes requested):
Please consider the suggestion below/

tjgurwara99:
Also, you still haven't run `gofmt` on your code. It will keep failing our automated tests until you do.

math/mean.go
```diff
@@ -0,0 +1,31 @@
+package maths
```
tjgurwara99, line 1:
```suggestion
package math
```

math/mean.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"golang.org/x/exp/constraints"
+)
+
+type Digit interface {
```
tjgurwara99, line 7: We already have those [constraints](https://github.com/TheAlgorithms/Go/blob/master/constraints/contraints.go#L28-L30)

math/mean.go
```diff
@@ -0,0 +1,31 @@
...
+
+	var summation float64 = 0
+
+	// Summation of all values in the slice
+	for _, singleValue := range values {
+		summation += float64(singleValue)
+	}
+
+	// Casting to float64 done for the len()
+	var average = summation / float64(len(values))
```
tjgurwara99, line 26:
```suggestion
	return summation / float64(len(values))
```

tjgurwara99: With this, everything after this is redundant.

tjgurwara99 (changes requested):
Please fix these issues.

math/mean.go
```diff
@@ -0,0 +1,25 @@
+package math
+
+import (
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+type Digit interface {
```
tjgurwara99, line 7: This is useless if you're using the `constraints` package.

We already have `constraints.Number` which is exactly the same as this. Please use that instead and remote this.

math/mean_test.go
```diff
@@ -0,0 +1,51 @@
+package math
```
tjgurwara99, line 1:
```suggestion
package math_test
```

tjgurwara99 (changes requested):

math/mean.go
Jason07311: Ya, I forgot to remove them in my last push. Sorry for that.

math/mean_test.go
Jason07311: Does this indicate of creating another directory for math_test ?

Jason07311: > Does this indicate of creating another directory for math_test ?

Or just moving the whole test script to the math_test.go under math package?

Jason07311: Done updating the package.

tjgurwara99 (approved):

raklaptudirm (approved):

Jason07311:
Thanks for the effort upon reviewing my contribution and providing the suggestions, really appreciate that.

PR #522, opened 2022-08-06
Added median.go

jo3zeph:
Hi, so sorry for the previous pull request, here is a more appropriate mathematical algorithm on Median. Thank you

feat: added median.go
test: added median.test.go
docs: added comments on both median.go and median_test.go


math/median/median.go
```diff
@@ -0,0 +1,31 @@
+// author(s) [jo3zeph](https://github.com/jo3zeph)
+// description: Find the median from a set of values
+// see median_test.go
+
+package median
```
tjgurwara99, line 5: This belongs in the math package not a new package.

math/median/median.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"sort"
+)
+
+
+func median(values []float64) float64 {
```
tjgurwara99, line 12: Please consider capitalising this function and using generics to make it more usable outside the context of `float64`.

If you don't capitalise the function then the automatic documentation tool that we use will not be able to pick that up.

math/median/median.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Sorting the values in order
+	sort.Float64s(values)
+
+	var median float64
+
+	// Find the length of values
+	l := len(values)
+
+	if l == 0 {
```
tjgurwara99, line 22: Please run `gofmt` on your code, its not formatted correctly.

math/median/median_test.go
```diff
@@ -0,0 +1,57 @@
...
+			name: "Test 2",
+			testValues: []float64{21,10,22,33,11,88},
+			ans: 21.5,
+		},
+
+		{
+			name: "Test 3",
+			testValues: []float64{11.2,32.5,2.5,37.8,21.8,5.2},
+			ans: 16.5,
+		},
```
tjgurwara99, line 35: Add an edge case with no values.

math/median/median.go
```diff
@@ -0,0 +1,31 @@
...
+        return 0
+    } else if l%2 == 0 {
+        median = (values[l/2-1] + values[l/2]) / 2
+    } else {
+        median = values[l/2]
+    }
```
tjgurwara99, line 28: Also, it might be better to use switch statement in this context.

```suggestion
	switch {
	case l == 0:
		return 0
	case l % 2 == 0:
		return (values[l/2-1] + values[l/2]) / 2
	default:
		return values[l/2]
	}
```

math/median/median_test.go
tjgurwara99:
```
testValues: []float64{},
```

tjgurwara99 (changes requested):
Please consider the suggestion I have provided. Also, please run gofmt on your code. The automatic checks would fail.

math/median.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/slices"
+)
+
+type Data interface {
+	constraints.Float | constraints.Integer
+}
+
+func Median[X Data](values []X) float64 {
```
tjgurwara99, line 16:
```suggestion
func Median[T Data](values []T) float64 {
```

math/median.go
```diff
@@ -0,0 +1,37 @@
...
+package math
+
+import (
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/slices"
+)
+
+type Data interface {
+	constraints.Float | constraints.Integer
+}
```
tjgurwara99, line 14: We already have the `constraints` [here](https://github.com/TheAlgorithms/Go/blob/master/constraints/contraints.go) and we also have all the sorting algorithms, you can find the usage [here](https://github.com/TheAlgorithms/Go/blob/master/sort/sorts_test.go)

Please refrain from using the `/x/` packages since they are subject to changes meaning they can break our repo builds even when we change the go version.

tjgurwara99 (changes requested):
Please consider the suggestion below.

math/median.go
jo3zeph: I see, I have already amended the package imports

tjgurwara99: I can still see those imports, you probably haven't committed your changes or pushed your changes.

tjgurwara99: Also, we have the [`constraints.Number`](https://github.com/TheAlgorithms/Go/blob/16c92748a3db40d1c6f916e390342d02e21c22d0/constraints/contraints.go#L28-L30) present in our constraints package. You should use that instead of creating a new `Data` constraint, which is exactly the same.

jo3zeph: My bad, i have already removed it

math/median_test.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func TestMedian(t *testing.T) {
+	testCases := []struct {
+		name       string
+		testValues []float64
+		ans        float64
+	}{
+
+		//test cases
```
tjgurwara99, line 18: This comment would create issues

```suggestion
		// test cases
```

math/median_test.go
```diff
@@ -0,0 +1,59 @@
+// author(s) [jo3zeph](https://github.com/jo3zeph)
+// median_test.go
+// see median.go
+
+package math
```
tjgurwara99, line 5:
```suggestion
package math_test
```

tjgurwara99 (changes requested):
Consider the suggested changes.

Please also run `go mod tidy`

math/median_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"testing"
+)
+
+func TestMedian(t *testing.T) {
+	testCases := []struct {
+		name       string
+		testValues []float64
+		ans        float64
```
tjgurwara99, line 15:
```suggestion
		answer        float64
```

math/median.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func Median[T constraints.Number](values []T) float64 {
+	// Sorting the values in order
+
+	sort.Bubble(values)
+
+	var median float64
+
+	// Find the length of values
```
tjgurwara99, line 19:
```suggestion
```

math/median.go
```diff
@@ -0,0 +1,34 @@
...
+
+package math
+
+import (
+	"github.com/TheAlgorithms/Go/constraints"
+	"github.com/TheAlgorithms/Go/sort"
+)
+
+func Median[T constraints.Number](values []T) float64 {
+	// Sorting the values in order
```
tjgurwara99, line 13:
```suggestion
```

math/median_test.go
tjgurwara99: But I actually remove removing this line completely, its clear what it is.

```suggestion
```

math/median_test.go
```diff
@@ -0,0 +1,59 @@
...
+			testValues: []float64{11.2, 32.5, 2.5, 37.8, 21.8, 5.2},
+			ans:        16.5,
+		},
+
+		{
+			testValues: []float64{},
+		},
+	}
+
+	//for loop to perform all test cases
```
tjgurwara99, line 42:
```suggestion
```

math/median_test.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+
+			//pass test.testValues as parameter to median function in median.go
```
tjgurwara99, line 46:
```suggestion
```

math/median_test.go
```diff
@@ -0,0 +1,59 @@
...
+			returnedMedian := Median(test.testValues)
+
+			//log activity in console when error occurs
```
tjgurwara99, line 49:
```suggestion
```

math/median_test.go
```diff
@@ -0,0 +1,59 @@
...
+			t.Log(test.testValues, " ", returnedMedian)
+
+			//catch error for value mismatch between test.ans and returnMedian
```
tjgurwara99, line 52:
```suggestion
```

go.sum
```diff
@@ -0,0 +1,2 @@
+golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e h1:+WEEuIdZHnUeJJmEUjyYC2gfUMj69yZXw17EnHg/otA=
```
tjgurwara99, line 1: This file should be empty since we're not using any external package. Please run `go mod tidy` to fix this.

go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/TheAlgorithms/Go
 
 go 1.19
+
+require golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
```
tjgurwara99, line 5: `go mod tidy`

tjgurwara99 (changes requested):

math/median_test.go
jo3zeph: There isn't a package called math_test correct? May i know the reason behind this change?

jo3zeph: This suggestion has caused another error in my `median_test.go`, may I know how to fix it?
image_2022-08-07_121334046

math/median.go
```diff
@@ -0,0 +1,32 @@
...
+	var median float64
+
+	l := len(values)
+
+	switch {
+	case l == 0:
+		return 0
+
+	case l%2 == 0:
+		median = float64((values[l/2-1] + values[l/2]) / 2)
```
tjgurwara99, line 25: why are you not retuning directly?

```suggestion
		return float64((values[l/2-1] + values[l/2]) / 2)
```

math/median.go
```diff
@@ -0,0 +1,32 @@
...
+
+	default:
+		median = float64(values[l/2])
```
tjgurwara99, line 28: Same here

```suggestion
		return float64(values[l/2])
```

math/median.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	return median
```
tjgurwara99, line 31: This would be redundant after those returns and you can remove the declaration of median on line 16 as well.

tjgurwara99 (changes requested):

math/median.go
jo3zeph: I prefer it this way because it is easier for me to comprehend. But sure, I'll make the changes

tjgurwara99 (approved):

raklaptudirm (approved):

jo3zeph:
Thank you for your patience to guide me and the opportunity to contribute :)

PR #523, opened 2022-08-06
Added mode.go & mode_test.go

CalvinNJK:
Hello, I would like to contribute an algorithm called Mode that related in Math(Statictis). Thank you very much!

feat: add mode.go
test: add mode_test.go
docs:  add comments and explanation to  mode.go & mode_test.go


math/mode/mode.go
```diff
@@ -0,0 +1,40 @@
+// mode_test.go
+// author(s): [CalvinNJK] (https://github.com/CalvinNJK)
+// description: Finding Mode Value In an Array
+// see mode.go
+
+package mode
```
tjgurwara99, line 6: This algorithm belongs in the `math` package not its own sub package.

Please move this to the `math` package.

math/mode/mode.go
```diff
@@ -0,0 +1,40 @@
...
+	// declare variables
+	maxCount := 0
+	maxValue := 0.0
+
+	// Loop for the whole array
+	for i := 0; i < n; i++ {
+
+		count := 0
+
+		// check the selected Value have any repeated same value
```
tjgurwara99, line 22: These comments are of no use, everyone who can read the code can see what you're trying to accomplish. When commenting on code, describe "why" instead of "what".

math/mode/mode_test.go
```diff
@@ -0,0 +1,59 @@
...
+		{
+			name:    "An array of negative real numbers",
+			numbers: []float64{-98.1, -26.8, -54.45, -26.8, -1.5, -26.8, -33, -19.5, -26.8},
+			mode:    -26.8,
+		},
+		{
+			name:    "An array of positve and negative real numbers",
+			numbers: []float64{-17, 28.9, -5.2, -19.5, 77.3, -5.2, 39.3, 28.5, -59.77, -5.2},
+			mode:    -5.2,
+		},
```
tjgurwara99, line 46: Add the zero value edge case

```
numbers: []float64{}
```

tjgurwara99 (changes requested):
Please consider the suggestion provided and please use generics, since this algorithm can clearly benefit from a generic implementation.

math/mode/mode.go
```diff
@@ -0,0 +1,40 @@
+// mode_test.go
```
tjgurwara99, line 1: The file name is wrong.

tjgurwara99:
It seems there is also a spelling error present in your code/comments

math/mode.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"errors"
+
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+type Number interface {
+	constraints.Float | constraints.Integer
+}
```
tjgurwara99, line 18: We already have [`constraints.Number`](https://github.com/TheAlgorithms/Go/blob/master/constraints/contraints.go#L28-L30)

math/mode.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Error when array is empty
+var emptyModeArr = errors.New("Array is empty")
+
+func Mode[T Number](numbers []T) (T, error) {
```
tjgurwara99, line 23:
```suggestion
func Mode[T constraints.Number](numbers []T) (T, error) {
```

math/mode.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// Assign when there is a Mode Value
+	maxCount := 0
+	var maxValue T = 0
+
+	if n == 0 {
+		return 0, emptyModeArr
+	} else {
+
+		// Check for each value has any repeated value
```
tjgurwara99, line 36: Please remove all comment explaining the code.

The code should be self explanatory, which it is in this case. Use comments only when you're explaining **"why"** you're doing something not **"what"** you're doing.

math/mode.go
```diff
@@ -0,0 +1,61 @@
+// mode.go
+// author(s): [CalvinNJK] (https://github.com/CalvinNJK)
+// description: Finding Mode Value In an Array
+// see mode.go
+
+package math
+
+//package constraints
```
tjgurwara99, line 8: Why do you have this comment in this file?

math/mode_test.go
```diff
@@ -0,0 +1,114 @@
...
+func TestModeInteger(t *testing.T) {
+
+	testCases := []struct {
+		name    string
+		numbers []int
+		mode    int
+		err     error
+	}{
+		// Test Cases
+		// For INTEGER values
```
tjgurwara99, line 21: This is a good comment. Can be in a single line though.

```suggestion
		// Test cases for `int` values
```

math/mode_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		{
+			name:    "If array has no value",
+			numbers: []int{},
+			mode:    0,
+			err:     emptyModeArr,
+		},
+	}
+
+	// Run for all the test cases
```
tjgurwara99, line 48: Not a necessary comment again.

math/mode_test.go
```diff
@@ -0,0 +1,114 @@
...
+				t.Errorf("\n Failed test %s,\n Numbers: %v,\n Correct Mode: %v,\n Returned Mode: %v\n",
+					test.name, test.numbers, test.mode, returnMode)
+			}
+			if !errors.Is(err, test.err) {
+				t.Errorf("\n Failed test %s,\n Numbers: %v,\n Correct Error: %v,\n Returned Error: %v\n",
+					test.name, test.numbers, test.err, err)
+			}
+		})
+	}
+}
```
tjgurwara99, line 62: You should combine all the tests into one test. Separate the test cases with the comment like I suggested.

change the struct like this:

```suggestion
func TestMode(t *testing.T) {

	testCases := []struct {
		name    string
		numbers []constraints.Number
		mode    int
		err     error
	}{
		// Test cases for int values
		{
			name:    "An array of positive whole numbers",
			numbers: []int{10, 52, 10, 92, 10, 75, 60, 10, 44, 29},
			mode:    10,
			err:     nil,
		},
		{
			name:    "An array of negative whole numbers",
			numbers: []int{-19, -12, -74, -19, -22, -56, -19, -19, -68, -93},
			mode:    -19,
			err:     nil,
		},
		{
			name:    "An array of positive & negative whole numbers",
			numbers: []int{18, -28, 33, -28, 2, 39, 48, -49, -87, 78, -28},
			mode:    -28,
			err:     nil,
		},
		{
			name:    "If array has no value",
			numbers: []int{},
			mode:    0,
			err:     emptyModeArr,
		},
		// Test cases for float64 values
		{
 			name:    "An array of positive real numbers",
 			numbers: []float64{1.5, 2.88, 84.4, 77.2, 29.8, 46.2, 33.7, 88.4, 88.4},
 			mode:    88.4,
 			err:     nil,
 		},
 		{
 			name:    "An array of negative real numbers",
 			numbers: []float64{-98.1, -26.8, -54.45, -26.8, -1.5, -26.8, -33, -19.5, -26.8},
 			mode:    -26.8,
 			err:     nil,
 		},
 		{
 			name:    "An array of positive and negative real numbers",
 			numbers: []float64{-17, 28.9, -5.2, -19.5, 77.3, -5.2, 39.3, 28.5, -59.77, -5.2},
 			mode:    -5.2,
 			err:     nil,
 		},
 		{
 			name:    "If array has no value",
 			numbers: []float64{},
 			mode:    0,
 			err:     emptyModeArr,
 		},
	}

	// Run for all the test cases
	for _, test := range testCases {
		t.Run(test.name, func(t *testing.T) {
			returnMode, err := Mode(test.numbers)
			if returnMode != test.mode {
				t.Errorf("\n Failed test %s,\n Numbers: %v,\n Correct Mode: %v,\n Returned Mode: %v\n",
					test.name, test.numbers, test.mode, returnMode)
			}
			if !errors.Is(err, test.err) {
				t.Errorf("\n Failed test %s,\n Numbers: %v,\n Correct Error: %v,\n Returned Error: %v\n",
					test.name, test.numbers, test.err, err)
			}
		})
	}
}
```

math/mode_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestModeFloat(t *testing.T) {
```
tjgurwara99, line 64: If you follow the above suggestion, then this is redundant and can be removed.

tjgurwara99 (changes requested):
Please consider the changes suggested.

math/mode.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+type Number interface {
+	constraints.Float | constraints.Integer
+}
+
+// Error when array is empty
+var emptyModeArr = errors.New("Array is empty")
```
tjgurwara99, line 21: This error should be constant.

```suggestion
var ErrEmptySlice = errors.New("empty slice provided")
```

PS: In Go, errors string don't start with a capital letter

math/mode_test.go
CalvinNJK: Hey, using your suggestion on Line 18 causes error on my side and I not sure how to resolve it.

The Error message is:
package constraints ("github.com/TheAlgorithms/Go/constraints")
cannot use type constraints.Number outside a type constraint: interface contains type constraintscompiler[MisplacedConstraintIface](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal?utm_source%3Dgopls#MisplacedConstraintIface)

image

tjgurwara99: The following should work
```suggestion
type testCase[T constraints.Number] struct {
	name    string
	numbers []T
	mode    T
	err     error
}

func testModeFramework[T constraints.Number](t *testing.T, testCases []testCase[T]) {
	for _, test := range testCases {
		t.Run(test.name, func(t *testing.T) {
			returnMode, err := math.Mode(test.numbers)
			if returnMode != test.mode {
				t.Errorf("\n Failed test %s,\n Numbers: %v,\n Correct Mode: %v,\n Returned Mode: %v\n",
					test.name, test.numbers, test.mode, returnMode)
			}
			if !errors.Is(err, test.err) {
				t.Errorf("\n Failed test %s,\n Numbers: %v,\n Correct Error: %v,\n Returned Error: %v\n",
					test.name, test.numbers, test.err, err)
			}
		})
	}
}

func TestMode(t *testing.T) {
	// test cases for integer values
	intTestCases := []testCase[int]{
		{
			name:    "An array of positive whole numbers",
			numbers: []int{10, 52, 10, 92, 10, 75, 60, 10, 44, 29},
			mode:    10,
			err:     nil,
		},
		{
			name:    "An array of negative whole numbers",
			numbers: []int{-19, -12, -74, -19, -22, -56, -19, -19, -68, -93},
			mode:    -19,
			err:     nil,
		},
		{
			name:    "An array of positive & negative whole numbers",
			numbers: []int{18, -28, 33, -28, 2, 39, 48, -49, -87, 78, -28},
			mode:    -28,
			err:     nil,
		},
		{
			name:    "If array has no value",
			numbers: []int{},
			mode:    0,
			err:     math.ErrEmptySlice,
		},
	}

	testModeFramework(t, intTestCases)

	// test cases for float64 values
	floatTestCases := []testCase[float64]{
		{
			name:    "An array of positive real numbers",
			numbers: []float64{1.5, 2.88, 84.4, 77.2, 29.8, 46.2, 33.7, 88.4, 88.4},
			mode:    88.4,
			err:     nil,
		},
		{
			name:    "An array of negative real numbers",
			numbers: []float64{-98.1, -26.8, -54.45, -26.8, -1.5, -26.8, -33, -19.5, -26.8},
			mode:    -26.8,
			err:     nil,
		},
		{
			name:    "An array of positive and negative real numbers",
			numbers: []float64{-17, 28.9, -5.2, -19.5, 77.3, -5.2, 39.3, 28.5, -59.77, -5.2},
			mode:    -5.2,
			err:     nil,
		},
		{
			name:    "If array has no value",
			numbers: []float64{},
			mode:    0,
			err:     math.ErrEmptySlice,
		},
	}

	testModeFramework(t, floatTestCases)
}
```

I suggested above thinking that the Go 1.19 would support struct field generics. They still haven't which is fine, maybe they might in the next release.

math/mode_test.go
```diff
@@ -0,0 +1,114 @@
+// mode.go
+// author(s): [CalvinNJK] (https://github.com/CalvinNJK)
+// description: Test for Finding Mode Value In an Array
+
+package math
```
tjgurwara99, line 5:
```suggestion
package math_test
```

tjgurwara99 (changes requested):

tjgurwara99:
Also, the code doesn't compile since you committed the change I suggested at one place without changes all the associated places.

math/mode.go
```diff
@@ -0,0 +1,48 @@
...
+
+package math
+
+import (
+	"errors"
+
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+// Error when array is empty
```
tjgurwara99, line 14: Just one last thing, everything else seems ok to me now.

```suggestion
// ErrEmptySlice is the error returned by functions in math package when
// an empty slice is provided to it as argument when the function expects
// a non-empty slice.
```

tjgurwara99 (changes requested):

math/mode.go
```diff
@@ -0,0 +1,50 @@
...
+
+			if count > maxCount {
+				maxCount = count
+				maxValue = numbers[i]
+			}
+		}
+
+		return maxValue, nil
+
+	}
```
tjgurwara99, line 49: You don't need this else statement.

```suggestion
	}

	for i := 0; i < n; i++ {
		count := 0
		for k := 0; k < n; k++ {
			if numbers[k] == numbers[i] {
				count++
			}
		}
		if count > maxCount {
			maxCount = count
			maxValue = numbers[i]
		}
	}
	return maxValue, nil
```

math/mode.go
```diff
@@ -0,0 +1,50 @@
...
+func Mode[T constraints.Number](numbers []T) (T, error) {
+
+	n := len(numbers)
+
+	maxCount := 0
+	var maxValue T = 0
+
+	if n == 0 {
+		return 0, ErrEmptySlice
+	} else {
```
tjgurwara99, line 28: This if block should go before definition of `maxCount` and `maxValue` (line 22) since everything you need for this block is already present at that point.

After that you don't need the else statement as mentioned in the other comment.

tjgurwara99 (changes requested):
Almost there

tjgurwara99 (approved):

raklaptudirm (changes requested):
The performance of the algorithm can be vastly improved using a hashmap. The implementation will also be really simple, so I think you should change that.

math/mode.go
```diff
@@ -18,32 +18,34 @@ var ErrEmptySlice = errors.New("empty slice provided")
...
+		temp := numbers[i]
+		value, check := countMap[temp]
+		if check == false {
+			countMap[temp] = 1
+		} else {
+			countMap[temp] = value + 1
 		}
+	}
+
+	var mod T
```
tjgurwara99, line 39: You can call this `mode` instead of `mod` - I believe you wrote it as `mod` because the package name was conflicting with it earlier but now it should be fine to call it `mode`

tjgurwara99 (changes requested):

math/mode.go
```diff
@@ -18,32 +18,34 @@ var ErrEmptySlice = errors.New("empty slice provided")
...
 	n := len(numbers)
 
 	if n == 0 {
 		return 0, ErrEmptySlice
 	}
 
-	maxCount := 0
-	var maxValue T = 0
-
 	for i := 0; i < n; i++ {
```
raklaptudirm, line 29: Use a `range` loop here.

math/mode.go
```diff
@@ -18,32 +18,34 @@ var ErrEmptySlice = errors.New("empty slice provided")
...
-
-		count := 0
-
-		for k := 0; k < n; k++ {
-
-			if numbers[k] == numbers[i] {
-				count++
-			}
+		temp := numbers[i]
+		value, check := countMap[temp]
```
raklaptudirm, line 31: Put this statement in the scope of the if statement.

math/mode.go
```diff
@@ -18,32 +18,34 @@ var ErrEmptySlice = errors.New("empty slice provided")
...
+		if check == false {
```
raklaptudirm, line 32: Don't check for equality with boolean constants.

raklaptudirm (changes requested):

tjgurwara99 (dismissed):

math/mode.go
```diff
@@ -0,0 +1,50 @@
...
+
+	countMap := make(map[T]int)
+
+	n := len(numbers)
+
+	if n == 0 {
+		return 0, ErrEmptySlice
+	}
+
+	for i := range numbers {
```
raklaptudirm, line 29:
```suggestion
	for _, number := range numbers {
```
You don't need to declare the `temp` variable just do it here.

math/mode.go
```diff
@@ -0,0 +1,50 @@
...
+		temp := numbers[i]
+		if value, check := countMap[temp]; check {
+			countMap[temp] = value + 1
```
raklaptudirm, line 32:
```suggestion
			countMap[temp]++
```

raklaptudirm (changes requested):

raklaptudirm (approved):

math/mode.go
tjgurwara99: Please don't mark things as resolved without a comment or without addressing the suggestion presented.

tjgurwara99 (changes requested):

tjgurwara99 (approved):

raklaptudirm (approved):

PR #524, opened 2022-08-09
Add tangent

Codesuckmehard:
To calculate the radians of tangent

### DIFF
--- a/math/tan.go
+++ b/math/tan.go
@@ -1,0 +1,9 @@
+package math
+
+import (
+	"math"
+)
+
+func Tan(x float64) float64 {
+	return Cot((math.Pi / 2) - x)
+}
--- a/math/tan_test.go
+++ b/math/tan_test.go
@@ -1,0 +1,47 @@
+package math_test
+
+import (
+	algmath "github.com/TheAlgorithms/Go/math"
+	stdmath "math"
+	"testing"
+)
+
+const epsilon = 0.001
+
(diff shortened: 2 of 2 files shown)


math/tan.go
```diff
@@ -0,0 +1,9 @@
+package math
+
+import (
+	"math"
+)
+
+func Tan(x float64) float64 {
+	return Cot((math.Pi / 2) - x)
```
tjgurwara99, line 8: I don't think we have `Cot` defined, do we?

Anyways, we have `Sin` and `Cos` defined so why not use the `Sin/Cos`?

tjgurwara99 (changes requested):

raklaptudirm:
IMO, this algorithm is too trivial.

tjgurwara99:
> IMO, this algorithm is too trivial.

Depends on what you mean by trivial. If you look at our [Cos](https://github.com/TheAlgorithms/Go/blob/feb2246c4605f9c9b4e00a491b2930524bbda2be/math/cos.go#L10) then I don't think composing these algorithms together to make new ones is trivial. If this is trivial, then our [Sin](https://github.com/TheAlgorithms/Go/blob/feb2246c4605f9c9b4e00a491b2930524bbda2be/math/sin.go#L9) is also quite trivial.

raklaptudirm (changes requested):
Format your code with <samp>gofmt</samp>.

PR #525, opened 2022-08-16
doc: Added a useful resource

Vaishnaviamirayada:
Have added a useful resource about the linked list in a data structure from the scaler topics.

### DIFF
--- a/structure/linkedlist/Readme.md
+++ b/structure/linkedlist/Readme.md
@@ -2,7 +2,7 @@
 ***
 ## What is it?
 ***
-[Linked list is a data structure](https://www.scaler.com/topics/linked-list/) that is a linear chain of data elements whose order is not given by their phyisical placement in memory. This structure is built up of nodes which point to the element ahead or behind this particular node (depending on the type of Linked List).
+[Linked list](https://www.scaler.com/topics/linked-list/) is a data structure that is a linear chain of data elements whose order is not given by their phyisical placement in memory. This structure is built up of nodes which point to the element ahead or behind this particular node (depending on the type of Linked List).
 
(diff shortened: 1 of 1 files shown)


structure/linkedlist/Readme.md
```diff
@@ -2,7 +2,7 @@
 ***
 ## What is it?
 ***
-Linked list is a data structure that is a linear chain of data elements whose order is not given by their phyisical placement in memory. This structure is built up of nodes which point to the element ahead or behind this particular node (depending on the type of Linked List).
+[Linked list is a data structure](https://www.scaler.com/topics/linked-list/) that is a linear chain of data elements whose order is not given by their phyisical placement in memory. This structure is built up of nodes which point to the element ahead or behind this particular node (depending on the type of Linked List).
```
tjgurwara99, line 5: I think just wrapping the `Linked list` into a link is better than the whole thing.

```suggestion
[Linked list](https://www.scaler.com/topics/linked-list/) is a data structure that is a linear chain of data elements whose order is not given by their phyisical placement in memory. This structure is built up of nodes which point to the element ahead or behind this particular node (depending on the type of Linked List).
```

tjgurwara99 (changes requested):
Consider the following suggestion

structure/linkedlist/Readme.md
tjgurwara99: But overall I would also recommend looking over the code to see the documentation code. There is an automated doc generator that we have in this repository that generates the documentation for all the implementations in the repository. I believe they are more representative and would also hold the most up to date information. 🙂

tjgurwara99 (approved):

siriak (approved):

PR #526, opened 2022-08-25
Add char-occurrence algorithm

mo1ein:
I added char charoccurrence algorithm.
for example `hello world` has:
```
l: 3                                                                                                                                               
o: 2                                                                                                                                               
 : 1                                                                                                                                               
w: 1                                                                                                                                               
r: 1                                                                                                                                               
d: 1                                                                                                                                               
h: 1                                                                                                                                               
e: 1 
```


strings/charoccurrence/charoccurrence.go
```diff
@@ -0,0 +1,19 @@
+// charoccurrence.go
+// description: Chars occurrence in the sentence
+// details: An algorithm to count chars in string
+// author(s) [Moein](https://github.com/mo1ein)
+// see charoccurrence_test.go
+
+package charoccurrence
```
tjgurwara99, line 7: This doesn't belong in a separate package - its a simple utility and specifically related to strings therefore it should be in the `strings` package.

strings/charoccurrence/charoccurrence.go
```diff
@@ -0,0 +1,19 @@
...
+
+func Count(text string) map[rune]int {
```
tjgurwara99, line 9: Personally I feel this is too trivial; what do you guys think? @siriak @raklaptudirm @yanglbme

strings/charoccurrence/charoccurrence.go
```diff
@@ -0,0 +1,19 @@
...
+	charOccurrence := make(map[rune]int, 0)
+	for _, c := range text {
+		if _, ok := charOccurrence[c]; ok {
+			charOccurrence[c]++
+		} else {
+			charOccurrence[c] = 1
+		}
```
tjgurwara99, line 16:
```suggestion
		if _, ok := charOccurrence[c]; !ok {
			charOccurrence[c] = 0
		}
		charOccurrence[c]++
```

strings/charoccurrence/charoccurrence_test.go
```diff
@@ -0,0 +1,40 @@
...
+		map[rune]int{32: 1, 33: 1, 1575: 2, 1583: 1, 1587: 1, 1604: 1, 1605: 1, 1606: 1, 1740: 1},
+	},
+}
+
+func TestCount(t *testing.T) {
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			funcResult := Count(test.input)
+			eq := reflect.DeepEqual(test.expected, funcResult)
+			if !eq {
```
tjgurwara99, line 35:
```suggestion
			if !reflect.DeepEqual(test.expected, funcResult) {
```

strings/charoccurrence/charoccurrence_test.go
```diff
@@ -0,0 +1,40 @@
+package charoccurrence
```
tjgurwara99, line 1: This should be tested outside the package like so
```suggestion
package charoccurrence_test
```
Although you should move this to the strings package rather than create its own package.

strings/charoccurrence/charoccurrence_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+var testCases = []struct {
+	name     string       // test description
+	input    string       // user input
+	expected map[rune]int // expected return
```
tjgurwara99, line 11:
```suggestion
	name     string
	input    string
	expected map[rune]int
```
These are quite clear, there is no need for such redundant comments.

strings/charoccurrence/charoccurrence_test.go
```diff
@@ -0,0 +1,40 @@
...
+	},
+}
+
+func TestCount(t *testing.T) {
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			funcResult := Count(test.input)
+			eq := reflect.DeepEqual(test.expected, funcResult)
+			if !eq {
+				t.Errorf("Expected answer '%v' for string '%s' but answer given was %v", test.expected, test.input, funcResult)
```
tjgurwara99, line 36:
```suggestion
				t.Errorf("expected: %v, got: %v", test.expected, funcResult)
```

tjgurwara99 (commented):
Please consider the following suggestions

tjgurwara99:
And please provide descriptions when you open PR's

strings/charoccurrence/charoccurrence.go
mo1ein: Do you mean put the files outside the `charoccurrence` dir and use `strings` package instead of `charoccurrence`?

strings/charoccurrence/charoccurrence_test.go
mo1ein: - [x] fixed

strings/charoccurrence/charoccurrence_test.go
mo1ein: - [x] fixed

strings/charoccurrence/charoccurrence_test.go
mo1ein: - [x] fixed

strings/charoccurrence/charoccurrence.go
mo1ein: - [x] fixed

strings/charoccurrence/charoccurrence.go
tjgurwara99: yes

strings/charoccurrence/charoccurrence.go
tjgurwara99: Please don't mark things as resolved when they haven't been resolved 🙂

strings/charoccurrence/charoccurrence.go
mo1ein: OK.

strings/charoccurrence.go
```diff
@@ -0,0 +1,18 @@
+// charoccurrence.go
+// description: Chars occurrence in the sentence
+// details: An algorithm to count chars in string
+// author(s) [Moein](https://github.com/mo1ein)
+// see charoccurrence_test.go
+
+package strings
+
+func Count(text string) map[rune]int {
```
tjgurwara99, line 9: Now this probably needs more context
```suggestion
func CountChars(text string) map[rune]int {
```

strings/charoccurrence_test.go
```diff
@@ -0,0 +1,39 @@
...
+		map[rune]int{32: 2, 19990: 1, 30028: 1},
+	},
+	{
+		"persian text",
+		"سلام دنیا!",
+		map[rune]int{32: 1, 33: 1, 1575: 2, 1583: 1, 1587: 1, 1604: 1, 1605: 1, 1606: 1, 1740: 1},
+	},
+}
+
+func TestCount(t *testing.T) {
```
tjgurwara99, line 30:
```suggestion
func TestCountChars(t *testing.T) {
```

strings/charoccurrence_test.go
```diff
@@ -0,0 +1,39 @@
+package charoccurrence_test
```
tjgurwara99, line 1:
```suggestion
package strings_test
```
The tests would be failing...

strings/charoccurrence_test.go
```diff
@@ -0,0 +1,39 @@
...
+		"persian text",
+		"سلام دنیا!",
+		map[rune]int{32: 1, 33: 1, 1575: 2, 1583: 1, 1587: 1, 1604: 1, 1605: 1, 1606: 1, 1740: 1},
+	},
+}
+
+func TestCount(t *testing.T) {
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			funcResult := Count(test.input)
```
tjgurwara99, line 33:
```suggestion
			funcResult := strings.CountChars(test.input)
```

strings/charoccurrence.go
```diff
@@ -0,0 +1,18 @@
...
+	charOccurrence := make(map[rune]int, 0)
```
tjgurwara99, line 10:
```suggestion
	charMap := make(map[rune]int, 0)
```

the name is too big. In Go, people usually prefer smaller names unless absolutely necessary to add more context. Here, its quite self explanatory.

strings/charoccurrence_test.go
```diff
@@ -0,0 +1,39 @@
+package charoccurrence_test
+
+import (
+	"reflect"
+	"testing"
+)
+
+var testCases = []struct {
```
tjgurwara99, line 8: This should probably be moved inside the test function - otherwise it can create conflict.

strings/charoccurrence.go
tjgurwara99, line 9: Also add documentation comment.

tjgurwara99 (changes requested):

strings/charoccurrence_test.go
mo1ein: And also must import `strings`. Yes?

strings/charoccurrence_test.go
mo1ein: - [x] fixed

strings/charoccurrence.go
mo1ein: - [x] fixed

strings/charoccurrence_test.go
mo1ein: - [x] fixed

strings/charoccurrence_test.go
mo1ein: - [x] fixed

strings/charoccurrence.go
mo1ein: - [x] fixed

mo1ein: - [x] fixed

strings/charoccurrence_test.go
tjgurwara99: Yes

mo1ein: - [x] fixed

strings/charoccurrence.go
```diff
@@ -0,0 +1,20 @@
+// charoccurrence.go
+// description: Chars occurrence in the sentence
+// details: An algorithm to count chars in string
+// author(s) [Moein](https://github.com/mo1ein)
+// see charoccurrence_test.go
+
+package strings
+
+// CountChars count chars occurrence from string
+// It returns the map of rune:int
```
tjgurwara99, line 10:
```suggestion
// CountChars counts the number of a times a character
// has occurred in the provided string argument and
// returns a map with `rune` as keys and the count as value.
```

strings/charoccurrence.go
```diff
@@ -0,0 +1,20 @@
+// charoccurrence.go
+// description: Chars occurrence in the sentence
+// details: An algorithm to count chars in string
```
tjgurwara99, line 3:
```suggestion
// details: An algorithm which counts the number
// of times a character occurred in a string.
```

tjgurwara99 (changes requested):

strings/charoccurrence.go
mo1ein: - [x] fixed

strings/charoccurrence.go
mo1ein: - [x] fixed

tjgurwara99: This is not fixed - look over what you have added.

strings/charoccurrence.go
```diff
@@ -0,0 +1,22 @@
+// charoccurrence.go
+// description: Chars occurrence in the sentence
+// details: An algorithm which counts the number
+// of times a character occurred in a string.
```
tjgurwara99, line 4:
```suggestion
// description: An algorithm which counts the number
// of times a character occurred in a string.
```

tjgurwara99 (dismissed):
One last thing but everything else LGTM.

strings/charoccurrence.go
mo1ein: Now the last thing is ok. :)

tjgurwara99: I think you don't know much about how the diffs work 🙂 Essentially, the `-` sign on the line just says to remove those lines and `+` means to add those lines. Since in this diff, we have 2 `+` and 3 `-` it means you needed to remove a line but you just added two lines.

To help, there is actually a commit suggestion button on suggested text which you can use to commit directly.

tjgurwara99 (approved):

strings/charoccurrence.go
mo1ein: I know but my place is not good for coding now. sorry :)

strings/charoccurrence/charoccurrence.go
siriak: @tjgurwara99, it's too simple as an algorithm, but can be a good entry point for somebody who only starts exploring this repo.

siriak: I think we may include it taking into account that tests, description and everything else is in place

siriak (approved):

raklaptudirm:
Merged as errors were from different files.

PR #528, opened 2022-08-28
add generics

crunchypi:
Added generic syntax for /structure/linkedlist; led to also having a need to change
some function signatures e.g
func (ll *Singly[T]) DelAtBeg() int // -1 on not-found
->
func (ll *Singly[T]) DelAtBeg() (T, bool)

### DIFF
--- a/structure/linkedlist/doubly.go
+++ b/structure/linkedlist/doubly.go
@@ -15,16 +15,16 @@ import "fmt"
 // `import "github.com/TheAlgorithms/Go/structure/linkedlist"`
 //
 // and call linkedlist.Doubly to create a new doubly linked list.
-type Doubly struct {
-	Head *Node
+type Doubly[T any] struct {
+	Head *Node[T]
 }
 
-func NewDoubly() *Doubly {
-	return &Doubly{nil}
+func NewDoubly[T any]() *Doubly[T] {
+	return &Doubly[T]{}
 }
 
 // AddAtBeg Add a node to the beginning of the linkedlist
-func (ll *Doubly) AddAtBeg(val any) {
+func (ll *Doubly[T]) AddAtBeg(val T) {
 	n := NewNode(val)
 	n.Next = ll.Head
 
@@ -37,7 +37,7 @@ func (ll *Doubly) AddAtBeg(val any) {
 }
 
 // AddAtEnd Add a node at the end of the linkedlist
-func (ll *Doubly) AddAtEnd(val any) {
+func (ll *Doubly[T]) AddAtEnd(val T) {
 	n := NewNode(val)
 
 	if ll.Head == nil {
@@ -53,9 +53,10 @@ func (ll *Doubly) AddAtEnd(val any) {
 }
 
 // DelAtBeg Delete the node at the beginning of the linkedlist
-func (ll *Doubly) DelAtBeg() any {
+func (ll *Doubly[T]) DelAtBeg() (T, bool) {
 	if ll.Head == nil {
-		return -1
+		var r T
+		return r, false
 	}
 
 	cur := ll.Head
@@ -64,14 +65,15 @@ func (ll *Doubly) DelAtBeg() any {
 	if ll.Head != nil {
 		ll.Head.Prev = nil
 	}
-	return cur.Val
+	return cur.Val, true
 }
 
 // DetAtEnd Delete a node at the end of the linkedlist
-func (ll *Doubly) DelAtEnd() any {
+func (ll *Doubly[T]) DelAtEnd() (T, bool) {
 	// no item
 	if ll.Head == nil {
-		return -1
+		var r T
+		return r, false
 	}
 
 	// only one item
@@ -86,11 +88,11 @@ func (ll *Doubly) DelAtEnd() any {
 
 	retval := cur.Next.Val
 	cur.Next = nil
-	return retval
+	return retval, true
 }
 
 // Count Number of nodes in the linkedlist
-func (ll *Doubly) Count() any {
+func (ll *Doubly[T]) Count() int {
 	var ctr int = 0
 
 	for cur := ll.Head; cur != nil; cur = cur.Next {
@@ -101,8 +103,8 @@ func (ll *Doubly) Count() any {
 }
 
 // Reverse Reverse the order of the linkedlist
-func (ll *Doubly) Reverse() {
-	var Prev, Next *Node
+func (ll *Doubly[T]) Reverse() {
+	var Prev, Next *Node[T]
 	cur := ll.Head
 
 	for cur != nil {
@@ -117,7 +119,7 @@ func (ll *Doubly) Reverse() {
 }
 
 // Display display the linked list
-func (ll *Doubly) Display() {
+func (ll *Doubly[T]) Display() {
 	for cur := ll.Head; cur != nil; cur = cur.Next {
 		fmt.Print(cur.Val, " ")
 	}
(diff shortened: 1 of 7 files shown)


siriak (approved):
Thanks! @tjgurwara99 what do you think?

tjgurwara99:
Haven't properly reviewed it - I glanced a bit and had some comments but didn't have time to completely review it. Will do that tonight 😄

structure/linkedlist/doubly.go
```diff
@@ -64,14 +64,14 @@ func (ll *Doubly) DelAtBeg() any {
...
+	return cur.Val, true
 }
 
 // DetAtEnd Delete a node at the end of the linkedlist
-func (ll *Doubly) DelAtEnd() any {
+func (ll *Doubly[T]) DelAtEnd() (T, bool) {
 	// no item
 	if ll.Head == nil {
-		return -1
+		return *new(T), false
```
tjgurwara99, line 74: This is actually quite confusing; I would prefer something like this instead
```suggestion
		var r T
		return r, false
```

For the first couple of minutes I was confused that I didn't see any `new` function then I remembered that it was the inbuilt new.

structure/linkedlist/singlylinkedlist.go
```diff
@@ -45,23 +45,25 @@ func (ll *Singly) AddAtEnd(val any) {
...
 }
 
-// DelAtBeg deletes the snode at the head(beginning) of the list and returns its value. Returns -1 if the list is empty.
-func (ll *Singly) DelAtBeg() any {
+// DelAtBeg deletes the snode at the head(beginning) of the list
+// and returns its value. Returns false if the list is empty.
+func (ll *Singly[T]) DelAtBeg() (T, bool) {
 	if ll.Head == nil {
-		return -1
+		return *new(T), false
```
tjgurwara99, line 52: Same as above

structure/linkedlist/singlylinkedlist.go
```diff
@@ -45,23 +45,25 @@ func (ll *Singly) AddAtEnd(val any) {
...
 }
 
-// DelAtEnd deletes the snode at the tail(end) of the list and returns its value. Returns -1 if the list is empty.
-func (ll *Singly) DelAtEnd() any {
+// DelAtEnd deletes the snode at the tail(end) of the list
+// and returns its value. Returns false if the list is empty.
+func (ll *Singly[T]) DelAtEnd() (T, bool) {
 	if ll.Head == nil {
-		return -1
+		return *new(T), false
```
tjgurwara99, line 66: Same as above

structure/linkedlist/doubly_test.go
```diff
@@ -63,15 +63,15 @@ func TestDoubly(t *testing.T) {
 
 	t.Run("Test DelAtBeg", func(t *testing.T) {
 		want := 3
-		got := newList.DelAtBeg()
+		got, _ := newList.DelAtBeg()
```
tjgurwara99, line 66: Since you changed the signature, I would expect this to be tested as well.

structure/linkedlist/doubly.go
```diff
@@ -64,14 +64,14 @@ func (ll *Doubly) DelAtBeg() any {
 	if ll.Head != nil {
 		ll.Head.Prev = nil
 	}
-	return cur.Val
+	return cur.Val, true
 }
 
 // DetAtEnd Delete a node at the end of the linkedlist
-func (ll *Doubly) DelAtEnd() any {
+func (ll *Doubly[T]) DelAtEnd() (T, bool) {
```
tjgurwara99, line 71: I think it makes sense to have similar signature for `Del*` and `Add*` methods. And personally, I think its better to have `error` instead of `bool`ean value.

structure/linkedlist/doubly.go
```diff
@@ -53,9 +53,9 @@ func (ll *Doubly) AddAtEnd(val any) {
 }
 
 // DelAtBeg Delete the node at the beginning of the linkedlist
-func (ll *Doubly) DelAtBeg() any {
+func (ll *Doubly[T]) DelAtBeg() (T, bool) {
 	if ll.Head == nil {
-		return -1
+		return *new(T), false
```
tjgurwara99, line 58: Same as above

structure/linkedlist/doubly.go
```diff
@@ -37,7 +37,7 @@ func (ll *Doubly) AddAtBeg(val any) {
 }
 
 // AddAtEnd Add a node at the end of the linkedlist
-func (ll *Doubly) AddAtEnd(val any) {
+func (ll *Doubly[T]) AddAtEnd(val T) {
```
tjgurwara99, line 40: Shouldn't the signature of `Add*` and `Del*`?

structure/linkedlist/doubly.go
```diff
@@ -15,16 +15,16 @@ import "fmt"
...
 
-func NewDoubly() *Doubly {
-	return &Doubly{nil}
+func NewDoubly[T any]() *Doubly[T] {
+	return &Doubly[T]{}
 }
 
 // AddAtBeg Add a node to the beginning of the linkedlist
-func (ll *Doubly) AddAtBeg(val any) {
+func (ll *Doubly[T]) AddAtBeg(val T) {
```
tjgurwara99, line 27: Signature remark as above.

tjgurwara99 (changes requested):
I didn't fully review completely but I left a few comments on things that I think could be better in a different way.

crunchypi:
Heya, I’ve had a look at the suggestions — will think on it and reply later on today

structure/linkedlist/doubly.go
crunchypi: Sure, I used that syntax for brevity but it's most likely faster using the `var r T` syntax since it'll keep the value on the stack

structure/linkedlist/doubly_test.go
crunchypi: Agreed

structure/linkedlist/doubly.go
crunchypi: What do you mean, having some return on the `Add*` methods or? Don't think I agree in that case, since adding doesn't seem to have any possible faults, and returning the new added value is redundant (since it's passed in as args). Can't really agree on returning an error for the deleters either, since there appears to be only one fault scenario for those (and the fault is pretty clear)

tjgurwara99: Sure, keep the `bool` in the return type then 😄

By having **similar** signature, I meant something like this

```Go
func (ll *Doubly[T]) DelAtEnd() (T, bool) { ... }

func (ll *Doubly[T]) AddAtEnd() bool { ... }
```
Though after your comment, I also think that it doesn't make sense to have a return type on `Add*`...

What do you guys think @raklaptudirm @siriak ?

crunchypi: Oh I see 👽But yeah in that case i'll amend the things we've discussed either today or tomorrow, then make a beep

crunchypi:
Alright, did the amending -- should cover the discussed things (+ changed the use of `new(...)` in  `func (ll *Singly[T]) ReversePartition` into the `var _ T` syntax).

structure/linkedlist/doubly.go
tjgurwara99: This isn't resolved - your comment suggested that you'd change it but you haven't but marked it as resolved.

tjgurwara99: That's what I'm confused about so I asked other maintainers @siriak @raklaptudirm but it seems they haven't seen it since you marked it as resolved.

crunchypi: Aah alright, then i’ll leave it open until they’ve had the chance to give their opinions 👽

siriak: It seems that the only possible return value is `true` for `Add*` function because there can be no errors, so it doesn't make sense to have a boolean return value as was already stated by @crunchypi

siriak (approved):

tjgurwara99 (approved):

PR #530, opened 2022-08-29
Implement Pollard's Rho Factorization algorithm

i-redbyte:
Pollard's rho algorithm


raklaptudirm:
Is creating a new package necessary?

i-redbyte:
> Is creating a new package necessary?

Fix this

math/pollard_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+func TestRho(t *testing.T) {
+	tests := []struct {
+		name    string
+		n       *big.Int
+		g       func(*big.Int, *big.Int) *big.Int
+		want    *big.Int
+		wantErr bool
+	}{
+		{"Rho for n:  11235 ", big.NewInt(11235), DefaultPolynomial, big.NewInt(21), false},
```
raklaptudirm, line 43:
```suggestion
		{"Factor of n:  11235 ", big.NewInt(11235), DefaultPolynomial, big.NewInt(21), false},
```
I think this naming convention will be more intutive.

math/pollard.go
```diff
@@ -0,0 +1,43 @@
...
+	bigTwo := big.NewInt(2)
+	xSquared := new(big.Int).Exp(x, bigTwo, n) // x^2 % n
+	xSquared.Add(xSquared, bigOne)             // (x^2 % n) + 1
+	xSquared.Mod(xSquared, n)                  // (x^2 + 1) % n
+	return xSquared
+}
+
+// Rho is an implementation of Pollard's rho factorization algorithm
+// using the default parameters x = y = 2
+func Rho(n *big.Int, g func(*big.Int, *big.Int) *big.Int) (*big.Int, error) {
```
raklaptudirm, line 27: I think naming this function `g` is a bit misleading, because `n` is constant in `g`. Instead, I would suggest changing the signature of this function to:
```go
func(n *big.Int)(func(x *big.Int)*big.Int)
```
The value of `n` can then be provided to this function to get `g(x)`.

raklaptudirm (changes requested):
The commenting style in this algorithm assumes a lot of knowledge about Pollard's Rho algorithm, and therefore is cryptic to the layman. I would suggest rewriting so it is more intutive.

math/pollard_test.go
i-redbyte: Done

math/pollard.go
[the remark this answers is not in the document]
i-redbyte: Done

i-redbyte: Honestly, I didn't understand your idea :(

i-redbyte: the name g for the function is used in the official documentation for the algorithm https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm

raklaptudirm: > the name g for the function is used in the official documentation for the algorithm https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm

I know that. Notice that the value of `n` is not an argument to `g(x)` as given in Wikipedia. So I was suggesting to take the input as a function which takes `n` and returns `g(x)`, what you are using is `g(x, n)`.

i-redbyte: I think I understand the idea, but I don't understand how to write it correctly in the code:
`func PollardsRhoFactorization(n *big.Int, f func(n *big.Int) func(x *big.Int) *big.Int) (*big.Int, error) {
	x, y, d := big.NewInt(2), big.NewInt(2), big.NewInt(1)
	bigOne := big.NewInt(1)
	g := f(n)
	for d.Cmp(bigOne) == 0 {
		x = g(x)
		y = g(g(y))
		sub := new(big.Int).Sub(x, y)
		d.GCD(nil, nil, sub.Abs(sub), n)
	}

	if d.Cmp(n) == 0 {
		return nil, errors.New("factorization failed")
	}

	return d, nil
}`

how to pass parameters correctly now?

`func TestPollardsRhoFactorization(t *testing.T) {
	tests := []struct {
		name    string
		n       *big.Int
		f       func(n *big.Int) func(x *big.Int) *big.Int
		want    *big.Int
		wantErr bool
	}{
		{"Factor of n:  11235 ", big.NewInt(11235), ?????, big.NewInt(21), false},
	}`

raklaptudirm: If you change `DefaultPolynomial` to:
```go
func DefaultPolynomial(n *big.Int) func(*big.Int) *big.Int {
	bigOne := big.NewInt(1)
	bigTwo := big.NewInt(2)

        return func(x *big.Int) *big.Int {
	        xSquared := new(big.Int).Exp(x, bigTwo, n) // see: https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm#Core_ideas
	        xSquared.Add(xSquared, bigOne)
	        xSquared.Mod(xSquared, n)
	        return xSquared
        }
}
```
You won't need to change the tests.

raklaptudirm (approved):

siriak (approved):
Looks good, thanks!

PR #532, opened 2022-09-01
feat: Add mobius function implementation

itsAkshayDubey:
Hi,

This PR contain commits related to Mobius function implementation.
Issue #529 

Kindly have a look.

Thanks,
Akshay


math/mobius/mobius.go
```diff
@@ -0,0 +1,44 @@
...
+
+	primeFactorCount := 0
+
+	for i := 1; i <= n; i++ {
+		if n%i == 0 && p.OptimizedTrialDivision(int64(i)) {
+			if n%(i*i) == 0 {
+				return 0
+			} else {
+				primeFactorCount += 1
+			}
```
tjgurwara99, line 35: This is a redundant `else` block.

math/mobius/mobius.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+
+	if primeFactorCount%2 == 0 {
+		return 1
+	} else {
+		return -1
+	}
```
tjgurwara99, line 43: Same here. This is a redundant `else` block.

math/mobius/mobius.go
```diff
@@ -0,0 +1,44 @@
...
+package mobius
+
+import (
+	p "github.com/TheAlgorithms/Go/math/prime"
+)
+
+// Mu is the Mobius function
+// This function returns μ(n) for given number
+func Mu(n int) int {
+
```
tjgurwara99, line 22: You have quite a few newlines when they aren't necessary. For example
```Go
func Mu(n int) int {
	if n <= 1 {
		return 1
	}

 	var primeFactorCount int
 	for i := 1; i <= n; i++ {
		...
```
This is mostly preference and I don't mind if you keep it as is but as a reader, I personally find it easier to work through a codebase with good logical blocks that are separated nicely.

math/mobius/mobius.go
```diff
@@ -0,0 +1,44 @@
...
+	if n <= 1 {
+		return 1
+	}
+
+	primeFactorCount := 0
+
+	for i := 1; i <= n; i++ {
+		if n%i == 0 && p.OptimizedTrialDivision(int64(i)) {
```
tjgurwara99, line 30: Do you know the way to achieve the logical equivalent of `Mu` through Liouville function? I think that might be faster than this one... Though I do remember there was an equivalent step in that way as well so it might not be... (I'm a functional analyst not a number theorist so my statement above might be wrong... 😅)

math/mobius/mobius_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{97, -1},
+		{98, 0},
+		{99, 0},
+		{100, 0},
+	}
+	return tests
+}
+
+func TestMu(t *testing.T) {
+
```
tjgurwara99, line 34: There is no need for this extra space

math/mobius/mobius_test.go
```diff
@@ -0,0 +1,49 @@
...
+package mobius
+
+import (
+	"testing"
+)
+
+func getTestsForMu() []struct {
+	n        int
+	expected int
+} {
```
tjgurwara99, line 15: I would much rather prefer you define a new struct instead of an anonymous struct on the return type...

```suggestion
type testRes struct {
	n        int
	expected int
}

func getTestsForMu() []testRes {
```
or something along those lines...

math/mobius/mobius.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Mu is the Mobius function
+// This function returns μ(n) for given number
+func Mu(n int) int {
+
+	if n <= 1 {
+		return 1
+	}
+
+	primeFactorCount := 0
```
tjgurwara99, line 27: If the first meaningful value for you is the default value then prefer declaration over initialisation. Though arguably its more verbose and its a matter of taste so I leave it up to you to either change it or leave it as is 😄 

```suggestion
	primeFactorCount := 0
```

tjgurwara99 (changes requested):
Please consider the comments that I have left over some minor styling and logical issues. Overall I like this 😄

tjgurwara99:
Also I forgot to mention this, but I think you should move your implementation to the `math` package rather than a `mobius` package.

math/mobius/mobius.go
itsAkshayDubey: Fixed this, in this [commit](https://github.com/TheAlgorithms/Go/pull/532/commits/99228ad2465fc6fba55b08f7b1b10df9153601c1).

math/mobius/mobius_test.go
itsAkshayDubey: Fixed [here](https://github.com/TheAlgorithms/Go/pull/532/commits/af9e45b37abe3ec7372f60bec064f063e932e18a)

math/mobius/mobius_test.go
itsAkshayDubey: Fixed [here](https://github.com/TheAlgorithms/Go/pull/532/commits/f7431a5908ad8c370b9fdd94c6290a2df921687d)

math/mobius/mobius.go
itsAkshayDubey: Fixed. in this [commit](https://github.com/TheAlgorithms/Go/pull/532/commits/735f339ad20576b3ca6facb0bc8dddcbdbdecbfb)

math/mobius/mobius.go
itsAkshayDubey: Fixed, in this [commit](https://github.com/TheAlgorithms/Go/pull/532/commits/735f339ad20576b3ca6facb0bc8dddcbdbdecbfb)

math/mobius/mobius.go
itsAkshayDubey: Yes, Mu is the same as the Liouville function with an added check for repeated prime factors. We can create Liouville function and call it in Mu with an additional check for repeated prime factors and in this Liouville function, we can use logic faster than this one. Something like I did in the JS repo as below:
`return number <= 0 ? 'Number must be greater than zero.' : PrimeFactors(number).length !== new Set(PrimeFactors(number)).size ? 0 : PrimeFactors(number).length % 2 === 0 ? 1 : -1`

itsAkshayDubey: We have primefactorization.go in this repo which returns a map with prime factor as key and occurrence of that factor as value, we can use this in Liouville function and further use this function in Mu.

tjgurwara99 (approved):

math/mobius.go
```diff
@@ -0,0 +1,38 @@
...
+//   μ(n) = −1 if n is a square-free positive integer with an odd number of prime factors.
+//   μ(n) = 0 if n has a squared prime factor.
+// wikipedia: https://en.wikipedia.org/wiki/M%C3%B6bius_function
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see mobius_test.go
+
+package math
+
+import (
+	p "github.com/TheAlgorithms/Go/math/prime"
```
tjgurwara99, line 16: Can you keep this `prime` instead of changing to `p`

tjgurwara99 (changes requested):

math/mobius.go
itsAkshayDubey: Fixed in [this](https://github.com/TheAlgorithms/Go/pull/532/commits/d784952d98d049ba87023ab3b71230e7bccc4f39) commit

tjgurwara99 (approved):

siriak (approved):

PR #538, opened 2022-09-10
Generic Sorting tests

ahmed-deftoner:
Added tests for float and string slices since most of the sorting algorithms are generic now.

### DIFF
--- a/sort/sorts_test.go
+++ b/sort/sorts_test.go
@@ -74,6 +74,73 @@ func testInt(t *testing.T, sortingFunction func([]int) []int) {
 	}
 }
 
+func testFloats(t *testing.T, sortingFunction func([]float32) []float32) {
+	sortTestsInt := []struct {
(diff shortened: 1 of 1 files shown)


tjgurwara99 (changes requested):
Personally I'm of the opinion that this is not a good change. What you're testing is a language feature not the implementation. It's like you're testing `int8`, and `int64`. Keep in mind that `string`, `float32` and `int` all have the `<` operator defined on them and these functions can fail **if and only if** the language changes the `<` operator for these types.

@siriak @raklaptudirm any opinions?

If the other maintainers (^) want to see this change then I think you need to properly abstract the tests rather than create three separate functions.

ahmed-deftoner:
You are right, personally, I was in the habit of testing generic code for different types in Rust, but generics in Go are new and unit testing is also different so my code seems immature on second viewing 😞

Issue #535, opened 2022-09-07
Binomial Coefficient implementation


itsAkshayDubey:
Hi,

I have not seen any implementation of [Binomial Coefficient](https://en.wikipedia.org/wiki/Binomial_coefficient) in this repo.
Could you please assign this issue to me?

Regards,
Akshay

raklaptudirm:
No need to open issues for implementing algorithms.

PR #539, opened 2022-09-12
algorithm: binomial coefficient

itsAkshayDubey:
Issue #535

### DIFF
--- a/math/binomialcoefficient.go
+++ b/math/binomialcoefficient.go
@@ -15,7 +15,7 @@ import (
 
 // C is Binomial Coefficient function
 // This function returns C(n, k) for given n and k
-func C(n int, k int) (int, error) {
+func Combinations(n int, k int) (int, error) {
 	var ErrPosArgsOnly error = errors.New("arguments must be positive")
 	if n < 0 || k < 0 {
 		return -1, ErrPosArgsOnly
--- a/math/binomialcoefficient.go
+++ b/math/binomialcoefficient.go
@@ -16,8 +16,9 @@ import (
 // C is Binomial Coefficient function
 // This function returns C(n, k) for given n and k
 func C(n int, k int) (int, error) {
+	var ErrPosArgsOnly error = errors.New("arguments must be positive")
 	if n < 0 || k < 0 {
-		return -1, errors.New("arguments must be positive")
+		return -1, ErrPosArgsOnly
 	}
 	if k > (n - k) {
 		k = n - k
--- a/math/binomialcoefficient.go
+++ b/math/binomialcoefficient.go
@@ -13,10 +13,11 @@ import (
 	"errors"
 )
 
+var ErrPosArgsOnly error = errors.New("arguments must be positive")
+
 // C is Binomial Coefficient function
 // This function returns C(n, k) for given n and k
 func Combinations(n int, k int) (int, error) {
-	var ErrPosArgsOnly error = errors.New("arguments must be positive")
 	if n < 0 || k < 0 {
 		return -1, ErrPosArgsOnly
 	}
--- a/math/binomialcoefficient.go
+++ b/math/binomialcoefficient.go
@@ -1,0 +1,31 @@
+// binomialcoefficient.go
+// description: Returns C(n, k)
+// details:
+// a binomial cofficient C(n,k) gives number ways
+// in which k objects can be chosen from n objects.
+// wikipedia: https://en.wikipedia.org/wiki/Binomial_coefficient
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see binomialcoefficient_test.go
+
+package math
+
+import (
+	"errors"
+)
+
+// C is Binomial Coefficient function
+// This function returns C(n, k) for given n and k
+func C(n int, k int) (int, error) {
+	if n < 0 || k < 0 {
+		return -1, errors.New("arguments must be positive")
+	}
+	if k > (n - k) {
+		k = n - k
+	}
+	res := 1
+	for i := 0; i < k; i++ {
+		res *= (n - i)
+		res /= (i + 1)
+	}
+	return res, nil
+}
--- a/math/binomialcoefficient.go
+++ b/math/binomialcoefficient.go
@@ -1,7 +1,7 @@
 // binomialcoefficient.go
 // description: Returns C(n, k)
 // details:
-// a binomial cofficient C(n,k) gives number ways
+// a binomial coefficient C(n,k) gives number ways
 // in which k objects can be chosen from n objects.
 // wikipedia: https://en.wikipedia.org/wiki/Binomial_coefficient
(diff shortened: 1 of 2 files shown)


tjgurwara99:
Mark PR's as draft when you're working on them so that maintainers don't get pinged continuously on every change 😄

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,39 @@
...
+		result, error := algmath.C(test.n, test.k)
+		t.Log(test.n, " ", result, " ", error)
+		if result != test.expected {
+			t.Errorf("Wrong result! Expected:%v, Returned:%v, Error:%v", test.expected, result, error.Error())
+		}
+	}
+}
+func BenchmarkBinomialCoefficient(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		algmath.C(65536, 65536)
```
tjgurwara99, line 37: This comes up because you have two return values and one of which is an `error` type. You can just write it like this

```suggestion
		_, _ = algmath.C(65536, 65536)
```
which would make it clear that they are just ignored and not forgotten to the linter.

itsAkshayDubey: Fixed [here](https://github.com/TheAlgorithms/Go/pull/539/commits/9d063f38d84e2228b8a476ec3ec0a537e35e01b5)

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,39 @@
...
+// description: Returns C(n, k)
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see binomialcoefficient.go
+
+package math_test
+
+import (
+	"testing"
+
+	algmath "github.com/TheAlgorithms/Go/math"
```
tjgurwara99, line 11: Why not just import it as is without `algmath` when there is no naming conflict now?

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,39 @@
...
+		n        int
+		k        int
+		expected int
+	}{
+		{5, 2, 10},
+		{7, 4, 35},
+		{0, 0, 1},
+		{-1, 1, -1},
+		{1, -1, -1},
+		{-1, -1, -1},
```
tjgurwara99, line 25: These are all errors but you're not checking whether you actually even get errors. If you define a new error variable 

```Go
var ErrPosArgsOnly error = errors.New("arguments must be positive")
```
and change the return statement in line 20 of the implementation file to 

```
    return -1, ErrPosArgsOnly
```
then you can check that indeed you get that exact error in this table test.

tjgurwara99 (changes requested):

math/binomialcoefficient.go
```diff
@@ -0,0 +1,32 @@
...
+package math
+
+import (
+	"errors"
+)
+
+// C is Binomial Coefficient function
+// This function returns C(n, k) for given n and k
+func C(n int, k int) (int, error) {
+	var ErrPosArgsOnly error = errors.New("arguments must be positive")
```
tjgurwara99, line 19: This needs to be outside the function not local to the function so that you can use it inside the test rather than creating a new error inside test.

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,51 @@
...
+		if result == -1 && error == nil {
+			t.Errorf("Wrong result! Expected:%v, Returned:%v", expectedError, error.Error())
+		}
+		if result != -1 && error != nil {
+			t.Errorf("Wrong result! Expected:%v, Returned:%v", nil, error.Error())
+		}
+		if result == -1 && !strings.EqualFold(expectedError.Error(), error.Error()) {
+			t.Errorf("Wrong result! Expected:%v, Returned:%v", expectedError.Error(), error.Error())
+		}
+	}
```
tjgurwara99, line 45: Put this under a `t.Run` block.


For example
```Go
t.Run(name, func(t *testing.T) {
            got := Split(tc.input, tc.sep)
            if !reflect.DeepEqual(tc.want, got) {
                t.Fatalf("expected: %v, got: %v", tc.want, got)
            }
        })
```

In this case,

```suggestion
		t.Run(name, func(t *testing.T) {
			result, err := math.C(test.n, test.k)
			if result != test.expected || test.expectedErr != err {
				t.Errorf("expected error: %s, got: %s; expected value: %v, got: %v", test.expectedErr, err, test.expectedErr)
			}	
        	})
	}
```

The above is more than enough - no need for `t.Log` and 4 different cases when one suffices.

This also means that you would have to update your test case struct to include `expectedErr` field and use the `math.ErrPosArgsOnly` (after you have taken it out of your function and defined it inside the package instead) on the ones you expect to fail rather than creating a new `error` and checking `string` values.

tjgurwara99 (changes requested):

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{"n = 5, k = 2", 5, 2, 10, nil},
+		{"n = 7, k = 4", 7, 4, 35, nil},
+		{"n = 0, k = 0", 0, 0, 1, nil},
+		{"n = -1, k = 1", -1, 1, -1, math.ErrPosArgsOnly},
+		{"n = 1, k = -1", 1, -1, -1, math.ErrPosArgsOnly},
+		{"n = -1, k = -1", -1, -1, -1, math.ErrPosArgsOnly},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			result, error := math.Combinations(test.n, test.k)
```
tjgurwara99, line 31: This is good but `error` is a name for an actual interface which is being shadowed here.

```suggestion
			result, err := math.Combinations(test.n, test.k)
```

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,42 @@
...
+			if result != test.expectedValue || test.expectedError != error {
```
tjgurwara99, line 32:
```suggestion
			if result != test.expectedValue || test.expectedError != err {
```

math/binomialcoefficient_test.go
```diff
@@ -0,0 +1,42 @@
...
+				t.Errorf("expected error: %s, got: %s; expected value: %v, got: %v", test.expectedError, error, test.expectedValue, result)
```
tjgurwara99, line 33:
```suggestion
				t.Errorf("expected error: %s, got: %s; expected value: %v, got: %v", test.expectedError, err, test.expectedValue, result)
```

tjgurwara99 (dismissed):
Left some non-blocking comments. It's quite standard to call error variables as `err` in Go - I recommend reading more about Go idioms 😄

raklaptudirm:
Shadowing predeclared identifiers can lead to a lot of confusion, so will await the change.

tjgurwara99 (approved):

raklaptudirm (approved):

PR #540, opened 2022-09-16
algorithm: Liouville lambda function

itsAkshayDubey:
This PR contains commits related to the implementation of [Liouville lambda function](https://en.wikipedia.org/wiki/Liouville_function). Kindly have a look.

### DIFF
--- a/math/liouville.go
+++ b/math/liouville.go
@@ -22,14 +22,14 @@ var ErrNonZeroArgsOnly error = errors.New("arguments must be greater than zero")
 // Lambda is the liouville function
 // This function returns λ(n) for given number
 func LiouvilleLambda(n int) (int, error) {
-	if n < 0 {
+	switch {
+	case n < 0:
 		return 0, ErrPosArgsOnly
-	}
-	if n == 0 {
+	case n == 0:
 		return 0, ErrNonZeroArgsOnly
-	}
-	if len(prime.Factorize(int64(n)))%2 == 0 {
+	case len(prime.Factorize(int64(n)))%2 == 0:
 		return 1, nil
+	default:
+		return -1, nil
 	}
-	return -1, nil
 }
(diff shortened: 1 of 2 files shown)


math/liouville.go
```diff
@@ -0,0 +1,35 @@
...
+	"errors"
+
+	"github.com/TheAlgorithms/Go/math/prime"
+)
+
+var ErrNonZeroArgsOnly error = errors.New("arguments must be greater than zero")
+
+// Lambda is the liouville function
+// This function returns λ(n) for given number
+func Lambda(n int) (int, error) {
```
tjgurwara99, line 24: There are many things in CS that are called `Lambda`. I think you should make this `LiouvilleLambda` or `LiouvilleFunc` to make it easier to not get this confused with other `Lambda`s.

math/liouville.go
```diff
@@ -0,0 +1,35 @@
...
+	if n < 0 {
+		return 0, ErrPosArgsOnly
+	}
+	if n == 0 {
+		return 0, ErrNonZeroArgsOnly
+	}
+	if len(prime.Factorize(int64(n)))%2 == 0 {
+		return 1, nil
+	}
+	return -1, nil
```
tjgurwara99, line 34: When you see more than 2 `if` statements, consider whether you can use a `switch` case instead. This is mostly styling preference and therefore not blocking.

tjgurwara99 (changes requested):

math/liouville.go
```diff
@@ -0,0 +1,35 @@
...
+
+package math
+
+import (
+	"errors"
+
+	"github.com/TheAlgorithms/Go/math/prime"
+)
+
+var ErrNonZeroArgsOnly error = errors.New("arguments must be greater than zero")
```
tjgurwara99, line 20: The name is not correct. The name of the error implies that the arguments also accepts negative number but the comment inside the error indicates that only positive numbers are allowed.

tjgurwara99 (changes requested):

tjgurwara99 (approved):

siriak (approved):
Looks good, thanks!

PR #544, opened 2022-09-29
Enhancements for stack implementation using slice

### LINKED ISSUE #543: Use slice appending instead of insertion at zeroth index in "stack array" implementation
**Description**
Current [stack array](https://github.com/TheAlgorithms/Go/blob/eaa2be2ffe9295b9eb996ec49af6e0e2e6381595/structure/stack/stackarray.go) implementation stores the last inserted value at zeroth index of the backing slice. This is inefficient as each push operation will require a new allocation and copying of contents in the current slice.
```go
func stackPush(n any) {
	stackArray = append([]any{n}, stackArray...)
}
```
Inserting latest element to the last index (append) will also help in reusing existing slices's remaining capacity when multiple push and pop operations happens in mixed order.
Also, the current implementation uses a global variable as the backing slice. It should be implemented as an object and global mutable variables should be avoided.

BulkBeing:
Closes #543


structure/stack/stack_array.go
```diff
@@ -0,0 +1,89 @@
...
+```
+However, this exposes the underlaying storage (slice) outside the package.
+A struct is used instead, so that the underlying storage is not accessible outside the package.
+*/
+
+// StackArray is an implementation of stack with slice as underlying storage.
+// ```
+// stack := new(StackArray[int])
+// ```
+type StackArray[T any] struct {
```
raklaptudirm, line 25: The word stack is already in the package name. The name `stack.StackArray` can be shortened to `stack.Array`.

structure/stack/stack_array.go
```diff
@@ -0,0 +1,89 @@
...
+// Push inserts a new element to the stack
+// Push on a nil stack will panic
+func (s *StackArray[T]) Push(val T) {
+	s.store = append(s.store, val)
+}
+
+// Peek the last inserted element without removing it from the stack
+// Return type is a pointer to represent an optional value.
+// If the stack is empty, nil value is returned
+func (s *StackArray[T]) Peek() *T {
```
raklaptudirm, line 42: Why return a pointer? Return a value. Is the stack is empty, return an error.

raklaptudirm (changes requested):

structure/stack/stack_array.go
BulkBeing: The `stack.Array` naming convention is the idiom in general. However, in this specific case, someone trying to learn the implementation in Go might be looking at this file alone (not with the perspective as a library code). Should I use the name `Array` itself or something like `ArrayStore` or `ArrayStack` (similar to `BtreeMap` - ordered map based on a B-Tree)

tjgurwara99: > However, in this specific case, someone trying to learn the implementation in Go might be looking at this file alone (not with the perspective as a library code).

Personally, I think someone trying to learn an implementation in Go should also learn the idioms of Go so that they learn the idioms too 😄 

I think the name `Array` should suffice.

raklaptudirm (approved):

structure/stack/stack_array.go
```diff
@@ -0,0 +1,89 @@
...
+type Array[T any] []T
+```
+However, this exposes the underlaying storage (slice) outside the package.
+A struct is used instead, so that the underlying storage is not accessible outside the package.
+*/
+
+// Array is an implementation of stack with slice as underlying storage.
+// ```
+// stack := stack.NewArray[int]()
+// ```
```
tjgurwara99, line 26:
```suggestion
// Array is an implementation of stack with slice as underlying storage.
// ```
// stack := stack.NewArray[int]()
// ```
// 
// Note that the type `Array` could also be implemented directly using a slice.
// ```
// type Array[T any] []T
// ```
// However, this exposes the underlaying storage (slice) outside the package.
// A struct is used instead, so that the underlying storage is not accessible
// outside the package.
```

tjgurwara99 (changes requested):
I have a few general thoughts that I think might actually be better for this implementation.

First thing to note is that there is already an implementation of Stack. This means that to make this into a really self contained package, I think you should create an `interface` that both implementations adhere to and expose only the interface outside the package.

structure/stack/stack_array.go
BulkBeing: @tjgurwara99 Updated with all changes. Interface is specified in `stack/stack.go` file.

structure/stack/doubly_linked_list.go
```diff
@@ -0,0 +1,93 @@
+package stack
+
+import (
+	"container/list"
```
tjgurwara99, line 4: Please use our internal data structures when creating new ones. Using `container` package hides a lot of material that could be used for learning.

tjgurwara99 (changes requested):

structure/stack/doubly_linked_list.go
BulkBeing: This code was already present (I just renamed the file). Since there is another implementation using linked list, shall I remove this one?

tjgurwara99: Keep it there, but create a new Issue to fix this 🙏🏼

structure/stack/linked_list_test.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+	})
+
+	t.Run("Stack Length", func(t *testing.T) {
+		if st.Len() != 2 {
+			t.Errorf("Expected stack length to be 2, got %d", st.Len())
+		}
+	})
+
+	st.Pop()
```
tjgurwara99, line 34: This needs to be checked. Please fix all the listing and styling errors raised by the CI.

tjgurwara99 (changes requested):

PR #545, opened 2022-10-01
feat: add lerp

b5710546232:
Add function lerp for calculating liner interpolation
 this function will return a value between input v0, and v1 with t%

### DIFF
--- a/math/lerp.go
+++ b/math/lerp.go
@@ -1,0 +1,6 @@
+package math
+
+func Lerp(v0, v1, t float64) float64 {
+	// see: https://en.wikipedia.org/wiki/Linear_interpolation
+	return (1-t)*v0 + t*v1
+}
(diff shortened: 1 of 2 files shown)


math/lerp.go
```diff
@@ -0,0 +1,6 @@
+package math
+
+func Lerp(v0, v1, t float64) float64 {
```
tjgurwara99, line 3: Please provide a good documentation comment for this function. Just having a wikipedia link is not a good strategy for contributing code to projects.

tjgurwara99 (changes requested):
Please provide description when contributing.

b5710546232:
Hi  @tjgurwara99  Thank for feedback, I have added :)

tjgurwara99 (approved):

raklaptudirm (approved):

Issue #480, opened 2022-03-24
Add `FuzzXxx` tests to cipher algorithms


raklaptudirm:
**Description**
Use the new `go1.18` `FuzzXxx` tests to test cipher algorithm encryption and decryption.
The tests should accept fuzzed strings, encrypt then decrypt them, and check for equality with the original. This can help catch more elusive bugs in the code.

HectorMalot:
I looked at this for a bit, but have a few questions for the fuzzing. As I'm writing the tests, I'm finding some edge cases and wanted to ask first if/how you want algorithms to change 

1. For rot13/caesar encryption: the current algorithms only expect a-zA-Z as input, but don't validate this. This can be either be solved on the fuzzer (only allowing valid inputs), or on the algorithm (providing a helpful error is an input is outside of the expected range. Any preference or other ideas?

2. Similarly, RSA encrypt/decrypt as currently tested fails on certain inputs. Specifically, in the test `p` and `q` are set to 61 and 53, leading to a modulus of 3233. e.g. try encrypting and then decrypting `string("\x82")` with these exponents. This fails because the expected output is `int64(65533)` but the actual output is `int64(873)`, which is equal to `65533%3233`. Notably, every input above `rune(3232)` will fail because the modulus is not large enough. 

     * Option 1: Encrypt byte by byte (the current string-rune conversion can sometimes return a different output than expected)
     * Option 2: invalidate modulus smaller than MaxInt64
     * Option 3: invalidate inputs with runes larger than the provided modulus
     * Option 4: leave as is, update test to larger modulus (I'll share a PR that does this a bit later)

Any preference / ideas on the above?

siriak:
1. Validate input in the algorithm
2. Option 1 or 2 whichever is simpler

mcaci:
Hi everyone, I have been adding some fuzz tests lately to contribute to this issue. I think now once these PRs are good on your side and merged there should be no more work to do on this issue. Let me know if you have any feedbacks about this. Thanks again!

LeonxLJX:
I'd like to take this one (`Add `FuzzXxx` tests to cipher algorithms`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

PR #549, opened 2022-10-02
Add Fuzz test for RSA

HectorMalot:
As per #480 , my proposal to include fuzzing for RSA. Note that the test will run as a normal test if fuzzing is not enabled, so I didn't duplicate the code, but rather updated the test to allow fuzzing as well as normal testing. 

Details of this change:

- Changed test from `*Testing.T` to `*Testing.F`
- Increased the testing prime numbers to larger numbers so that larger rune inputs also work. (i.e. up to MaxInt32)
- Included a unicode test case (that would fail on the previous test)

### DIFF
--- a/cipher/rsa/rsa_test.go
+++ b/cipher/rsa/rsa_test.go
@@ -33,47 +33,52 @@ var rsaTestData = []struct {
 		"Encrypt full sentence from rsacipher.go main function",
 		"I think RSA is really great",
 	},
+	{
+		"Encrypt uncommon unicode",
+		string("\x82"),
+	},
 }
 
-func TestEncryptDecrypt(t *testing.T) {
+func FuzzEncryptDecrypt(f *testing.F) {
 	// Both prime numbers
-	p := int64(61)
-	q := int64(53)
+	p := int64(3079)
+	q := int64(983)
 
 	n := p * q
 
 	delta := lcm.Lcm(p-1, q-1)
 
-	e := int64(17) // Coprime with delta
+	e := int64(2837) // Coprime with delta
 
 	if gcd.Recursive(e, delta) != 1 {
-		t.Fatal("Algorithm failed in preamble stage:\n\tPrime numbers are chosen statically and it shouldn't fail at this stage")
+		f.Fatal("Algorithm failed in preamble stage:\n\tPrime numbers are chosen statically and it shouldn't fail at this stage")
 	}
 
 	d, err := modular.Inverse(e, delta)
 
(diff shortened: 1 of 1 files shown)


tjgurwara99 (changes requested):
I personally am not in favour of removing old test and just having the fuzzing tests. I think both should be present.

HectorMalot:
Hi @tjgurwara99,

That is how I started, but I realized that it would mean a lot of duplication, or a helper that contains all this code, called by nearly identical Test… and Fuzz… functions. 

This is in the go documentation: 

> When fuzzing is disabled, the fuzz target is called with the seed inputs registered with F.Add and seed inputs from testdata/fuzz/<Name>. In this mode, the fuzz test acts much like a regular test, with subtests started with F.Fuzz instead of T.Run.

That gives me the impression that in this case, the proposed setup in the PR makes most sense. It functions as a normal test during a normal run, and can be used to fuzz when called with the fuzzing options. 

Source: https://pkg.go.dev/testing#hdr-Fuzzing

Happy to change, but hope this makes you reconsider. 

Cheers
Dennis

tjgurwara99:
> That is how I started, but I realized that it would mean a lot of duplication, or a helper that contains all this code, called by nearly identical Test… and Fuzz… functions.

Duplicate code for doing two separate things (in this case Fuzzing and Testing) are valid reasons for duplicating code.

@raklaptudirm @siriak what are your opinions?

I'm personally against having test and fuzzing mixed because of the inherent flaky nature of fuzzing.

siriak:
Both ways are acceptable IMO

HectorMalot:
***@***.***,That is how I started, but I realized that it would mean a lot of duplication, or a helper that contains all this code, called by nearly identical Test… and Fuzz… functions. This is in the go documentation: > When fuzzing is disabled, the fuzz target is called with the seed inputs registered with F.Add and seed inputs from testdata/fuzz/<Name>. In this mode, the fuzz test acts much like a regular test, with subtests started with F.Fuzz instead of T.Run.That convinced me that in this case, the proposed setup in the PR makes most sense. It functions as the normal test, and can be used to fuzz as well. Source: https://pkg.go.dev/testing#hdr-FuzzingHappy to change, but hope this makes you reconsider. CheersDennisOn 2 Oct 2022, at 19:00, Taj ***@***.***> wrote:﻿
@tjgurwara99 requested changes on this pull request.

I personally am not in favour of removing old test and just having the fuzzing tests. I think both should be present.

PR #550, opened 2022-10-04
algorithm: Twin prime

itsAkshayDubey:
This PR contains related to the addition of an algorithm to this repo that finds the twin prime of a number.

Regards,
Akshay


math/twinprime.go
```diff
@@ -0,0 +1,24 @@
+// twinprime.go
+// description: Returns Twin Prime of n
+// details:
+// For any integer n, twin prime is (n + 2)
+// if and only if both n and (n + 2) both are prime
+// wikipedia: https://en.wikipedia.org/wiki/Twin_prime
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see twinprime_test.go
+
+package math
```
tjgurwara99, line 10: There is already a [prime package](https://github.com/TheAlgorithms/Go/blob/4d9b1e48789f24959739ef03b52b26d074a98de8/math/prime/sieve.go#L6) I believe. Please move this to that package and change the name to `Twin` instead of `TwinPrime` when you do indeed move it there since `prime` is in the name.

tjgurwara99 (changes requested):

math/prime/twin.go
```diff
@@ -0,0 +1,20 @@
...
+
+// This function returns twin prime for given number
+// returns (n + 2) if both n and (n + 2) are prime
+// -1 otherwise
+func Twin(n int) int {
+	if OptimizedTrialDivision(int64(n)) && OptimizedTrialDivision(int64(n+2)) {
+		return n + 2
+	}
+	return -1
+}
```
tjgurwara99, line 20:
```suggestion
func Twin(n int) (int, bool) {
	if OptimizedTrialDivision(int64(n)) && OptimizedTrialDivision(int64(n+2)) {
		return n + 2, true
	}
	return -1, false
}
```

tjgurwara99 (changes requested):
A minor change. You might also need to update the test but its again minor.

math/prime/twin.go
itsAkshayDubey: Added this in [this](https://github.com/TheAlgorithms/Go/pull/550/commits/85c67572b1e449c24de1201d1c6efaacf3729ce8) commit :smiley: , have a look.

math/prime/twin_test.go
```diff
@@ -0,0 +1,37 @@
+// twin_test.go
+// description: Returns Twin Prime of n
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see twin.go
+
+package prime
```
tjgurwara99, line 6:
```suggestion
package prime_test
```

tjgurwara99 (changes requested):

tjgurwara99 (approved):

raklaptudirm (approved):

PR #551, opened 2022-10-06
Add Red-Black Tree

guuzaa:
## What I have implemented:

1. `NewRBTree`:  Create a new Red-Black Tree

2. `Insert`:  Insert a key into the tree

3. `InOrder`:  Traverse the tree via inorder

4. `Delete`:  Delete a node of Red-Black Tree

5. `Max`:  Return the max value of the tree

6. `Min`:  Return the min value of the tree

7. `Has`:  Determine the tree has the node of key

8. `Predecessor`:  Return the predecessor of the node of key

9. `Successor`:  Return the successor of the node of key

## References
https://www.programiz.com/dsa/red-black-tree
https://en.wikipedia.org/wiki/Red_black_tree


akshitagupta15june (approved):

structure/rbtree/rbtree.go
```diff
@@ -0,0 +1,457 @@
...
+)
+
+type Color byte
+
+const (
+	Black Color = iota
+	Red
+)
+
+type Node[T constraints.Ordered] struct {
```
tjgurwara99, line 25: Node already exists in `binarytree` package. Please either modify that and use that or create an abstract version of both since having the same thing in multiple places is not a good practice. When contributing, please read whether what you're implementing already exists or not.

structure/rbtree/rbtree.go
```diff
@@ -0,0 +1,457 @@
+// Red-Black Tree is a kind of self-balancing binary search tree.
+// Each node stores "color" ("red" or "black"), used to ensure that the tree remains balanced during insertions and deletions.
+//
+// For more details check out those links below here:
+// Programiz article : https://www.programiz.com/dsa/red-black-tree
+// Wikipedia article: https://en.wikipedia.org/wiki/Red_black_tree
+// authors [guuzaa](https://github.com/guuzaa)
+// see rbtree.go
+
+package rbtree
```
tjgurwara99, line 10: I recommend moving both the binarysearchtree and rbtree package to be moved to a parent package called `tree` inside the `structure` package.

tjgurwara99 (changes requested):

structure/rbtree/rbtree.go
guuzaa: Thanks for your suggestions. I will move `binarysearchtree` and `rbtree` package to `tree` package.

tjgurwara99: If you can, yes do it. Though if I remember correctly there were some issues with `avl`, so you might come across some issues there. Feel free to move them around and we will review the changes; if something needs to change we can iterate 😄

structure/rbtree/rbtree.go
guuzaa: The `Node ` existing in `binarysearchtree` doesn't use generics, but `Node ` I submitted does. To create an abstract version, we should modify the existing code. I am willing to do that. Can I change the existing code to use generics ? If not, I modify the Node I submitted .

structure/rbtree/rbtree.go
guuzaa: Ok, I take a try.

tjgurwara99: By moving I meant moving all the files into a single package level `tree` not sub packages like `tree/avl`, `tree/binarysearchtree` etc.

guuzaa: Got it. As per the above https://github.com/TheAlgorithms/Go/pull/551#discussion_r990012608, we should modify the `Node` to use generics. It takes time. I am willing to do it. There will be a one-day wait 😄

tjgurwara99: No rush 😄

guuzaa: Hi @tjgurwara99, I'm done 😀 . 
I have moved `avl`, `binarysearchtree` and `rbtree` to `tree` package. 
To avoid conflict, `binarysearchtree` has been wrapped into a struct: `BSTree`. 
As for the above discussed `Node`, inspired by https://github.com/emirpasic/gods/tree/master/trees , I didn't create an abstract version, but create three different versions: `Node` (for BSTree), `AVLNode` (for avl) and `RBNode` (for RBTree). 
For simplicity, I didn't change `avl`.

akshitagupta15june (dismissed):

structure/tree/avl_test.go
```diff
@@ -1,125 +1,125 @@
-package avl
+package tree
```
tjgurwara99, line 1: This test will completely break when you change this to `tree_test`. Try it!

tjgurwara99: The reason I mention this is because you have the Tree but you can access any of its values. Its quite pointless if you can't use this outside the package right?

guuzaa: Sorry, I don't get it. I think we can use `avl` outside the package. Every member of `avl` is exported.

structure/tree/avl_test.go
```diff
@@ -263,14 +263,14 @@ func TestGet(t *testing.T) {
 	Insert(&root, 3)
 
 	n := Get(root, 4)
-	if n.Key != 4 {
+	if n.key != 4 {
 		t.Error("key should be 4")
 	}
-	if n.Right.Key != 5 {
+	if n.right.key != 5 {
```
tjgurwara99, line 269: I'm talking about this. You said that every member is exported but you are only testing is not exported.

guuzaa: Sorry about that. I'm used to make each field of struct un-exported. I will make it exported as soon as possible.

tjgurwara99: Haha, I know. I have the same habit 😂

guuzaa: Exactly. I need to pay more attention to this.

tjgurwara99:
I looked over at your different node implementations, and I must say that I really think that all different kinds of `Node`s that you have added, if you take a union of all the fields inside them and use that instead of different one's you would be better off.


For example, you have these


```Go

 // Node of a binary-search tree
 type Node[T constraints.Ordered] struct {
 	Key   T
 	Left  *Node[T]
 	Right *Node[T]
 }

 // AVLNode of a AVL tree
 type AVLNode struct {
 	Key         int
 	Height      int
 	Left, Right *AVLNode
 }

 type Color byte

 const (
 	Black Color = iota
 	Red
 )

 // RBNode: Node of Red-Black Tree
 type RBNode[T constraints.Ordered] struct {
 	Key    T
 	Parent *RBNode[T]
 	Left   *RBNode[T]
 	Right  *RBNode[T]
 	Color  Color
 }
```

but instead you could just as easily have this:

```Go
type Node[T constraints.Ordered] struct {
 	Key           T
 	Parent      *Node[T]
 	Left          *Node[T]
 	Right        *Node[T]
 	Color        Color
 	Height      int
 }
```

This solves the problem of repeating and quite concise. Keep the tree's separate though since they are indeed different algorithms.

PS: for AVL you can just use the specific `Node` for example

```Go
func NewAVLTree() *Node[int] {
 	return nil
 }
```

and that should be enough.


Also, please make the AVL stuff into methods rather than functions otherwise the package is not consistent. I still need to review this PR further but think about the above suggestions (and implement them if you agree).

guuzaa:
Thanks for your suggestions. 
It's a good idea. It makes `Node` neater, but a bit wasteful, since any `tree` doesn't have to own all fields. This is a tradeoff between neatness and storage. I think that is fine. As for `avl`, I will wrap it in a struct. Honestly, it's a challenge for me, but I'm pleased to take it.  I will implement it in the following day. 
Recently, I'm applying for a job. Therefore, I won't have much spare time. But I will finish it as soon as possible.

tjgurwara99:
Take your time, there is no rush. As soon as there is an update on this PR I will be notified so I will review them whenever I have time.

Good luck with your job application!

guuzaa:
## Documentation for `tree` package:

1. `NewRBTree`:  Create a new Red-Black Tree
```go
tree := NewRBTree[int]() // without any parameters
```
2. `NewAVLTree`:  Create a new AVL Tree
```go
tree := NewAVLTree[int]() // without any parameters
```
4. `NewBSTree`:  Create a new BST
```go
tree := NewBSTree[int]() // without any parameters
```
5. `Insert`:  Insert a chain of keys into the tree.  Implemented by **BST, AVL and Red-Black Tree**.
```go
tree := NewBSTree[int]()  // or tree := NewRBTree[int]() or tree := NewAVLTree[int]()
nums := []int{1,2,3,4}
tree.Insert(nums...)
tree.Insert(10)
tree.Insert(10, 20)
// The above three ways make sense.
```
6. `InOrder`, `PreOrder` , `PostOrder` and `LevelOrder` :  Traverse the tree via four different ways. Implemented by **BST, AVL and Red-Black Tree**.
```go
tree.Insert(nums...)
ret := tree.PreOrder() 
ret = tree.InOrder()
ret = tree.PostOrder()
ret = tree.LevelOrder()
// ret is a slice of ints
```
7. `Delete`:  Delete a node of Tree.  Implemented by **BST, AVL and Red-Black Tree**.
```go
tree := NewAVLTree[int]()
tree.Insert(5)
tree.Insert(4)
tree.Insert(3)
tree.Insert(2)

tree.Delete(5)
```
8. `Max` and `Min` :  Return the max or min value of the tree. Implemented by **BST, AVL and Red-Black Tree**.
```go
key, ok := tree.Max() // if the tree is empty, ok is false, otherwise it's true.
```
10. `Has` :  Determine the tree has the node of key. Implemented by **BST, AVL and Red-Black Tree**.
```go
tree := NewAVLTree[int]()
tree.Insert(5)
tree.Insert(4)
tree.Insert(3)
tree.Insert(2) 
have := tree.Has(5) 
```
12. `Get`: Get the node of key. Implemented by **BST, AVL and Red-Black Tree**.
```go
tree := NewAVLTree[int]()
tree.Insert(5)
tree.Insert(4)
tree.Insert(3)
tree.Insert(2) 
n, ok := tree.Get(15) // if 15 isn't in the tree, ok is false, otherwise it's true.
```
12. `Depth`: Returns the depth of the tree. Implemented by **BST, AVL and Red-Black Tree**.

14. `Predecessor`:  Return the predecessor of the node of key. Only Implemented by **Red-Black Tree**.
```go
tree := NewRBTree[int]()
nums := []int{10, 8, 88, 888, 4, -1, 100}
tree.Insert(nums...)
ret, ok := tree.Predecessor(100) // ret = 88, ok = true
// If there is no predecessor, ok = false
```
15. `Successor`:  Return the successor of the node of key. Only Implemented by **Red-Black Tree**.
```go
tree := NewRBTree[int]()
nums := []int{10, 8, 88, 888, 4, -1, 100}
tree.Insert(nums...)
ret, ok := tree.Successor(100) // ret = 888, ok = true
// If there is no successor, ok = false
```

structure/tree/rbtree.go
```diff
@@ -0,0 +1,352 @@
...
+		y.Parent.Right = y
+	} else {
+		y.Parent.Left = y
+	}
+
+	y.Right = x
+	x.Parent = y
+}
+
+func (t *RBTree[T]) insertFix(k *Node[T]) {
```
tjgurwara99, line 150: Can you think of a way to reduce the nesting? I think for beginners it's quite difficult to follow this...

tjgurwara99 (changes requested):
This is looking quite good so far. Thanks for doing this! 😄 

I have left some comments mostly related to naming.

structure/tree/bstree.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// // Delete removes the node of val
+func (t *BSTree[T]) Delete(val T) *Node[T] {
+	return t.deleteHelper(t.Root, val)
+}
+
+// AccessNodesByLayer  accesses nodes layer by layer instead of printing the results as one line.
+func (t *BSTree[T]) AccessNodesByLayer() [][]T {
```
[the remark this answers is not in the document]
guuzaa: Sorry, I forgot to abstract it into `binaryTree`.  It is common in all trees. I have modified it in the following commit.

structure/tree/rbtree.go
guuzaa: I will try to do it. Honestly, I think it's impossible. As we all known, Red-Black tree is complex. 😭

tjgurwara99 (approved):

guuzaa:
My pleasure. 😄 
By the way, README file should be updated.

akshitagupta15june (approved):
LGTM

raklaptudirm (approved):

PR #554, opened 2022-10-07
feat: add perfectnumber

vil02:
#### Description of Change

I have added an utility allowing to check if given number is [_perfect_](https://en.wikipedia.org/wiki/Perfect_number).
The check is done by comparing the sum of the [proper divisors](https://mathworld.wolfram.com/ProperDivisor.html) with the input.
The proper divisors are obtained by iterating though all of the natural numbers up to the square root of the input. 

#### Checklist

Notes: 
Added [`perfectnumber.go`](https://github.com/vil02/Go/blob/ff1a8f074977ecb94ae402f1aa6e7d9b360abcf6/math/perfectnumber.go) and respective [tests](https://github.com/vil02/Go/blob/ff1a8f074977ecb94ae402f1aa6e7d9b360abcf6/math/perfectnumber_test.go).

### DIFF
--- a/math/perfectnumber_test.go
+++ b/math/perfectnumber_test.go
@@ -9,7 +9,10 @@ type SumOfProperDivisorsExample struct {
 	expected uint
 }
 
-//example data verified with https://oeis.org/A001065/b001065.txt
+// getSumOfProperDivisorsExamples returns an array of example data
+// for the tests of the function SumOfProperDivisors.
+// Example data was verified using [A001065].
+// [A001065]: https://oeis.org/A001065/b001065.txt
 func getSumOfProperDivisorsExamples() []SumOfProperDivisorsExample {
 	return []SumOfProperDivisorsExample{
 		{1, 0},
@@ -97,17 +100,9 @@ func BenchmarkSumOfProperDivisors(b *testing.B) {
 	}
 }
 
-//example data verified with https://oeis.org/A000396
-var perfectNumberExamples = []struct {
-	perfectNumber uint
-}{
-	{6},
-	{28},
-	{496},
-	{8128},
-	{33550336},
-}
-
+// getPerfectNumberExamples returns an array of some perfect numbers.
+// Example data was verified using [A000396].
+// [A000396]: https://oeis.org/A000396
 func getPerfectNumberExamples() []uint {
 	return []uint{
 		6,
--- a/math/perfectnumber_test.go
+++ b/math/perfectnumber_test.go
@@ -13,7 +13,7 @@ type testCaseSumOfProperDivisors struct {
 
 // getSumOfProperDivisorsTestCases returns an array of test data
 // for the tests of the function SumOfProperDivisors.
-// Example data was verified using [A001065].
+// Data was verified using [A001065].
 // [A001065]: https://oeis.org/A001065/b001065.txt
 func getSumOfProperDivisorsTestCases() []testCaseSumOfProperDivisors {
 	return []testCaseSumOfProperDivisors{
@@ -112,7 +112,7 @@ type isPerfectNumberTestCase struct {
 
 // getIsPerfectNumberTestCases returns an array of test data
 // for the tests of the function IsPerfectNumber
-// Example data was verified using [A000396].
+// Data was verified using [A000396].
 // [A000396]: https://oeis.org/A000396
 func getIsPerfectNumberTestCases() []isPerfectNumberTestCase {
 	return []isPerfectNumberTestCase{
--- a/math/perfectnumber_test.go
+++ b/math/perfectnumber_test.go
@@ -1,0 +1,144 @@
+package math
+
+import (
+	"testing"
+)
+
+//example data verified with https://oeis.org/A001065/b001065.txt
+var sumOfProperDivisorsExamples = []struct {
+	input    uint
+	expected uint
+}{
+	{1, 0},
+	{2, 1},
+	{3, 1},
+	{4, 3},
+	{5, 1},
+	{6, 6},
+	{7, 1},
+	{8, 7},
+	{9, 4},
+	{10, 8},
+	{11, 1},
+	{12, 16},
+	{13, 1},
+	{14, 10},
+	{15, 9},
+	{16, 15},
+	{17, 1},
+	{18, 21},
+	{19, 1},
+	{20, 22},
+	{21, 11},
+	{22, 14},
+	{23, 1},
+	{24, 36},
+	{25, 6},
+	{26, 16},
+	{27, 13},
+	{28, 28},
+	{29, 1},
+	{30, 42},
+	{31, 1},
+	{32, 31},
+	{33, 15},
+	{34, 20},
+	{35, 13},
+	{36, 55},
+	{37, 1},
+	{38, 22},
+	{39, 17},
+	{40, 50},
+	{41, 1},
+	{42, 54},
+	{43, 1},
+	{44, 40},
+	{45, 33},
+	{46, 26},
+	{47, 1},
+	{48, 76},
+	{49, 8},
+	{50, 43},
+	{51, 21},
+	{52, 46},
+	{53, 1},
+	{54, 66},
+	{55, 17},
+	{56, 64},
+	{57, 23},
+	{58, 32},
+	{59, 1},
+	{60, 108},
+	{800, 1153},
+	{8204, 8260},
+	{9646, 8498},
+	{9756, 14996},
+	{9829, 1},
+	{9954, 15006},
+	{10000, 14211},
+}
+
+func TestSumOfProperDivisors(t *testing.T) {
+	for _, tc := range sumOfProperDivisorsExamples {
+		actual := SumOfProperDivisors(tc.input)
+		if actual != tc.expected {
+			t.Errorf("Expected SumOfProperDivisors(%d) to be: %d, but got: %d", tc.input, tc.expected, actual)
+		}
+	}
+}
+
+func BenchmarkSumOfProperDivisors(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		SumOfProperDivisors(800)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


math/perfectnumber_test.go
```diff
@@ -0,0 +1,144 @@
+package math
+
+import (
+	"testing"
+)
+
+//example data verified with https://oeis.org/A001065/b001065.txt
+var sumOfProperDivisorsExamples = []struct {
```
tjgurwara99, line 8: This is a mutable value, try to change that into a return of a function so that this variable is always deterministic in multiple tests.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,144 @@
+package math
+
+import (
+	"testing"
+)
+
+//example data verified with https://oeis.org/A001065/b001065.txt
```
tjgurwara99, line 7: The comments styling in Go is very specific. Please read up on it [here](https://tip.golang.org/doc/comment)

math/perfectnumber_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func BenchmarkSumOfProperDivisors(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		SumOfProperDivisors(800)
+	}
+}
+
+//example data verified with https://oeis.org/A000396
```
tjgurwara99, line 96: Comment needs to follow the convention linked in the previous comment.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,144 @@
...
+var perfectNumberExamples = []struct {
```
tjgurwara99, line 97: This is also mutable. Please make this value as a return of a function rather than a package level value as it can easily be modified.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func TestIsPerfectNumber(t *testing.T) {
+	for _, tc := range perfectNumberExamples {
+		if !IsPerfectNumber(tc.perfectNumber) {
+			t.Errorf("%d is expected to be a perfect number.", tc.perfectNumber)
+		}
+	}
+}
+
+var notPerfectNumberExamples = []struct {
```
tjgurwara99, line 115: Same applies here.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,144 @@
+package math
```
tjgurwara99, line 1:
```suggestion
package math_test
```

tjgurwara99 (changes requested):

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
+package math_test
+
+import (
+	"testing"
+
+	"github.com/TheAlgorithms/Go/math"
+)
+
+type sumOfProperDivisorsExample struct {
```
tjgurwara99, line 9:
```suggestion
type testCase struct {
```

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
...
+	input    uint
+	expected uint
+}
+
+// getSumOfProperDivisorsExamples returns an array of example data
+// for the tests of the function SumOfProperDivisors.
+// Example data was verified using [A001065].
+// [A001065]: https://oeis.org/A001065/b001065.txt
+func getSumOfProperDivisorsExamples() []sumOfProperDivisorsExample {
```
tjgurwara99, line 18:
```suggestion
func getTestCases() []sumOfProperDivisorsExample {
```

tjgurwara99 (changes requested):

math/perfectnumber_test.go
vil02: I am not sure if this is a good idea to use a _generic_ name like `testCase` here, since there are more functions returning some test cases in this file. Maybe getting rid of the type `sumOfProperDivisorsExample` and returning a `map[uint]uint` from the function `getSumOfProperDivisorsExamples()` would be an option? Something like:

```golang
func getSumOfProperDivisorsExamples() map[uint]uint {
	return map[uint]uint{
		0:     0,
		1:     0,
		2:     1,
		3:     1,
               //...
	}
}

func TestSumOfProperDivisors(t *testing.T) {
	for input, expected := range getSumOfProperDivisorsExamples() {
		actual := math.SumOfProperDivisors(input)
		if actual != expected {
			t.Errorf(
				"Expected SumOfProperDivisors(%d) to be: %d, but got: %d",
				input, expected, actual)
		}
	}
}
```

math/perfectnumber_test.go
vil02: As above.

math/perfectnumber_test.go
tjgurwara99: My point here was not to have generic name `testCase` but rather a prefix - I used that same comment in many places and in many of your PR's since you tend to use the `Example` suffix a lot. The suffix you use is `Example` which is not really an example but a case for a test - it should also contain bad cases to make sure that they test the bad route of a function.

Use your own judgement here, if there are two things that could be called a testCase then add more context to it. For example, `testCaseSumOfProperDivisors` though I dislike the long name, it's an option.

>  Maybe getting rid of the type sumOfProperDivisorsExample and returning a map[uint]uit from the function getSumOfProperDivisorsExamples() would be an option?

Do you really think it helps someone reading your code?

I would also recommend adding more context in the other PR's where you've added `testCase` to something like `testCaseXyz`.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
...
+type testCaseSumOfProperDivisors struct {
+	input    uint
+	expected uint
+}
+
+// getSumOfProperDivisorsTestCases returns an array of example data
+// for the tests of the function SumOfProperDivisors.
+// Example data was verified using [A001065].
+// [A001065]: https://oeis.org/A001065/b001065.txt
+func getSumOfProperDivisorsTestCases() []testCaseSumOfProperDivisors {
```
tjgurwara99, line 18: There is no testCase for the bad path...

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
...
+func BenchmarkSumOfProperDivisors(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		math.SumOfProperDivisors(800)
+	}
+}
+
+// getPerfectNumberExamples returns an array of some perfect numbers.
+// Example data was verified using [A000396].
+// [A000396]: https://oeis.org/A000396
+func getPerfectNumberExamples() []uint {
```
tjgurwara99, line 111: `TestCases` not `Examples`.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
...
+	return []uint{
+		6,
+		28,
+		496,
+		8128,
+		33550336}
```
tjgurwara99, line 117:
```suggestion
		33550336,
	}
```

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+func TestIsPerfectNumber(t *testing.T) {
+	for _, tc := range getPerfectNumberExamples() {
+		if !math.IsPerfectNumber(tc) {
+			t.Errorf("%d is expected to be a perfect number.", tc)
+		}
+	}
+}
+
+func getNotPerfectNumberExamples() []uint {
```
tjgurwara99, line 128: It is probably better to have the testCases in one place under a constant struct similar to what you did with the other one.

```Go
type perfectNumTestCase struct {
	num      int
	perfect bool
}
```

and that should be enough to combine your `getPerfectNumberExamples` and `getNotPerfectNumberExamples` into one consistent function.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,157 @@
...
+		7,
+		100,
+		219,
+		997,
+		33550335,
+		33550337,
+	}
+}
+
+func TestIsNotPerfectNumber(t *testing.T) {
```
tjgurwara99, line 145: Once you apply the above, you should remove this since this will be redundant.

If you really want different outputs please read up on subtests [here](https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks).

tjgurwara99 (changes requested):

math/perfectnumber_test.go
vil02: I really do not see a bad path in the function `SumOfProperDivisors(inNumber uint) uint` (integer overflow possibly?). Could you please explain in more detail what do you have in mind?

tjgurwara99: Nevermind, I didn't see the test cases properly. The `{0, 0}` and `{1, 0}` are the bad cases in my mind.

math/perfectnumber_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/TheAlgorithms/Go/math"
+)
+
+type testCaseSumOfProperDivisors struct {
+	input    uint
+	expected uint
+}
+
+// getSumOfProperDivisorsTestCases returns an array of example data
```
tjgurwara99, line 14:
```suggestion
// getSumOfProperDivisorsTestCases returns an array of test data
```

math/perfectnumber_test.go
```diff
@@ -0,0 +1,151 @@
...
+		math.SumOfProperDivisors(800)
+	}
+}
+
+type isPerfectNumberTestCase struct {
+	number    uint
+	isPerfect bool
+}
+
+// getIsPerfectNumberTestCases returns an array of example data
```
tjgurwara99, line 113:
```suggestion
// getIsPerfectNumberTestCases returns an array of test data
```

tjgurwara99 (changes requested):
Other than the following comments I think everything else looks good 😄

tjgurwara99 (approved):

raklaptudirm (approved):

Issue #552, opened 2022-10-07
Mergesort using goroutines 


CorentinGS:
**Description**
It would be interesting to add a mergesort algorithm that uses goroutines. It's a good way to show how routines can be used to improve classical algorithm. 

For enhancement:
- Add a ParallelMergesort function that use goroutines  
- Benchmark & test it and compare how it performs against classical mergesort 

I can do a pr in the weekend if you are ok with that suggestion.

raklaptudirm:
Don't open issues for missing implementations. Just create a pull request and we will review it.

PR #558, opened 2022-10-08
feat: Parallel MergeSort using goroutines

CorentinGS:
#### Description of Change

Related issue:  #552 

I've added the Parallel MergeSort algorithm using goroutines and the corresponding tests.

#### Checklist

### DIFF
--- a/sort/mergesort.go
+++ b/sort/mergesort.go
@@ -61,6 +61,7 @@ func MergeIter[T constraints.Ordered](items []T) []T {
 	return items
 }
 
+// MergeParallel Perform merge sort on a slice using goroutines
 func MergeParallel[T constraints.Ordered](items []T) []T {
 
 	if len(items) < 2 {
--- a/sort/mergesort.go
+++ b/sort/mergesort.go
@@ -60,3 +60,28 @@ func MergeIter[T constraints.Ordered](items []T) []T {
 	}
 	return items
 }
+
+func MergeParallel[T constraints.Ordered](items []T) []T {
+
+	if len(items) < 2 {
+		return items
+
+	}
+
+	if len(items) < 2048 {
+		return Merge(items)
+	}
+
+	done := make(chan struct{})
+	defer close(done)
+
+	var middle = len(items) / 2
+	var a []T
+	go func() {
+		a = MergeParallel(items[:middle])
+		done <- struct{}{}
+	}()
+	var b = MergeParallel(items[middle:])
+	<-done
+	return merge(a, b)
+}
--- a/sort/mergesort.go
+++ b/sort/mergesort.go
@@ -61,7 +61,7 @@ func MergeIter[T constraints.Ordered](items []T) []T {
 	return items
 }
 
-// MergeParallel Perform merge sort on a slice using goroutines
+// ParallelMerge Perform merge sort on a slice using goroutines
 func ParallelMerge[T constraints.Ordered](items []T) []T {
 
 	if len(items) < 2 {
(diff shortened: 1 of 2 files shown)


sort/mergesort.go
```diff
@@ -60,3 +60,29 @@ func MergeIter[T constraints.Ordered](items []T) []T {
...
+
+// MergeParallel Perform merge sort on a slice using goroutines
+func MergeParallel[T constraints.Ordered](items []T) []T {
+
+	if len(items) < 2 {
+		return items
+
+	}
+
+	if len(items) < 2048 {
```
tjgurwara99, line 72: Your implementation is never tested.

CorentinGS: Hi ! 

Indeed this conditions has been added to prevent having too many goroutines.

Would you prefer me to remove it even if it might cause performance issues or should I add a specific test case using a bigger array like this one : 
```go
	t.Run("MergeParallel on big array", func(t *testing.T) {
		rand.Seed(time.Now().UnixNano())
		size := 100000
		randomBigArray := make([]int, size)
		for i := range randomBigArray {
			randomBigArray[i] = rand.Intn(size)
		}

		sortedArray := sort.MergeParallel[int](randomBigArray)
		for i := 0; i < len(sortedArray)-1; i++ {
			if sortedArray[i] > sortedArray[i+1] {
				t.Errorf("MergeParallel failed")
			}
		}
	})
```

tjgurwara99: Add a specific test rather than sticking to the framework in the test package for sorting tests.

CorentinGS: Fixed in [8ad3642](https://github.com/TheAlgorithms/Go/pull/558/commits/8ad364295e212aa906d1e061e85a688d7a89a13a)

sort/mergesort.go
```diff
@@ -60,3 +60,29 @@ func MergeIter[T constraints.Ordered](items []T) []T {
 	}
 	return items
 }
+
+// MergeParallel Perform merge sort on a slice using goroutines
+func MergeParallel[T constraints.Ordered](items []T) []T {
+
```
tjgurwara99, line 66: Non-blocking but this newline here is really not looking nice to me 😂

sort/mergesort.go
```diff
@@ -60,3 +60,29 @@ func MergeIter[T constraints.Ordered](items []T) []T {
...
+	defer close(done)
+
+	var middle = len(items) / 2
+	var a []T
+	go func() {
+		a = MergeParallel(items[:middle])
+		done <- struct{}{}
+	}()
+	var b = MergeParallel(items[middle:])
+	<-done
```
tjgurwara99, line 86: non-blocking but I think wait groups might be better here (which internally use this anyways but they have more protection built into them)

sort/mergesort.go
```diff
@@ -60,3 +60,29 @@ func MergeIter[T constraints.Ordered](items []T) []T {
 	}
 	return items
 }
+
+// MergeParallel Perform merge sort on a slice using goroutines
+func MergeParallel[T constraints.Ordered](items []T) []T {
```
tjgurwara99, line 65:
```suggestion
func ParallelMerge[T constraints.Ordered](items []T) []T {
```

tjgurwara99 (changes requested):

sort/mergesort.go
CorentinGS: This should be ok now

tjgurwara99 (approved):
NIce! LGTM, thanks!

raklaptudirm (approved):

PR #559, opened 2022-10-08
feat(cipher/caesar): add fuzzy test to caesar cipher

VictorAssunc:
Added Fuzzy test to Caesar Cipher

I had to make a small change in the algorithm approach and treat the input text as a byte array, as the old way some characters were converted wrong. An example obtained when running fuzzy was the character `ǃ` (note that this is different from `!`) was converted to `�` and not returned to normal in `Decrypt`.

image
In this image you can better see the difference between the two characters (left: the one responsible for the problem; right: the normal character)

I don't know if I chose the best approach. I'm willing to change that and also give more details about the problem.

Ref https://github.com/TheAlgorithms/Go/issues/480

### DIFF
--- a/cipher/caesar/caesar.go
+++ b/cipher/caesar/caesar.go
@@ -9,12 +9,12 @@ func Encrypt(input string, key int) string {
 	key8 := byte(key%26+26) % 26
 
 	var outputBuffer []byte
-	// r is a rune, which is the equivalent of uint32.
-	for _, r := range input {
-		newByte := byte(r)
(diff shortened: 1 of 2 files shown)


tjgurwara99 (approved):
Indeed, this is a correct change.

VictorAssunc:
> Indeed, this is a correct change.

ok, thanks. it would be nice if you can label this PR with `hacktoberfest-accepted` :)

raklaptudirm (approved):

Issue #531, opened 2022-08-31
LFU cache implementation


itsAkshayDubey:
Hi,

I have not seen any implementation of [least frequently used cache](https://en.m.wikipedia.org/wiki/Least_frequently_used) in this repo.
Could you please assign this issue to me?

Regards,
Akshay

sandyydk:
@raklaptudirm Can I pick this up?

raklaptudirm:
@sandyydk Sure.

raklaptudirm:
Just open a pr and we will review it

PR #561, opened 2022-10-09
lfu cache implementation using priority queue

the-pinbo:
I've added the implemented LFU cache using priority queue, and also written a test file which covers most of the edge cases and some additional functions to the linked list package 
#531

### DIFF
--- a/cache/lfu/lfu.go
+++ b/cache/lfu/lfu.go
@@ -1,0 +1,136 @@
+// Package lfu (Least Frequently Used) is a type of cache algorithm used to manage memory within a computer.
+// ref: https://en.m.wikipedia.org/wiki/Least_frequently_used
+
+package lfu
+
+import (
+	"container/heap"
+	"errors"
+)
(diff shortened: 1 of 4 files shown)


cache/lfu/lfu.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	}
+
+	item.counter++
+	item.value = value
+	item.timeStamp = c.timeStamp
+	c.pq.Update(item)
+}
+
+type PriorityQueue []*Item
```
tjgurwara99, line 91: This doesn't belong in this package. This is a general structure that should be in the structure package not here.

tjgurwara99 (changes requested):

cache/lfu/lfu.go
the-pinbo: Go repo does not have an implementation of priority queue.I'll first make a PR for that in structure package.

PR #568, opened 2022-10-13
algorithm: distance between two points

Chetan07j:
### Describe your change:

Euclidean plane: Find distance between points

- [X] Add an algorithm?
- [ ] Fix a bug or typo in an existing algorithm?
- [ ] Documentation change?

### Checklist:

- [X] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Go/blob/master/CONTRIBUTING.md).
- [X] This pull request is all my own work -- I have not plagiarized.
- [X] I know that pull requests will not be merged if they fail the automated tests.
- [X] This PR only changes one algorithm file.
- [X] All new Go files are placed inside an existing directory.
- [X] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.


raklaptudirm (approved):

math/geometry/distancebetweenpoints.go
```diff
@@ -0,0 +1,32 @@
...
+package geometry
+
+import (
+	"math"
+)
+
+// This function calculates the distance based on Cartesian Plane Distance Formula
+// Formula: √((x1-x2) * (x1-x2) + (y1-y2) * (y1-y2))
+// Takes x and y coordinates of two point
+// Returns distnce in floating number
```
raklaptudirm, line 19:
```suggestion
// Returns distance in floating number
```

raklaptudirm (changes requested):

math/geometry/distancebetweenpoints.go
```diff
@@ -0,0 +1,32 @@
...
+
+	// Find difference between y-coordinates
+	yDiff := y2 - y1
+
+	// Apply formula
+	distance := (xDiff * xDiff) + (yDiff * yDiff)
+
+	// Return square root
+	return math.Sqrt(distance)
+}
```
tjgurwara99, line 32: This should be more general IMO, making this general is relatively straight forward.

I think the following should work

```suggestion
type Point []float64

var ErrDimMismatch = errors.New("mismatched dimensions")

func EuclideanDistance(p1 Point, p2 Point) (float64, error) {
	n := len(p1)
	if len(p2) != n {
		return -1, ErrDimMismatch
	}
	diff_squared := make([]float64, n)
	for i, x_i := range p1 {
		diff := math.Abs(x_i - p2[i])
		diff_squared[i] = diff * diff
	}
	var total float64 = 0
	for _, val := range diff_squared {
		total += val
	}
	return math.Sqrt(total), nil
}
```

tjgurwara99 (changes requested):

math/geometry/distancebetweenpoints.go
Chetan07j: @tjgurwara99 Thanks for your suggestion.
I have updated the code as per your suggestion but with slight modifications, could you please have a look at it again?

math/geometry/distancebetweenpoints.go
```diff
@@ -0,0 +1,53 @@
+// distancebetweenpoints.go
+// Find distance between two points
+// Details:
+// Given coordinates of 2 points on a cartesian plane, find the distance between them rounded up to nearest integer.
+// Link: https://practice.geeksforgeeks.org/problems/distance-between-2-points3200/1?page=1&category[]=geometric
+
+// author(s) [Chetan Patil](https://github.com/Chetan07j)
```
tjgurwara99, line 7:
```suggestion
// distancebetweenpoints.go
// Find Euclidean distance between two points
// author(s) [Chetan Patil](https://github.com/Chetan07j)
```

The link referred doesn't apply anymore since the algorithm is widely different from the initial one you proposed.

I also recommend changing the file name to `distance.go` or `euclideandistance.go`. I prefer the former.

math/geometry/distancebetweenpoints.go
```diff
@@ -0,0 +1,53 @@
...
+// Package geometry contains geometric algorithms
+package geometry
+
+import (
+	"errors"
+	"math"
+)
+
+// Point defines a point with x and y coordinates.
+type CartesianPoint []float64
```
tjgurwara99, line 18:
```suggestion
type Point []float64
```

Cartesian points are only two dimensional. This is `n` dimensional.

tjgurwara99 (changes requested):

math/geometry/distancebetweenpoints.go
Chetan07j: type `Point` is already defined in package geometry in another file, so I could not name it the same here.

math/geometry/distancebetweenpoints.go
Chetan07j: done

math/geometry/distancebetweenpoints.go
```diff
@@ -0,0 +1,53 @@
...
+		// Find difference
+		diff := math.Abs(x_i - p2[i])
+
+		// Add diff * diff to total value
+		total += diff * diff
+	}
+
+	// Last step of formula
+	// return square root of total
+	return math.Sqrt(total), nil
```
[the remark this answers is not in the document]
Chetan07j: Updated it. Thanks for valuable feedback

math/geometry/distancebetweenpoints.go
```diff
@@ -0,0 +1,53 @@
...
+
+var ErrDimMismatch = errors.New("mismatched dimensions")
+
+// This function calculates the distance based on Cartesian Plane Distance Formula
+// Formula: √((x1-x2) * (x1-x2) + (y1-y2) * (y1-y2))
+// Takes x and y coordinates of two point
+// Returns distance in floating number
+func FindDistanceBetweenTwoPoints(p1 CartesianPoint, p2 CartesianPoint) (float64, error) {
+	// Find length of point 1
+	p1Length := len(p1)
```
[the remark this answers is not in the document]
Chetan07j: Done.

math/geometry/distancebetweenpoints.go
tjgurwara99: Sure. Though I was thinking maybe the other one could be called `CartesianPoint` or `Coordinate`.

Chetan07j: Ok. Done made it as `EuclideanPoint`

Could you please review PR again?

math/geometry/distance.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Package geometry contains geometric algorithms
+package geometry
+
+import (
+	"errors"
+	"math"
+)
+
+// Point defines a point with x and y coordinates.
```
tjgurwara99, line 13:
```suggestion
// EuclideanPoint defines a point with x and y coordinates.
```

Whenever refactoring, it is also a good idea to refactor all the comments associated with them. Otherwise the comments and code go out of sync.

tjgurwara99 (changes requested):

math/geometry/distance.go
Chetan07j: Done @tjgurwara99

math/geometry/distance_test.go
```diff
@@ -0,0 +1,104 @@
...
+	p1 EuclideanPoint
+	p2 EuclideanPoint
+}
+
+func TestFindDistanceBetweenTwoPoints(t *testing.T) {
+	tests := []struct {
+		name    string
+		args    args
+		want    float64
+		wantErr bool
```
tjgurwara99, line 17: None of your tests are using this. This method is used to test whether you get the error you expect. Please write an additional case to test for `ErrDimMismatch`.

math/geometry/distance_test.go
```diff
@@ -0,0 +1,104 @@
...
+		{
+			"(2,2) and (14,11)",
+			args{
+				EuclideanPoint{2, 2},
+				EuclideanPoint{14, 11},
+			},
+			15,
+			false,
+		},
+	}
```
tjgurwara99, line 95: You don't need all these test cases. Pick one and run the benchmark loop on that case only. The benchmark is giving an answer to the question "how long does it take for this algorithm to give answer to three test cases" not "how long does it take to run on one case".

tjgurwara99 (changes requested):

math/geometry/distance_test.go
Chetan07j: Noted and done.

math/geometry/distance_test.go
Chetan07j: Noted & done.

math/geometry/distance_test.go
```diff
@@ -0,0 +1,77 @@
...
+				EuclideanPoint{},
+			},
+			-1,
+			true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := EuclideanDistance(tt.args.p1, tt.args.p2)
+			if (err != nil) != tt.wantErr {
```
tjgurwara99, line 59:
```suggestion
			if (err != nil) != tt.wantErr && errors.Is(err, geometry.ErrDimMismatch)({
```

You're still not checking the error 😄

math/geometry/distance_test.go
```diff
@@ -0,0 +1,77 @@
+package geometry
```
tjgurwara99, line 1:
```suggestion
package geometry_test
```

This would mean you won't be able to call your function directly. You would need to import the `geometry` package like this


```Go
import "github.com/TheAlgorithms/Go/math/geometry
```

and calling things would require you to prefix it like this

```Go
geometry.EuclideanPoint{}
```
etc.

tjgurwara99 (changes requested):

tjgurwara99 (approved):

raklaptudirm (approved):

PR #571, opened 2022-10-14
LFU cache implementation using linkedhashmap

the-pinbo:
I've added the implemented LFU cache using linkedhashmap ,and also written a test file which covers most of the edge cases with documentation. 
Took a lot of effort to write this, hope my PR gets merged this time🤞🏽.
#531
#561


cache/lfu.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"fmt"
+
+	"github.com/TheAlgorithms/Go/structure/linkedlist"
+)
+
+// LFUCache is a struct that holds the LFU cache
+type LFUCache struct {
+	lookUp       map[any]*linkedlist.Node[any]
+	frequency    map[any]*linkedlist.Doubly[any]
```
tjgurwara99, line 14: You're using `any` here without an actual benefit since you case everything in your implementations. I would suggest using `LFUNode` directly here...

```suggestion
	lookUp       map[any]*linkedlist.Node[LFUNode]
	frequency    map[any]*linkedlist.Doubly[LFUNode]
```

cache/lfu.go
```diff
@@ -0,0 +1,112 @@
...
+		cache.lookUp[key] = cache.frequency[cnt].Front()
+	}
+}
+
+// Evict removes the least frequently used key
+func (cache *LFUCache) Evict() {
+	evictElement := cache.frequency[cache.minFrequency].Back()
+	delete(cache.lookUp, evictElement.Val.(LFUNode).key)
+	cache.frequency[cache.minFrequency].Remove(evictElement)
+}
```
tjgurwara99, line 112: I really like your implementation so far 😄 Good job getting it this far. There are some things I wanted to try out, so I checkout out your branch and did somethings inside it. This is the code now, as you can see that there is no type casting in the following


```suggestion
// Package lfu (Least Frequently Used) is a type of cache algorithm used to manage memory within a computer.
// ref: https://en.m.wikipedia.org/wiki/Least_frequently_used
package cache

import (
	"fmt"

	"github.com/TheAlgorithms/Go/structure/linkedlist"
)

// LFU is a struct that holds the LFU cache
type LFU[T comparable] struct {
	lookUp       map[T]*linkedlist.Node[Node[T]]
	frequency    map[int]*linkedlist.Doubly[Node[T]]
	minFrequency int
	size         int
}

// Node is a node for LFU cache
type Node[T comparable] struct {
	key   T
	value any
	count int
}

// NewLFU initializes the object with the capacity of the data structure.
func NewLFU[T comparable](capacity int) LFU[T] {
	cache := LFU[T]{}
	cache.lookUp = map[T]*linkedlist.Node[Node[T]]{}
	cache.frequency = map[int]*linkedlist.Doubly[Node[T]]{}
	cache.size = capacity
	cache.minFrequency = -1
	return cache
}

// Touch updates the frequency of the key and moves it to the front
// of the list of the new frequency bucket
func (cache *LFU[T]) Touch(element *linkedlist.Node[Node[T]]) {
	key, val := element.Val.key, element.Val.value
	cnt := element.Val.count
	node := Node[T]{key, val, cnt + 1}
	cache.frequency[cnt].Remove(element)
	if cache.minFrequency == cnt && cache.frequency[cnt].Count() == 0 {
		cache.minFrequency += 1
	}
	if _, found := cache.frequency[cnt+1]; !found {
		cache.frequency[cnt+1] = linkedlist.NewDoubly[Node[T]]()
	}
	cache.frequency[cnt+1].AddAtBeg(node)
	cache.lookUp[key] = cache.frequency[cnt+1].Front()
}

// Print prints the LFU cache
func (cache *LFU[T]) Print() {
	fmt.Println("print map")
	for _, v := range cache.lookUp {
		node := v.Val
		fmt.Println(node.key, node.value, node.count)
	}
	fmt.Println("end")
}

// Get gets the value of the key if the key exists in the cache.
// Otherwise, throws an error
func (cache *LFU[T]) Get(key T) (any, error) {
	if e, found := cache.lookUp[key]; found {
		cache.Touch(e)
		return e.Val.value, nil
	} else {
		return nil, fmt.Errorf("key not found")
	}
}

// Put updates the value of the key if present,
// or inserts the key if not already present.
// When the cache reaches its capacity,
// it should invalidate and remove the least
// frequently used key before inserting a new item.
// In this implementation, when there is a tie
// (i.e., two or more keys with the same frequency),
// the least recently used key would be invalidated.
func (cache *LFU[T]) Put(key T, value any) {
	if cache.size <= 0 {
		return
	}

	if e, found := cache.lookUp[key]; found {
		node := e.Val
		node.value = value
		e.Val = node
		cache.Touch(e)
	} else {
		if len(cache.lookUp) == cache.size {
			cache.Evict()
		}
		cnt := 1
		cache.minFrequency = 1
		node := Node[T]{key: key, value: value, count: cnt}
		if _, found := cache.frequency[cnt]; !found {
			cache.frequency[cnt] = linkedlist.NewDoubly[Node[T]]()
		}
		cache.frequency[cnt].AddAtBeg(node)
		cache.lookUp[key] = cache.frequency[cnt].Front()
	}
}

// Evict removes the least frequently used key
func (cache *LFU[T]) Evict() {
	evictElement := cache.frequency[cache.minFrequency].Back()
	delete(cache.lookUp, evictElement.Val.key)
	cache.frequency[cache.minFrequency].Remove(evictElement)
}
```

Its not a lot of change but I think its quite meaningful here in my opinion.

cache/lfu_test.go
```diff
@@ -0,0 +1,178 @@
+package cache
```
tjgurwara99, line 1: For testing, I really recommend you use table testing rather than multiple tests such as `TestLFU1` etc.

Also, really think you should use black box testing in this case. You may have missed something trivial that is not discoverable locally. Use
```suggestion
package cache_test
```

tjgurwara99 (commented):
Really great work. LFU is notoriously difficult (at least to me) so I really appreciate what you've done. I've posted some suggested changes in a (very) big block since there are certain things that are better off without type casting.

Consider the following suggested changes.

PR #572, opened 2022-10-14
Added the implementation of heap and priority queue

the-pinbo:
I have added the implementation of heap and priority queue and one test case which covers most of the edge cases along with detailed documentation.

### DIFF
--- a/structure/priorityqueue/heap.go
+++ b/structure/priorityqueue/heap.go
@@ -1,0 +1,110 @@
+// Heap data structure
+// description: based on `geeksforgeeks` heap is a specialized tree-based data structure which is essentially an almost complete tree that satisfies the heap property.
+//
+//	In a heap, the highest (or lowest) priority element is always stored at the root.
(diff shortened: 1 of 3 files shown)


tjgurwara99 (changes requested):
Please read our contribution guidelines. This code is almost identical to the `heap` implementation from the standard library without any proper accreditation to the original authors, as such I consider this plagiarised. @siriak @raklaptudirm care to comment?

raklaptudirm:
I would agree with the assessment. This pull request has been plagiarised from <samp>container/head</span>, including the priority queue implementation.

the-pinbo:
I am really sorry for this, I didn't mean to do this intentionally. I had made a PR on implementation of LFU cache using priority queues #561 , which uses the container/heap.It mentioned that it uses a bsd-style license which allows proprietary use and allows the software released under the license to be incorporated into proprietary products, so I referenced code from container/heap so that I would not have to change much of my code. I am really sorry for the inconvenience 🙇.

PR #574, opened 2022-10-16
Add Disjoint Set Union

Mohammadreza-mz:
I've added 3 different implementations for DSU
dsu comparison by size,
dsu comparison by rank with path compression,
dsu by iterating on the smaller set and changing their parents.
I've also defined the DSU interface to test these three implementations with the same tests.


tjgurwara99:
There is already an implementation of DSU in the repository. If you're creating an interface, please make sure that the other implementation also follows it.

PR #567, opened 2022-10-13
refactor: simplity signature of `LongestCommonSubsequence`

vil02:
#### Description of Change

In this PR I remove the _length of the strings_ arguments in `LongestCommonSubsequence`, update the relevant tests and add a test case, which justifies the usage of `utf8.RuneCountInString` instead of `len`.

I think that having the length of the input strings in `LongestCommonSubsequence` is a bit redundant. Also by looking at implementations in other languages like:
- [python](https://github.com/TheAlgorithms/Python/blob/6118b05f0efd1c2839eb8bc4de36723af1fcc364/dynamic_programming/longest_common_subsequence.py#L9),
- [rust](https://github.com/TheAlgorithms/Rust/blob/5bb80cdfbe8092cfaf111f06a8abe9f205849856/src/dynamic_programming/longest_common_subsequence.rs#L5),
- [java](https://github.com/TheAlgorithms/Java/blob/c805437c0c5e057f3ad60c397fe6550eec7be453/src/main/java/com/thealgorithms/dynamicprogramming/LongestCommonSubsequence.java#L5),

I am convinced that this is a good direction.

#### Checklist

- [x] Added description of change
- [x] Added tests and example, test must pass
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/Go/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes: 
This PR shortens the argument list of the function `LongestCommonSubsequence`.

PR #578, opened 2022-10-17
fix: handle multibyte chars propertly in `LongestCommonSubsequence`

vil02:
#### Description of Change

in PR #567 I have introduced some UTF-8 related bug. This PR hopefully fixes it. The bug relates to the problem of accessing the character at given index in a string in `Go` (cf. [stackoverflow](https://stackoverflow.com/questions/15018545/how-to-index-characters-in-a-golang-string)).

#### Checklist

Notes: 
This PR fixes some UTF-8 related bug in [`LongestCommonSubsequence`](https://github.com/TheAlgorithms/Go/blob/d61cee077bf76f9bc5c16ee4eb88385d1796b03c/dynamic/longestcommonsubsequence.go#L12).


tjgurwara99 (approved):
LGTM. @raklaptudirm Seems like the last commit the PR author made to this algorithm had some issues that I couldn't identify. Could you review this properly please? 🙏🏼 Your understanding of utf8 surpasses mine, so it would be helpful to have your eyes on this 😄

raklaptudirm:
@tjgurwara99 I didn't understand what you meant, the checks seem to be passing, could you elaborate on the symptoms?

tjgurwara99:
I meant that this algorithm was recently changed by the same author in previous PR's and I didn't notice the mistake at that time. I thought maybe you might be able to review it as well to ensure I didn't miss anything 😄

raklaptudirm:
@tjgurwara99 could you provide the pr number?

dynamic/longestcommonsubsequence.go
```diff
@@ -8,10 +8,14 @@ import (
 	"unicode/utf8"
 )
 
+func extractData(inStr string) (int, []rune) {
```
raklaptudirm, line 11: This function seems unnecessary and obfuscates the code. The rune count should equal the length of the rune array.

raklaptudirm (changes requested):

dynamic/longestcommonsubsequence.go
vil02: > The rune count should equal the length of the rune array.

Please have a look [here](https://go.dev/play/p/JKbfU7z0o4o).

The problem is that if you have `someString` (of the type `string`), then with `someString[i]` you access the byte at the position `i`.

raklaptudirm: I understand that, however, you are converting it to a `[]rune` so it should not be a problem.

vil02: Exactly. That is the reason to have the function `extractData` - it is not unnecessary.

raklaptudirm: You can just do:
```go
A := []rune(a)
aLen := len(A)
```
I don't see how that function is useful or more intuitive.

vil02: The same has to be done for `b`. The function `extractData` helps to avoid code duplication.

raklaptudirm: This duplication is not so severe that we must sacrifice readability and intuitiveness for it.

tjgurwara99: I agree with @raklaptudirm

vil02: Having even a _trivial_ duplication like this 
```golang
A := []rune(a)
aLen := len(A)
B := []rune(b)
bLen := len(B)
```
is a bad idea because of (at least) two reasons:
- possibility of having a typo,
- these lines are not _testable_. One can write tests for the function `extractData`, but it is not possible to test the lines above separately.

In the discussed case, we really want to do the same thing for the strings `a` and `b`: we want to compute their length and cast it to `[]rune`. It is not just a coincidence, that the _code looks the same_. We want to do the exact same thing. That motivates the introduction of this function - having it underlines that _we really do the same thing_. Sometimes it happens that the code _looks the same_ but the purpose of it is slightly different - abstracting it into a single function would be _bad_.

@raklaptudirm I have to admit that I like your idea of getting rid of `utf8.RuneCountInString` while  computing the number of _characters_.

After thinking about it, I think that the best way would be to change `LongestCommonSubsequence` to accept `[]rune`, i.e. to have `LongestCommonSubsequence([]rune a, []rune b) int`. Then it is clear just from looking at the signature that this function only _solves the problem_ and how does it handle the _multibyte_ characters. @tjgurwara99 @raklaptudirm please let me know what do you think about that.

tjgurwara99: The problem I have with your function `extractData` is that it's not really a great name for a function and the order of `return` types. I argue that it's really not conveying what it does when reading the code. Questions like "what does it extract?" or "why does it extract?" pop up as soon as I read it which means before I can read further I have to step into the code and read that code causing me to switch context a few more times than I need to. The simplest approach is quite straight forward imo as @raklaptudirm mentioned - but if you want to use a function, calling it something self explanatory is better than calling it `extractData`. Something like `strToRuneSlice(s string) (r []rune, size int)` is much clearer to me at least.

> these lines are not testable. One can write tests for the function `extractData`, but it is not possible to test the lines above separately.

Why would you want to test something like that separately?

> possibility of having a typo,

That possibility exists everywhere - why do you think PR reviews exist? 😂 I can do this

```Go
extractData(a)
extratcData(b)
```

The compiler would complain in the above case and any other type that you may want to think of in this context.

raklaptudirm: Go's type system almost completely removes the possibility of typos going undetected, because either the symbol doesn't exist or it's type is different. One of the two cases is true a majority of the time.

vil02: @tjgurwara99 thank you for the idea for better name, I really like the `strToRuneSlice`.

> That possibility exists everywhere - why do you think PR reviews exist?

The code review is about the things, which cannot be verified by automated tools. I was more thinking about the typo like:
```golang
A := []rune(b)
aLen := len(A)
B := []rune(a)
bLen := len(B)
```
Above would not be detected by the compiler or other tools - all of the variables are used etc. but of course the _logic_ behind the names is broken. Note that the result of `LongestCommonSubsequence` will remain correct but the potential reader of the code would be a bit confused.


> Why would you want to test something like that separately?

In this _simple case_, I would not. But is is a good practice to write _testable_ code.


@raklaptudirm @tjgurwara99 I think I can mark this as resolved.

raklaptudirm: I still don't think sacrificing readability is a good idea. Typos can happen everywhere as previously stated.

vil02: I do not understand why introducing this function reduces the readability. It is like:
```golang
aRunes, aLen := strToRuneSlice(a)
bRunes, bLen := strToRuneSlice(b)
```
vs.
```golang
aRunes := []rune(a)
aLen := len(A)
bRunes := []rune(b)
bLen := len(B)
```
I think that a person, who does know the `Go` syntax is more likely to have less problems with reading the first one. In the first case there is this extra bit of information that _we really intend to do the exact same thing to `a` and `b`_. Moreover the first one is simply twice shorter.

tjgurwara99: I think if you use the `strToRuneSlice`, I don't mind personally.

> I think that a person, who does know the Go syntax is more likely to have less problems with reading the first one.

That is under the assumption that this is being **read** by a person who does indeed know Go, and its syntax and terminology (eg `Rune` and `Slice`). But this repository is for learning purposes, so you can see why that `strToRuneSlice` would require the person not familiar with Go to step into the code and see (albeit its in the same file so I personally don't mind) but still... 😄

tjgurwara99 (approved):

dynamic/longestcommonsubsequence.go
vil02: @raklaptudirm Can I mark this conversation as resolved?

raklaptudirm (approved):

PR #585, opened 2022-10-21
refactor: fmt.Printf in strings/kmp

uh-zz:
## Description
Hi!
This is my first pull request to this repository!

Refactored `fmt.Printf` so that it is done in each block, and do not insert a newline code at the beginning of `fmt.Printf`.

I'll also include the output below.

```
go test -v
=== RUN   TestKMP
=== RUN   TestKMP/String_comparison_on_single_pattern_match
        comparing characters C a at positions 0 0
        comparing characters P a at positions 1 0
        comparing characters M a at positions 2 0
        comparing characters _ a at positions 3 0
        comparing characters a a at positions 4 0 - match
        comparing characters n n at positions 5 1 - match
        comparing characters n n at positions 6 2 - match
        comparing characters u o at positions 7 3
        comparing characters u a at positions 7 0
        comparing characters a a at positions 8 0 - match
        comparing characters l n at positions 9 1
        comparing characters l a at positions 9 0
        comparing characters _ a at positions 10 0
        comparing characters c a at positions 11 0
        comparing characters o a at positions 12 0
        comparing characters n a at positions 13 0
        comparing characters f a at positions 14 0
        comparing characters e a at positions 15 0
        comparing characters r a at positions 16 0
        comparing characters e a at positions 17 0
        comparing characters n a at positions 18 0
        comparing characters c a at positions 19 0
        comparing characters e a at positions 20 0
        comparing characters _ a at positions 21 0
        comparing characters a a at positions 22 0 - match
        comparing characters n n at positions 23 1 - match
        comparing characters n n at positions 24 2 - match
        comparing characters o o at positions 25 3 - match
        comparing characters u u at positions 26 4 - match
        comparing characters n n at positions 27 5 - match
        comparing characters c c at positions 28 6 - match
        comparing characters e e at positions 29 7 - match
=== RUN   TestKMP/String_comparison_on_multiple_pattern_match
        comparing characters A A at positions 0 0 - match
        comparing characters A A at positions 1 1 - match
        comparing characters B B at positions 2 2 - match
        comparing characters A A at positions 3 3 - match
=== RUN   TestKMP/String_comparison_with_not_found_pattern
        comparing characters A A at positions 0 0 - match
        comparing characters A A at positions 1 1 - match
        comparing characters B B at positions 2 2 - match
        comparing characters A C at positions 3 3
        comparing characters A A at positions 3 0 - match
        comparing characters A A at positions 4 1 - match
        comparing characters C B at positions 5 2
[...]

PR #586, opened 2022-10-22
Add: kmp(Knuth-Morris-Pratt) algorithm

uh-zz:
### What I have implemented:
The kmp algorithm was already created, but I've implemented it differently, so I'm going to propose it in this pull request!

### Relates
https://github.com/TheAlgorithms/Go/pull/585


strings/kmp/kmp.go
```diff
@@ -1,117 +1,49 @@
...
-			cnd++
-			t[pos] = cnd
-			pos++
-		} else if cnd > 0 {
-			cnd = t[cnd]
-		} else {
-			t[pos] = 0
-			pos++
+// Table building for kmp algorithm.
+func Table(w string) []int {
```
tjgurwara99, line 36: I think this shouldn't be exported. Is there a reason why this is exported?

strings/kmp/kmp.go
```diff
@@ -1,117 +1,49 @@
...
+	var (
+		t []int = []int{-1}
+		k int
+	)
+	for j := 1; j < len(w); j++ {
+		k = j - 1
+		for w[0:k] != w[j-k:j] && k > 0 {
```
tjgurwara99, line 43: This feels weird `w[0:k] != w[j-k:j]`

What you're checking is simply, `j == k` isn't it? Why involve pointers when you could just as easily solve the problem without any pointer checks?

tjgurwara99 (changes requested):

strings/kmp/kmp.go
uh-zz: @tjgurwara99 
Thanks for review!

You are right, I thought it did not need to be made public too, so I changed it to private!
[78ac44c](https://github.com/TheAlgorithms/Go/pull/586/commits/78ac44cafb45b89d7436fca7a65abe18b297d95c)

strings/kmp/kmp.go
uh-zz: @tjgurwara99 
Thanks for review.

> w[0:k] ! = w[j-k:j]!

The assumption is that  calculating the number of consecutive matching characters in the string `w`.

The reason for this condition is to compare a string of length k from the beginning of w with a string of length k from the end of `w`.

If they do not match, it subtracts 1 from k and repeats until they match or k=0.

e.g. ) The process flow I envision.

```
w="teste"

// 1. 
j = 5
k = 4
w[0:k] = "test"
w[j-k:j] = "este"

The above does not match, so k = k-1

// 2.
j = 5
k = 3
w[0:k] = "tes"
w[j-k:j] = "ste"

The above does not match, so k = k-1

// 3.
j = 5
k = 2
w[0:k] = "te"
w[j-k:j] = "te"

Match found, so exit from loop as k = 2
```

If I am wrong, or if you find this process complicated, I would appreciate any advice you can give me!

tjgurwara99: Ok, I understand. The following is non-blocking and you can mark this resolved: but I argue that since the condition is part of the core logic not something that should just sit loop conditional (kind of makes it obscure what is happening and difficult to reason about.)

I suggest something like this.

```suggestion
		for k > 0 {
			if w[0:k] != w[j-k:j] {
				break
			}
```

notice that there is no difference but it shows that if condition is the central part of the logic

uh-zz: @tjgurwara99 
Thanks for the suggestion.:+1:

This logic seems to accomplish what I was hoping to do.!
But you are right, I think that it difficult to deduce from this logic too.

I thought it might be a good idea to change the logic a little as follows.
I would appreciate your feedback.

```go
// e.g. word = "tested"
// expected: [-1 0 0 1 2 0]
func table(word string) (t []int) {
	t = make([]int, len(word))
	var maxMatchLength int

	t[0] = -1

	for index := 1; index < len(t); index++ {
		// step 2.1: index = 4 is w[4] = "e", maxMatchLength = 1 is w[1] = "e"
		// -> will not get into loop
		//
		// step 3.1: index = 5 is w[5] = "d", maxMatchLength = 2 is w[2] = "s"
		// -> maxMatchLength = 1 is w[1] = "t"
		for maxMatchLength > 0 && word[index] != word[maxMatchLength] {
			maxMatchLength = t[maxMatchLength-1]
		}

		// step 1: index = 3 is w[3] = "t", maxMatchLength = 0 is w[0] = "t"
		// -> maxMatchLength = 1
		//
		// step 2.2: index = 4 is w[4] = "e", maxMatchLength = 1 is w[1] = "e"
		// -> maxMatchLength = 2
		//
		// step 3.2: index = 5 is w[5] = "d", maxMatchLength = 1 is w[1] = "t"
		// -> not match
		if word[index] == word[maxMatchLength] {
			maxMatchLength++
		}
		t[index] = maxMatchLength
	}
	return t
}
```

tjgurwara99: The above still doesn't really help the reader in anyway. I think you should stick to the earlier version (and maybe accept the above recommendation since that is much easier to follow) but this was a non-blocking comment, so I'm approving it nonetheless...

tjgurwara99 (approved):

raklaptudirm (approved):

PR #587, opened 2022-10-22
feat: add sets algorithms for proper subset and proper superset check

kalyan-arepalle:
The sets structure already has functions to check whether a set is subset or superset of another, added functions to check if a set is proper subset or proper superset of another.


structure/set/set.go
```diff
@@ -28,8 +28,16 @@ type Set interface {
 	In(item any) bool
 	// IsSubsetOf: checks whether set is subset of set2 or not.
 	IsSubsetOf(set2 Set) bool
+	// IsProperSubsetOf: checks whether set is proper subset of set2 or not.
+	// ex: [1,2,3] proper subset of [1,2,3,4] -> true
+	// ex: [1,2,3] proper subset of [2,3,1] -> false
```
tjgurwara99, line 33: I don't think you need an example here

structure/set/set.go
```diff
@@ -28,8 +28,16 @@ type Set interface {
...
+	IsProperSubsetOf(set2 Set) bool
 	// IsSupersetOf: checks whether set is superset of set2 or not.
 	IsSupersetOf(set2 Set) bool
+	// IsProperSupersetOf: checks whether set is proper superset of set2 or not.
+	// ex: [1,2,3,4] proper superset of [1,2,3] -> true
+	// ex: [1,2,3] proper superset of [2,3,1] -> false
```
tjgurwara99, line 39: Again, I don't think you need this.

structure/set/set.go
```diff
@@ -88,10 +96,18 @@ func (st *set) IsSubsetOf(superSet Set) bool {
 	return true
 }
 
+func (st *set) IsProperSubsetOf(superSet Set) bool {
+	return st.IsSubsetOf(superSet) && st.Len() != superSet.Len()
```
tjgurwara99, line 100: I think you should flip the condition. The first thing to check for proper subset is size usually and its more readable. (I think it might even be more efficient as well though readability takes precedence in my mind)

```suggestion
	if st.Len() == superSet.Len() {
		return false
	}
	return st.IsSubsetOf(superSet)
```

structure/set/set.go
```diff
@@ -88,10 +96,18 @@ func (st *set) IsSubsetOf(superSet Set) bool {
...
+}
+
 func (st *set) IsSupersetOf(subSet Set) bool {
 	return subSet.IsSubsetOf(st)
 }
 
+func (st *set) IsProperSupersetOf(subSet Set) bool {
+	return st.IsSupersetOf(subSet) && st.Len() != subSet.Len()
```
tjgurwara99, line 108:
```suggestion
	if st.Len() == subSet.Len() {
		return false
	}
	return st.IsSupersetOf(subSet)
```

tjgurwara99 (changes requested):

tjgurwara99 (approved):

raklaptudirm (approved):

PR #588, opened 2022-10-22
feat(cipher/rot13): add fuzz test to rot13 cipher

mcaci:
Hi, in the context of #480 I'm adding quickly the fuzz tests for the rot13 algorithm. I tested it with `go test ./cipher/rot13/` to test the corpus and `go test -fuzz=FuzzRot13 ./cipher/rot13/` to run the fuzzer itself.
Is that ok for you to add it?
If you have any feedbacks about my change let me know so that I can adapt it


cipher/rot13/rot13_test.go
```diff
@@ -64,7 +64,19 @@ func TestRot13Decrypt(t *testing.T) {
 func assertRot13Output(t *testing.T, input, expected string) {
 	actual := rot13(input)
 	if actual != expected {
-		t.Fatalf("With input string '%s' was expecting '%s' but actual was '%s'",
+		t.Fatalf("With input string %q was expecting %q but actual was %q",
```
tjgurwara99, line 67: Nice! I like the quote formatter - I recently learnt about them.

tjgurwara99 (approved):

cipher/rot13/rot13_test.go
mcaci: Eheh, yes I saw it and I said it was a good opportunity to change it with %q. It's actually a great formatting verb for strings in my opinion

siriak (approved):
Looks good, thanks!

PR #589, opened 2022-10-23
fix: Tree package bug

guuzaa:
1. Implement `Successor` and `Predecessor` methods for all trees (AVL, RB and BinarySearch). 
2. Modify the bug of `NIL` and make `NIL` a sentinel for all trees .
3. Suppose all keys in trees are different. If we push two identical keys, it has only one key inserted.
4. Now, each tree has the following methods.
```go
type Tree[T constraints.Ordered] interface {
	Push(...T)
	Delete(T) bool
	Get(T) (*Node[T], bool)
	Has(T) bool
	Depth() int
	Max() (T, bool)
	Min() (T, bool)
	Predecessor(T) (T, bool)
	Successor(T) (T, bool)
	PreOrder() []T
	InOrder() []T
	PostOrder() []T
	LevelOrder() []T
	AccessNodesByLayer() [][]T
	Print()
}
```


structure/tree/avl.go
```diff
@@ -11,10 +11,11 @@ type AVL[T constraints.Ordered] struct {
 
 // NewAVL create a novel AVL tree
 func NewAVL[T constraints.Ordered]() *AVL[T] {
+	var leaf *Node[T]
```
tjgurwara99, line 14: I don't think this was a problem to begin with. I'm curious to know why you think this was a bug. Then I would present my case for why I don't think (at the moment - might change my mind after seeing your argument) 😄

guuzaa: You are right. For AVL and BinarySearch `this.NIL == nil`. There is no need to introduce a variable `leaf` which is equivalent to `nil` .

guuzaa: Actually, the bug exists in `RB`. I didn't make the parent of sentinel value NIL pointing to NIL. As a result, I had to introduce a lot of nil's, which is misleading. 
In this commit, I replace all nil to `t.NIL`(In BinarySearch and AVL, it's equal to nil. In RB, it's equal to `leaf`, a sentinel value).

```go
// NewBinarySearch create a novel Binary-Search tree
func NewBinarySearch[T constraints.Ordered]() *BinarySearch[T] {
	return &BinarySearch[T]{
		binaryTree: &binaryTree[T]{
			Root: nil,
			NIL:  nil,
		},
	}
}

// NewAVL create a novel AVL tree
func NewAVL[T constraints.Ordered]() *AVL[T] {
	return &AVL[T]{
		binaryTree: &binaryTree[T]{
			Root: nil,
			NIL:  nil,
		},
	}
}

// Create a new Red-Black Tree
func NewRB[T constraints.Ordered]() *RB[T] {
	leaf := &Node[T]{Color: Black, Left: nil, Right: nil}
	leaf.Parent = leaf  // the parent of NIL is pointing to NIL now
	return &RB[T]{
		binaryTree: &binaryTree[T]{
			Root: leaf,
			NIL:  leaf,
		},
	}
}

```

tjgurwara99: Doesn't this mean that the bug where it exists should be solved instead of changing everything that is dependent on a base that is shared between different implementations. What you're currently doing is that you're fixing a bug in one implementation (RB) that are then causing issues in other implementations elsewhere. In real productions systems, this should be avoided at all costs otherwise this can easily become an issue (in this case its a small issue but trust me, it can create further issues in big systems).

structure/tree/avl.go
```diff
@@ -37,18 +37,25 @@ func (avl *AVL[T]) Delete(key T) bool {
 }
 
 func (avl *AVL[T]) pushHelper(root *Node[T], key T) *Node[T] {
-	if root == nil {
+	if avl.isNil(root) {
```
tjgurwara99, line 40: this should be the same as before since this is not the implementation that was causing the issue.

tjgurwara99 (changes requested):
Could you remove all instances of `isNil`, it doesn't provide any real benefit than just checking directly.

structure/tree/avl.go
guuzaa: Thanks for you suggestion. The implementation  of `isNil` is 
```go
func (t *binaryTree[T]) isNil(node *Node[T]) bool {
	return node == t.NIL
}
```
which is equivalent to `node == nil` in `AVL` and `BinarySearch`. The reason why I add `isNIL` to both trees is make them in accordance with `RB`.

guuzaa: `isNil` could be misleading for the beginner ?  How about explicitly change it to `node == t.NIL`.

tjgurwara99: Please don't mark conversations as resolved unless we come to an agreement 🙏🏼 

Both `AVL` and `BinarySearch` work with `node == nil` then isn't the problem with RB? Unfortunately since we have combined the implementations we have come to a point of leaky abstractions 😭 Our `binaryTree` struct is not basic enough for us to use it as a base for all three trees. What this PR is doing is masking the real problem by using `t.NIL`,  which shouldn't even exist in all binary trees structures like BinarySearch tree and AVL since no child means that it is a leaf by definition.

I really think we're heading into the wrong direction 😓 I guess I'm responsible here apologies for that 🙏🏼 . A few problems that I could identify so far (I don't know how I missed this previously - maybe I was in a rush and didn't review it properly).

Some leaky abstractions that I could identify:

1. The `binaryTree` struct should NOT contain `NIL` as a field (sorry I asked you do combine all this). That field is only specific to RB trees. A leaf node is a node without any children in almost all trees except RB so this is a problem with RB not a general problem with all trees so we need to come up with a better way to figure out how to solve the issue.
2. The `binaryTree` struct should NOT contain `printHelper` that is also leaking logic of what the child composed with `binaryTree` need to do. In fact, Trees don't need to be able to represent themselves. Printing structures are algorithms on their own and should be separated (I think we don't need this in trees for now - can open an issue to create a separate `display` or `printer` package that prints structures in a nice way.)

This is probably my fault, when I suggested in your last PR to combine the `Node` together, I didn't anticipate the issues that will arise from combining all these values - `Height` field in `Node` is a calculable field so it would suffice to make that into a method not a field. The `Color` field however is an issue and the way I see it being corrected is by something quite different. But that is not an easy problem.

I believe that for now, I'm reluctant to accept this PR to fix an issue that is actually a design problem. I would think about this for a few days and come up with an idea on how to combine it all together if we can and if we can't then I would just separate out RB tree from `binaryTree`. Will keep this PR open in case you come up with a better approach.

guuzaa: First, it's impossible to have a base struct, which can hold all three trees, in Go. But we can do this using `Interface` , as the following.
```go
var tree.Tree[int]
t= tree.NewBinarySearch[int]() // or t= tree.NewAVL[int](), t= tree.NewRB[int]()
t.Push(1, 4, 10)
t.Push(-8)
```
Second, I think it's fine to include `NIL` in `binaryTree`. According to the definition, the leaf node is a node without any children in `AVL` and `BinarySearch`, but it is a sentinel value for `nil` in `RB`. Obviously, they're two different perspectives. But if we stand in the shoes of `AVL` and `BinarySearch`, the leaf node is a node and its children are `nil`. We could define `NIL` as a sentinel value for `nil` in all three trees.

Third, create a separate display package that prints structures is better than `printHelper`. I'll remove those methods.

As for the `Height` and `Color`, they definitely shouldn't exist in all three trees. 

- `Height` is used to speed up calculating depth and balance factor. In addition to making it a method instead of a field, we can apply `Height` to the other two trees.

- `Color` represents color in `RB`. We never used it in `AVL` and `BinarySearch`.  This is not a simple problem. The solution I can find is  separate out `RB` from `binaryTree`.

guuzaa: In the latest commit, I have separated `Node` to three different nodes. They all complete `Node` interface.
```go
type Node[T constraints.Ordered] interface {
	Key() T
	Parent() Node[T]
	Left() Node[T]
	Right() Node[T]
}
```
I still save NIL(which helps to avoid implementing duplicate logic) in three trees, but make it unexported now.
Apart from that, I add `Empty` method to the `Tree` interface, as the following.
```go
type Tree[T constraints.Ordered] interface {
	Push(...T)
	Delete(T) bool
	Get(T) (Node[T], bool)
	Empty() bool
	Has(T) bool
	Depth() int
	Max() (T, bool)
	Min() (T, bool)
	Predecessor(T) (T, bool)
	Successor(T) (T, bool)
	PreOrder() []T
	InOrder() []T
	PostOrder() []T
	LevelOrder() []T
	AccessNodesByLayer() [][]T
}
```

tjgurwara99:
Sorry I didn't have time in the past few weeks, will review PR in the weekend. Commenting to make sure the stalebot doesn't mark this as stale.

guuzaa:
To save your time, I will review what this PR does in the following. 😃
Basically, all the three trees complete the following interface.
```go
type Tree[T constraints.Ordered] interface {
	Push(...T)
	Delete(T) bool
	Get(T) (Node[T], bool)
	Empty() bool
	Has(T) bool
	Depth() int
	Max() (T, bool)
	Min() (T, bool)
	Predecessor(T) (T, bool)
	Successor(T) (T, bool)
	PreOrder() []T
	InOrder() []T
	PostOrder() []T
	LevelOrder() []T
	AccessNodesByLayer() [][]T
}
```
Apart from `Push` and `Delete`, the other methods share the same logic. In the earlier version, we introduced the `binaryTree` struct which wraps these methods. However, this way leads to a point of leaky abstractions. Because all the three trees share the same `Node`. To avoid this, in this version, I introduce a `Node` interface as the following.
```go
type Node[T constraints.Ordered] interface {
	Key() T
	Parent() Node[T]
	Left() Node[T]
	Right() Node[T]
}
```
It's completed by three different `Node`s (see `avl.go`, `rbtree.go` and `bstree.go`). These `Node`s have different fields. As we say, there is a lot of methods having the same logic. To avoid implementing the duplicate logic , I also write some helper functions (see `tree.go`) . Those functions take the `Node` interface as its parameter, which helps to avoid leaky abstractions. Note that I still save the sentinel value  `NIL`, but make it unexported.

guuzaa:
Hi there, 
Could someone review this PR? Thanks in advance.

raklaptudirm:
@tjgurwara99

tjgurwara99:
Yes, I have reviewed it - I think for now it's fine. I would be changing this in a few days/weeks to remove the sentinel value where we don't need this but that's a concern for later. Thanks for the PR

tjgurwara99 (approved):

raklaptudirm (approved):

raklaptudirm:
@guuzaa Please update this branch to be up to date.

guuzaa:
@raklaptudirm Great, I'm done.

raklaptudirm:
@tjgurwara99 The test package `tree_test` is not ignored by your documentation tool.

tjgurwara99:
👍🏼 I'll take a look this week

PR #590, opened 2022-10-23
feat: added bogosort algorithm

Focusucof:
Description:

Added bogosort algorithm under sort/bogosort.go. I also implemented the test implementations in a file other than sorts_test.go due to nature of how long bogosort takes to finish.

Checklist
[x] Added description of change
[x] Added tests and example, test must pass
[x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/Go/blob/master/CONTRIBUTING.md#Commit-Guidelines)
[x] Search previous suggestions before making a new one, as yours may be a duplicate.
[x] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/sort/bogosort.go
+++ b/sort/bogosort.go
@@ -3,18 +3,18 @@
 // details:
 // A simple bogosort algorithm implementation. [Wikipedia](https://en.wikipedia.org/wiki/Bogosort)
 // author(s): [Focusucof](https://github.com/Focusucof)
-// see sorts_test.go for testing
+// see bogosort_test.go for tests
 
 package sort
 
 import (
 	"github.com/TheAlgorithms/Go/constraints"
-	"sort"
 	"math/rand"
+	"sort"
 )
 
 // shuffles the given int array
-func shuffle[T constraints.Ordered](arr *[]int) {
+func shuffle(arr *[]int) {
 	rand.Shuffle(len(*arr), func(i, j int) {
 		(*arr)[i], (*arr)[j] = (*arr)[j], (*arr)[i] // swap values
 	})
--- a/sort/bogosort.go
+++ b/sort/bogosort.go
@@ -3,7 +3,7 @@
 // details:
 // A simple bogosort algorithm implementation. [Wikipedia](https://en.wikipedia.org/wiki/Bogosort)
 // author(s): [Focusucof](https://github.com/Focusucof)
-// see sorts_test.go for test
+// see sorts_test.go for testing
 
 package sort
 
--- a/sort/bogosort.go
+++ b/sort/bogosort.go
@@ -1,0 +1,31 @@
+// bogosort.go
+// description: Implementation of the bogosort algorithm
+// details:
+// A simple bogosort algorithm implementation. [Wikipedia](https://en.wikipedia.org/wiki/Bogosort)
+// author(s): [Focusucof](https://github.com/Focusucof)
+// see bogosort_test.go for test
+
+package sort
+
+import (
+	"sort"
+	"math/rand"
+)
+
+// shuffles the given int array
+func shuffle(arr *[]int) {
+	rand.Shuffle(len(*arr), func(i, j int) {
+		(*arr)[i], (*arr)[j] = (*arr)[j], (*arr)[i] // swap values
+	})
+}
+
+// loops until array is sorted
+func bogosort(arr []int) []int {
+	for {
+		if sort.IntsAreSorted(arr) { // if array is sorted, break loop and return it
+			return arr
+		} else {
+			shuffle(&arr) // else, shuffle array again
+		}
+	}
+}
--- a/sort/bogosort.go
+++ b/sort/bogosort.go
@@ -20,7 +20,7 @@ func shuffle(arr *[]int) {
 }
 
 // loops until array is sorted
-func bogosort(arr []int) []int {
+func BogoSort(arr []int) []int {
 	for {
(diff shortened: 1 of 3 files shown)


raklaptudirm (commented):
Testing bogosort will absolutely kill the workflows, so I dont think it is a good idea. In fact, I don't think there is anything to learn from bogosort and vote to not add this algorithm. @tjgurwara99 @siriak what do you guys think?

siriak (approved):
From CI logs: `ok  	github.com/TheAlgorithms/Go/sort	4.228s`. It took only 4 seconds to test all sorts. Conceptually, I would include it because it's a sorting algorithm which works. Maybe not a practical one, but we collect algorithms here for education and I think that this one is worth knowing. If performance is a concern, test cases can be shrunk from 10 elements to 9, 8,  7, 6, or 5, depending on how fast we want tests to run.

siriak (approved):
@tjgurwara99 what do you think?

sort/bogosort.go
```diff
@@ -0,0 +1,32 @@
...
+
+// shuffles the given int array
+func shuffle(arr *[]int) {
+	rand.Shuffle(len(*arr), func(i, j int) {
+		(*arr)[i], (*arr)[j] = (*arr)[j], (*arr)[i] // swap values
+	})
+}
+
+// loops until array is sorted
+func BogoSort[T constraints.Ordered](arr []int) []int {
```
tjgurwara99, line 24:
```suggestion
func Bogo[T constraints.Ordered](arr []T) []T {
```

I think `Bogo` is the correct name here, sort is in the name of the package.

Also, you used `Ordered` `constraint` but didn't really use them anywhere in the algorithm. The function may seem generic to you right now, but it won't accept float slices even when you provide it with the float constraint. (you should check this yourself to better understand where you went wrong!)

sort/bogosort.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// shuffles the given int array
+func shuffle(arr *[]int) {
+	rand.Shuffle(len(*arr), func(i, j int) {
+		(*arr)[i], (*arr)[j] = (*arr)[j], (*arr)[i] // swap values
+	})
+}
+
+// loops until array is sorted
```
tjgurwara99, line 23:
```suggestion
// Bogo sorts the provided array by looping over and shuffling its elements until the array is sorted.
```

sort/bogosort.go
```diff
@@ -0,0 +1,32 @@
...
+
+package sort
+
+import (
+	"github.com/TheAlgorithms/Go/constraints"
+	"math/rand"
+	"sort"
+)
+
+// shuffles the given int array
```
tjgurwara99, line 16:
```suggestion
// shuffle shuffles the given int array
```

sort/bogosort.go
```diff
@@ -0,0 +1,32 @@
...
+func shuffle(arr *[]int) {
```
tjgurwara99, line 17:
```suggestion
func shuffle[T constraints.Ordered](arr *[]T) {
```

sort/bogosort_test.go
```diff
@@ -0,0 +1,143 @@
...
+			input:    []int{},
+			expected: []int{},
+			name:     "Empty Slice",
+		},
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		for _, test := range sortTests {
+			BogoSort[int](test.input)
+		}
```
tjgurwara99, line 141: This is not how you benchmark. The idea of benchmarking is to check one thing consistently to get the answer to question: "how fast does it sort an unsorted slice?" but your benchmark answers the question: "how fast does it sort 8 different scenarios?"

You only need one case in benchmark.

sort/bogosort.go
```diff
@@ -0,0 +1,32 @@
...
+	rand.Shuffle(len(*arr), func(i, j int) {
+		(*arr)[i], (*arr)[j] = (*arr)[j], (*arr)[i] // swap values
+	})
+}
+
+// loops until array is sorted
+func BogoSort[T constraints.Ordered](arr []int) []int {
+	for {
+		if sort.IntsAreSorted(arr) { // if array is sorted, break loop and return it
```
tjgurwara99, line 26: Implement this yourself. If you're importing the `sort` package from the standard library then it doesn't really teach anyone anything.

sort/bogosort.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/TheAlgorithms/Go/constraints"
+	"math/rand"
+	"sort"
+)
+
+// shuffles the given int array
+func shuffle(arr *[]int) {
+	rand.Shuffle(len(*arr), func(i, j int) {
+		(*arr)[i], (*arr)[j] = (*arr)[j], (*arr)[i] // swap values
+	})
```
tjgurwara99, line 20: I also think this should be implemented yourself as well rather than depending on the `rand` package but it might be difficult (without the assembly language primitives) so I leave it up to you to decide whether you should do it or not.

tjgurwara99 (changes requested):
I don't mind keeping this algorithm. I just think that it shouldn't be run inside the CI since there is an unbounded number of shuffles that could happen in a slice the larger the slice gets - think powerset 2^{sizeof(slice)}.

Also, I think that the current implementation of the algorithm heavily depends on the standard library `sort.Package` which really defeats the purpose of this package 😂 If the author is willing to create their own implementation of `Sorted` then I don't mind this algorithm being in the repo 😄 (just `Skip` the test as prescribed)

PR #593, opened 2022-10-25
feat: added a1z26 cipher

Focusucof:
Description:

Added a1z26 cipher under cipher/a1z26.go. Test cases are implemented in cipher/a1z26_test.go.

Checklist
[x] Added description of change
[x] Added tests and example, test must pass
[x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/Go/blob/master/CONTRIBUTING.md#Commit-Guidelines)
[x] Search previous suggestions before making a new one, as yours may be a duplicate.
[x] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/cipher/a1z26/a1z26.go
+++ b/cipher/a1z26/a1z26.go
@@ -1,0 +1,66 @@
+// a1z26.go
+// description: Implementation of the a1z26 cipher
+// details:
+// The A1Z26 cipher is a simple substiution cipher where each letter is
+// replaced by the number of the order they're in. For example, A corresponds to
+// 1, B = 2, C = 3, etc..
+//
+// author(s): [Focusucof](https://github.com/Focusucof)
+// see a1z26_test.go for tests
+
+// ref: https://www.dcode.fr/letter-number-cipher
+package a1z26
+
+import (
+	"strings"
+	"strconv"
+)
+
+var encryptionMap = []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"}
+
+// helper function to find the index of a given value in an array
+func indexOfLetter(slice []string, item string) string {
+    for i := range slice {
+        if slice[i] == item {
+            return strconv.Itoa(i)
+        }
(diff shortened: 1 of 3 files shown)


tjgurwara99 (changes requested):
This implementation fails to consider what happens when you provide information not within the bounds of a-z. For example, if I gave you a string such as `thê quick brown fox` what do you think the output would be? what other information would be missing if you don't consider it? and what do you think the output should be?

Focusucof:
Added input validation for the encryption function 🙂

cipher/a1z26/a1z26.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+var encryptionMap = []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"}
+
+// function to check if input is valid
+func isValidInput(input string) bool {
+	return regexp.MustCompile(`^[a-z A-Z]+$`).MatchString(input) // check if input contains only valid letters
+}
+
+// helper function to find the index of a given value in an array
```
tjgurwara99, line 27:
```suggestion
// indexOfLetter function to find the index of a given value in an array
```

Please use the name of the function as the first word in the documentation comment. Please read the commenting conventions of Go language.

cipher/a1z26/a1z26.go
```diff
@@ -0,0 +1,76 @@
...
+// details:
+// The A1Z26 cipher is a simple substitution cipher where each letter is
+// replaced by the number of the order they're in. For example, A corresponds to
+// 1, B = 2, C = 3, etc..
+//
+// author(s): [Focusucof](https://github.com/Focusucof)
+// see a1z26_test.go for tests
+
+// ref: https://www.dcode.fr/letter-number-cipher
+package a1z26
```
tjgurwara99, line 12: I don't like the name of the package. I think you can just call it `sub` instead of `a1z26`. As its a simple substitution cipher (though I feel its more like encoding rather than a cipher).

tjgurwara99 (changes requested):

PR #595, opened 2022-10-27
algorithm: Pronic number

itsAkshayDubey:
This PR contains commits related to pronic number implementation.

Thanks.


math/pronicnumber.go
```diff
@@ -0,0 +1,26 @@
...
+// details: A number n is a pronic number if
+// it is equal to product of two consecutive numbers m and m+1.
+// wikipedia: https://en.wikipedia.org/wiki/Pronic_number
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see pronicnumber_test.go
+
+package math
+
+// This function returns true if argument passed to the function is pronic, false otherwise.
+func PronicNumber(n int) (bool, error) {
```
vil02, line 12: You could simplify your code a lot by changing the argument type to `uint` - then you do not need to check if `n` is nonnegative and you do not need to worry about the `error`.

Moreover you can optimize the computation:
- every pronic number is even, so if `n` is odd you can return `false`,
- it is enough to check if `n == x*(x+1)`, where `x` is `floor(sqrt(n))` (pseudo code) - this will reduce the time compexity.

raklaptudirm: `int` is more commonly used than `uint` in code doing actual maths(which will be concerned with whether a number is pronic or not), so I don't think that's a good idea.

math/pronicnumber.go
```diff
@@ -0,0 +1,26 @@
...
+	switch {
+	case n < 0:
+		return false, ErrPosArgsOnly
```
raklaptudirm, line 15: Returning an error doesn't make sense. The function's job is to check if the number is pronic so just return `false`. The `error` value is unnecessary overhead.

raklaptudirm (changes requested):

math/pronicnumber.go
vil02: In this case I would simply return `false` for negative inputs and get rid of the `error` as you suggest [below](https://github.com/TheAlgorithms/Go/pull/595#discussion_r1006852041).

math/pronicnumber.go
itsAkshayDubey: Fixed in [this](https://github.com/TheAlgorithms/Go/pull/595/commits/fa17bc22ff61a99c28cb38d86f0a5a860ea45c9e) commit.

math/pronicnumber.go
itsAkshayDubey: All the suggested changes have been pushed, kindly have a look.

Thanks, @raklaptudirm & @vil02 for your suggestions.

raklaptudirm (approved):

math/pronicnumber_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+import (
+	"testing"
+
+	"github.com/TheAlgorithms/Go/math"
+)
+
+func TestPronicNumber(t *testing.T) {
+	var tests = []struct {
+		name          string
```
vil02, line 16: The `name` seems to be redundant in this structure. One can save lots of bytes by _generating_ the test name using `n`.

math/pronicnumber.go
```diff
@@ -0,0 +1,24 @@
...
+
+import "math"
+
+// This function returns true if argument passed to the function is pronic, false otherwise.
+func PronicNumber(n int) bool {
+	switch {
+	case n < 0 || n%2 == 1:
+		return false
+	case n == 0:
+		return true
```
vil02, line 19: The 'case == 0' could be treated in the `default` case. I suggest to use `if ... else` here.

math/pronicnumber.go
```diff
@@ -0,0 +1,24 @@
...
+// description: Returns true if the number is pronic, false otherwise
+// details: A number n is a pronic number if
+// it is equal to product of two consecutive numbers m and m+1.
+// wikipedia: https://en.wikipedia.org/wiki/Pronic_number
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see pronicnumber_test.go
+
+package math
+
+import "math"
```
vil02, line 11: In order not to confuse [`math`](https://github.com/TheAlgorithms/Go/tree/master/math) with the [standard math](https://pkg.go.dev/math)
```suggestion
import stdMath "math"
```

math/pronicnumber.go
```diff
@@ -0,0 +1,24 @@
...
+// This function returns true if argument passed to the function is pronic, false otherwise.
+func PronicNumber(n int) bool {
+	switch {
+	case n < 0 || n%2 == 1:
+		return false
+	case n == 0:
+		return true
+	default:
+		x := math.Floor(math.Sqrt(float64(n)))
+		return n == int(x*(x+1))
```
vil02, line 22: I am not sure how does this behave in case of an overflow. @itsAkshayDubey could you please add two test cases with the biggest pronic number, which fits to 32 and 64 bit int? Maybe we will be lucky and the largest 32/64 bit pronic numbers are still small enough.

vil02 (changes requested):

math/pronicnumber_test.go
tjgurwara99: I don't this that is required. This is fine to do IMO.

math/pronicnumber.go
```diff
@@ -0,0 +1,24 @@
...
+// it is equal to product of two consecutive numbers m and m+1.
+// wikipedia: https://en.wikipedia.org/wiki/Pronic_number
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see pronicnumber_test.go
+
+package math
+
+import "math"
+
+// This function returns true if argument passed to the function is pronic, false otherwise.
```
tjgurwara99, line 13: Please use Go's documenting conventions.

```suggestion
// PronicNumber returns true if argument passed to the function is pronic and false otherwise.
```

Also, I think you should provide a definition of what `pronic` means.

tjgurwara99 (changes requested):

itsAkshayDubey:
@tjgurwara99 and @vil02 I added all the suggested changes. 
Kindly have a look.

Thanks,
Akshay

math/pronicnumber_test.go
```diff
@@ -0,0 +1,438 @@
...
+		{"n = 393", 393, false},
+		{"n = 394", 394, false},
+		{"n = 395", 395, false},
+		{"n = 396", 396, false},
+		{"n = 397", 397, false},
+		{"n = 398", 398, false},
+		{"n = 399", 399, false},
+		{"n = 400", 400, false},
+		{"n = 2147441940", 2147441940, true},
+		{"n = 9223372033963249664", 9223372033963249664, true},
```
vil02, line 423: `9223372033963249664` doesn't seem to be a pronic number. Please have a look [here](https://www.online-python.com/CxcDYaWjsZ).

vil02 (changes requested):
There is some overflow issue. Interesting.

math/pronicnumber_test.go
itsAkshayDubey: @vil02 I have found out that the largest 64-bit Pronic number is `9223372033963249500` and can be checked [here](https://www.online-python.com/npLgcyVFHG). 
I've push the related code fix and unit tests. Kindly have a look.

Thanks

vil02: @itsAkshayDubey it seems to be correct now. Good job.

math/pronicnumber_test.go
```diff
@@ -0,0 +1,439 @@
...
+
+import (
+	"testing"
+
+	"github.com/TheAlgorithms/Go/math"
+)
+
+func TestPronicNumber(t *testing.T) {
+	var tests = []struct {
+		name          string
```
vil02, line 16: I [still](https://github.com/TheAlgorithms/Go/pull/595#discussion_r1007868973) think that `name` is redundant here. If you insist to have it, I the _rows_ should look like:
- `{"41 is not pronic", 41, false}`,
- `{"42 is pronic", 42, true}`.

vil02 (commented):
Looks good to me, besides the `name` in the _test cases_.

math/pronicnumber_test.go
```diff
@@ -0,0 +1,439 @@
+// pronicnumber_test.go
+// description: Returns true if number is pronic, false otherwise
```
tjgurwara99, line 2:
```suggestion
```

math/pronicnumber.go
```diff
@@ -0,0 +1,21 @@
+// pronicnumber.go
+// description: Returns true if the number is pronic and false otherwise
+// pronic: Consisting of a base or root or first power plus a higher power.
```
tjgurwara99, line 3: It seems I was mistaken, you had already defined `pronic number` but I had overlooked it in my previous review. Apologies. You can remove the line now 😄

```suggestion
```

or change the `details` block to the definition like so
```suggestion
// Pronic number: For any integer n, if there exists integer m such that n = m * (m + 1) then n is called a pronic number.
```

I ask you to remove it because that sentence doesn't make any sense 😓

tjgurwara99 (changes requested):

math/pronicnumber.go
itsAkshayDubey: Fixed this 👍

math/pronicnumber_test.go
itsAkshayDubey: Fixed this 👍

math/pronicnumber.go
```diff
@@ -0,0 +1,21 @@
+// pronicnumber.go
+// description: Returns true if the number is pronic and false otherwise
+// details:
+//  Pronic number: For any integer n, if there exists integer m
+//  such that n = m * (m + 1) then n is called a pronic number.
```
vil02, line 5:
```suggestion
//  Pronic number: A number n is pronic, if there
//  exists an integer m such that n = m * (m + 1)
```
I am not sure about the proper indentation here - @tjgurwara99  knows for sure how it should be done exactly.
What do you think about being consistent and using the same symbol for the _number that exists_? Now in the definition `m` is used and in the implementation there is `x`.

tjgurwara99: The way described above was done based on my suggestion earlier - which is the proper way to define mathematical definitions in proper logical order. Though arguably English is ambiguous and that order also makes sense.

vil02: There is a subtle difference between the proposed definitions. If you would have an exercise about proving that a given number is pronic and the "current definition" is used, then you have to:
- show that your number is an integer,
- show that it is a product of two consecutive integers.

_Obviously_ the first statement follows from the second one. But, in order to be fully correct, you still have to do it. Therefore I suggest to omit the _requirement_ that a pronic number is an integer.

Moreover using the quantifier "For any n..." makes the thing much more complicated.

tjgurwara99: I don't know what you mean by "current definition" which one are you referring to? I think you're referring to your definition to prove your point but lets see.

The exercise "if you would have an exercise about proving that a given number is pronic" could just as easily be changed to "If you would have an exercise about proving that a given integer is pronic". Now you don't need to prove it and if you were to look closely at the function definition, the priori of the number being an integer is already there, that is, it only accepts integers.

> Obviously the first statement follows from the second one. 

Yes, and it is trivially so... unless you're proving the integer's are closed under multiplication

> Moreover using the quantifier "For any n..." makes the thing much more complicated.

Depends, but lets not go into that in this discussion

vil02: > I don't know what you mean by "current definition" which one are you referring to? I think you're referring to your definition to prove your point but lets see.

I was referring to:
```
//  Pronic number: For any integer n, if there exists integer m
//  such that n = m * (m + 1) then n is called a pronic number.
```
Going in this direction one could also write thing like:
```
//  Pronic number: For any positive and even integer n greater or equal 2, if there exists integer m
//  such that n = m * (m + 1) then n is called a pronic number.
```
Therefore I suggest [this](https://github.com/TheAlgorithms/Go/pull/595#discussion_r1017509813).

tjgurwara99: Hahah, please read the part about function definition 😄 it isn't `func PronicNumber(n any)` it is `func PronicNumber(n int)`

vil02: > Hahah, please read the part about function definition 😄 it isn't `func PronicNumber(n any)` it is `func PronicNumber(n int)`

[We already discussed that](https://github.com/TheAlgorithms/Go/pull/595#discussion_r1006597506). It should be `func PronicNumber(n uint)`. Jokes aside: you are mixing two things here: the documentation is _for the human_ and the _function signature_ is for _for the computer_ (and for the human). Very roughly speaking any definition is easier to understand/work with if there are fewer requirements.



> If you want to be combative about things and your suggestions, I would suggest you don't do that here 🙏🏼

I am faraway of doing that. Suggestion is just a suggestion. If it will not be accepted, well I do not care. I will rethink it, think about the reasons why it was not accepted and next time I will make a better one.

tjgurwara99: I like this conversation, let's move this to a [discussion](https://github.com/TheAlgorithms/Go/discussions/605). If needed we can make changes to the docs in a later PR - let's not block it for now. 😄

tjgurwara99 (approved):

raklaptudirm (approved):

PR #596, opened 2022-10-29
fix: implement BalancedParenthesis for different types of brackets

kalyan-arepalle:
Current parenthesis function can check whether a string has balanced parenthesis but is applicable to only one type of brackets - (), implemented the same to support other types {}, [] as well using stack


kalyan-arepalle:
I see it is already implemented in a different file nestedbrackets.go

PR #599, opened 2022-10-30
algorithm: check if string is isogram

M3talM0nk3y:


strings/isogram/isisogram.go
```diff
@@ -0,0 +1,41 @@
+// Checks if a given string is an isogram.
+// An isogram is a word, phrase, or sentence in which no letter of the alphabet occurs more than once.
+// wiki: https://en.wikipedia.org/wiki/Heterogram_(literature)#Isograms
+// Author: M3talM0nk3y
+
+package isogram
```
raklaptudirm, line 6: Declaring a new package is unnecessary. Make this a method of the `string` package.

raklaptudirm (changes requested):

strings/isogram/isisogram.go
M3talM0nk3y: @raklaptudirm Made method part of strings package. Thanks for the feedback.

strings/isisogram.go
```diff
@@ -0,0 +1,41 @@
...
+	re := regexp.MustCompile(`\d`)
+	return re.MatchString(text)
+}
+
+func hasSymbol(text string) bool {
+	re := regexp.MustCompile(`[-!@#$%^&*()+]`)
+	return re.MatchString(text)
+}
+
+func IsIsogram(text string) bool {
```
tjgurwara99, line 23: What you have here is a first order isogram. The ideas of isograms is more general and there are first, second and third order isograms.

Can you generalise this to figure out first second and third order isograms?

M3talM0nk3y: @tjgurwara99 Checking for first-order, second-order, and third-order isograms. Thanks for the feedback.

M3talM0nk3y:
Hey Taj, anything else that should be improved? Thanks. @tjgurwara99

tjgurwara99:
There are a few things that I wanted to write about but I haven't had any time - apologies, I've been busy. Will write up a proper review sometime later this week (or weekend).

A kind request 🙏🏼 unless it is really urgent please don't tag the maintainers for **reviews**, it creates unnecessary notification noise for people who are working (this is a voluntary role and I have a day job as well) I hope you understand 😄

PS: I'm only talking about reviews - other things such as discussions and such are fine 😄

M3talM0nk3y:
Taj, totally understand and appreciate your response. Feel free to tag me when you can. 😄

Thanks again and hope you have a great day! Talk soon.

strings/isisogram.go
```diff
@@ -0,0 +1,54 @@
...
```
tjgurwara99, line 25: When I suggested the previous change, I didn't mean that you should modify your function such that the user doesn't know which Isogram it is. Currently, the caller of this function has no idea whether they received first, second or third order isogram. My point was to change this function such that it takes the order as argument.

```suggestion
func IsIsogram(s string, order int) (bool, error) {
	if order == 0 {
		return false, errors.New("please provide the order to check for ordered isogram the string")
	}
```

tjgurwara99 (changes requested):

strings/isisogram.go
M3talM0nk3y: Thanks for the feedback, Taj. Made the suggested changes.

strings/isisogram.go
```diff
@@ -0,0 +1,69 @@
...
+	"errors"
+	"regexp"
+	"strings"
+)
+
+const (
+	firstOrder  int = 1
+	secondOrder int = 2
+	thirdOrder  int = 3
+)
```
tjgurwara99, line 20: The pattern for this is a bit different.

The idea is to create a new type and pass that type and create these constants (that should be exported)

```suggestion
type order int

const (
	First  order = iota +1
	Second
	Third
)
```

The order is in the name of the new type we defined to can easily be inferred. Check this [playgound](https://go.dev/play/p/7dcoMZe0wZz) for more information about this pattern

strings/isisogram.go
```diff
@@ -0,0 +1,69 @@
...
+	return re.MatchString(text)
+}
+
+func hasSymbol(text string) bool {
+	re := regexp.MustCompile(`[-!@#$%^&*()+]`)
+	return re.MatchString(text)
+}
+
+func IsIsogram(text string, order int) (bool, error) {
+	if order != firstOrder && order != secondOrder && order != thirdOrder {
```
tjgurwara99, line 33: Check the [playground](https://go.dev/play/p/7dcoMZe0wZz) for a simpler implementation.

tjgurwara99 (changes requested):

tjgurwara99:
Also, I wanted to mention another pattern that is quite idiomatic in Go, you can read more about it [here](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

strings/isisogram.go
M3talM0nk3y: Makes sense. I did have to export the new type as part of the tests. I had initially added an isogram package but it was suggested to move this to the root of the strings package. Let me know what you think.

strings/isisogram.go
M3talM0nk3y: Simplified the validation. Thanks for pointing that out.

strings/isisogram.go
```diff
@@ -13,10 +13,12 @@ import (
 	"strings"
 )
 
+type IsogramOrder int
```
tjgurwara99, line 16: This doesn't need to be exported - your constants are exported and would work properly without it.

tjgurwara99 (changes requested):

strings/isisogram.go
M3talM0nk3y: IMO, the solution is to follow Go's standards and have an `isogram` package similar to the others (ie. `palindrome`). Thoughts?

tjgurwara99: What do you mean? You don't need to export a `type` to export its `constants`. That is standard in Go, you can try looking at [this](https://go.dev/play/p/znzjqzxObwp).

tjgurwara99: > we would normally also have the test file under the same package

No, that is why the `{package_name}_test` is allowed to live in the same directory, otherwise it won't compile if you change the package name to anything else.

M3talM0nk3y: Ah, yes. That's where I got confused since my concern was with using that type with the struct defined for the tests. Apologies for not explaining that better and creating this back and forth. Appreciate your comments in this.

tjgurwara99 (approved):

raklaptudirm (approved):

raklaptudirm (approved):

PR #600, opened 2022-10-31
feat(cipher/polybius): add fuzz test to polybius cipher

mcaci:
Hi again, I keep adding fuzz tests for #480, this time for Polybius. I tested it with go test ./cipher/polybius/ to test the corpus and go test -fuzz=FuzzPoly ./cipher/polybius/ to run the fuzzer itself.
If that Is that ok for you to add it let me know. Also if you have any feedbacks about my change let me know so that I can adapt it.
You will see that I did some changes to the errors and restrict some more the possible inputs to the algorithm based a bit on the description of the algorithm in the wikipedia page mentioned at the beginning of the polybius file.
Thanks again for your work!

### DIFF
--- a/cipher/polybius/polybius.go
+++ b/cipher/polybius/polybius.go
@@ -19,11 +19,22 @@ type Polybius struct {
 // If the size of "chars" is longer than "size",
 // "chars" are truncated to "size".
 func NewPolybius(key string, size int, chars string) (*Polybius, error) {
+	if size < 0 {
+		return nil, fmt.Errorf("provided size %d cannot be negative", size)
+	}
 	key = strings.ToUpper(key)
+	if size > len(chars) {
+		return nil, fmt.Errorf("provided size %d is too small to use to slice string %q of len %d", size, chars, len(chars))
+	}
+	for _, r := range chars {
+		if (r < 'a' || r > 'z') && (r < 'A' || r > 'Z') {
+			return nil, fmt.Errorf("provided string %q should only contain latin characters", chars)
+		}
+	}
 	chars = strings.ToUpper(chars)[:size]
-	for idx, ch := range chars {
-		if strings.Contains(chars[idx+1:], string(ch)) {
-			return nil, fmt.Errorf("\"chars\" contains same character: %c", ch)
+	for i, r := range chars {
+		if strings.ContainsRune(chars[i+1:], r) {
+			return nil, fmt.Errorf("%q contains same character %q", chars[i+1:], r)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


cipher/polybius/polybius.go
```diff
@@ -8,6 +8,54 @@ import (
 	"strings"
 )
 
+type ErrNegativeSize struct{ size int }
```
tjgurwara99, line 11: Is there a reason why do you need context aware errors? I don't see any use of context that you're passing - a simple `fmt.Errorf("whatever")` works better (IMO)

You can simply use the `fmt.Errorf("size %d provided", size)` to add context for logging right? This pattern is unusual IMO - unless you want the user to care about different types of `error`s (bad idea in my opinion generally speaking unless you have a custom logging library or custom recovery scenarios - which is not the case here...). I think in this case `error`s should be just `error`s rather than types that satisfy the `error` interface.

The reason being that

```Go

func something() error {
    return yourCustomErr{1}
}

err := something()
customErr, _ := err.(yourCustomErr)
customErr = yourCustomErr{2} // its modifiable 😱 
```

Since its modifiable, then there is no real benefit here (IMO).

I would rather use this pattern instead (but still advise against it in this case)

```Go

type Error string

func (e Error) Error() string {
    return string(e)
}

// added benefit - its immutable but provides no value
// without `fmt.Errorf("%w: size: %d", ErrSizeTooSmall, size)`
// the returned error can be checked like this `if errors.Is(err, polybius.ErrSizeTooSmall)`
const ErrSizeTooSmall = Error("provide size is too small")


func something(size) error {
    if size < 0 {
        return fmt.Errorf("%w: size %d", ErrSizeTooSmall, size) 
    }
}
```

PS: I used to believe that this was the valid and correct pattern but I don't anymore 😓

tjgurwara99 (changes requested):

cipher/polybius/polybius.go
mcaci: I understand your point of view, I'm checking to change the errors to follow the pattern you suggested me in the last example. The original idea for me was that I added these new types to use them in the errors.As to check the error types returned by the algorithm.

mcaci: Hi @tjgurwara99, I changed the code following your suggestions of removing the context aware errors and use just `fmt.Errorf` and passed the tests again. Let me know if you have any other feedbacks and I will address them. Thanks a lot!

tjgurwara99 (approved):

raklaptudirm (approved):

mcaci:
I have resolved the conversation after the approvals

tjgurwara99 (approved):

tjgurwara99:
@raklaptudirm @siriak I think there is something wrong with my permissions. I'm unable to `squash and merge` the PR - are you both able to do this?

raklaptudirm:
@tjgurwara99 Looks like the base branch was out of date so merging was blocked.

PR #601, opened 2022-10-31
feat(cipher/transposition): add fuzz test to transposition cipher #600

mcaci:
Hi again, in the context of #480 I'm adding the fuzz tests for the transposition algorithm. I tested it with go test -./cipher/transposition/ for the corpus and go test -fuzz=FuzzTransposition ./cipher/transposition/ to run the fuzzer itself.
Let me know if you have any feedbacks about it.
Thanks!


cipher/transposition/transposition_test.go
```diff
@@ -101,3 +101,45 @@ func TestEncryptDecrypt(t *testing.T) {
...
+		if !runesStartsWith([]rune(decrypted), []rune(input)) {
+			t.Fatalf("expected: %+v, got: %+v", []rune(input), []rune(decrypted))
+		}
+	})
+}
+
+func runesStartsWith(s, start []rune) bool {
+	if len(start) == 0 {
+		return true
+	}
```
tjgurwara99, line 138: 😕

A simple `string(decrypted) == input` might suffice no?

mcaci: Unfortunately there are two issues about using it:
The first one is that the decrypted string contains whitespaces after the end of the word (I'm adding an example at the end).

The second reason is that the fuzzer complains when it starts encoding special chars, i.e. [65533](https://www.compart.com/en/unicode/U+FFFD)

Example, when comparing using this code

```go
if !strings.Contains(string(decrypted), input) {
	t.Fatalf("expected: %+v, got: %+v", []rune(input), []rune(decrypted))
}
```

I get this result when launching the fuzzer

```bash
go test ./cipher/transposition -fuzz=FuzzTransposition
fuzz: elapsed: 0s, gathering baseline coverage: 0/139 completed
failure while testing seed corpus entry: FuzzTransposition/deaae2ca3cf382f428a20635cec48a44d2ff6a25a92dd36639ac7d49a542b8c6
fuzz: elapsed: 0s, gathering baseline coverage: 4/139 completed
--- FAIL: FuzzTransposition (0.02s)
    --- FAIL: FuzzTransposition (0.00s)
        transposition_test.go:130: expected: [65533], got: [65533 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32]
```

tjgurwara99: > The first one is that the decrypted string contains whitespaces after the end of the word (I'm adding an example at the end).

Then isn't that a `bug` rather than something that this test should accommodate?

mcaci: Yeah, it was a weird behavior to me when I discovered it the first time. At first I thought to keep this behavior as it was implemented in this way in the first place and the existing unit tests are accommodating it, but not that we are discussing about it I feel more comfortable to change and fix this thing. Thanks for the feedback, I'll work on it

cipher/transposition/transposition.go
```diff
@@ -8,19 +8,15 @@
...
 import (
+	"fmt"
 	"sort"
 	"strings"
 )
 
-type NoTextToEncryptError struct{}
-type KeyMissingError struct{}
+var ErrNoTextToEncrypt = fmt.Errorf("no text to encrypt")
+var ErrKeyMissing = fmt.Errorf("missing Key")
```
tjgurwara99, line 17:
```suggestion
var ErrNoTextToEncrypt = errors.New("no text to encrypt")
var ErrKeyMissing = errors.New("missing Key")
```

There is a technical overhead in using `fmt` package for errors which can be avoided if there is nothing contextual in the error by creating it with the `errors` package.

cipher/transposition/transposition_test.go
```diff
@@ -93,11 +92,54 @@ func TestEncryptDecrypt(t *testing.T) {
...
+	if len(r) != len(s) {
+		return false
+	}
+	for i := range r {
+		if r[i] != s[i] {
+			return false
+		}
+	}
+	return true
+}
```
tjgurwara99, line 145: Seems that you have changed the implementation. Isn't `string(r) == string(s)` working?

tjgurwara99 (changes requested):

cipher/transposition/transposition_test.go
mcaci: Unfortunately it doesn't work when it compares an input string that are not utf8 valid, i.e. containing special characters.

With this check:
```go
if string(decrypted) != string(input) {
	t.Logf("comparing %q with %q as strings", string(input), string(decrypted))
	t.Logf("comparing %q with %q as []rune: equals? %t", []rune(input), []rune(decrypted), equals([]rune(input), []rune(decrypted)))
	t.Logf("checking utf-8 compliance of input: %t; and decrypted: %t", utf8.ValidString(input), utf8.ValidString(decrypted))
	t.Fatalf("expected: %+v, got: %+v", []rune(input), []rune(decrypted))
}
```

I can see that the decrypted not utd8 valid string is encoded in an utf8 valid []rune when the string is assigned to `Decrypt` and this makes the comparison as string at the end failing; even though the rune representation is the same.

```bash
$ go test ./cipher/transposition -fuzz=^FuzzTransposition$
fuzz: elapsed: 0s, gathering baseline coverage: 0/464 completed
failure while testing seed corpus entry: FuzzTransposition/00d38231075839cae8d63ba3ea8aff49bfcfd394f690a57e1cd2923ba0f4d7e3
fuzz: elapsed: 0s, gathering baseline coverage: 2/464 completed
--- FAIL: FuzzTransposition (0.04s)
    --- FAIL: FuzzTransposition (0.00s)
        transposition_test.go:131: comparing "\xcea9" with "�a9" as strings
        transposition_test.go:132: comparing ['�' 'a' '9'] with ['�' 'a' '9'] as []rune
        transposition_test.go:133: checking utf-8 compliance of input: false; and decrypted: true
        transposition_test.go:134: expected: [65533 97 57], got: [65533 97 57]
    
FAIL
exit status 1
FAIL    github.com/TheAlgorithms/Go/cipher/transposition        0.047s
```

So I can present you these alternatives:
1) remove and equals func I created and use `decrypted != string([]rune(input))` to compare input and result
2) before encrypting a string filter out all non utf8 valid strings and return error in case the input is not utf8 valid
3) leave the check with the equals (or reflect.DeepEqual if you prefer) as the underlying []rune representation for both input and decrypted string is the same.

What do you think is the best solution for you?

tjgurwara99: Thanks for the comprehensive information. I have a few thoughts, let me know what you think 😄 


The solution I see is none of those, but rather fixing the main problem of conversion in the function signatures. This is not always possible in real world but in this case nothing in our project makes any backwards compatibility promises therefore you can break the api contracts (for example function signatures - feel free to change them if they are indeed the problem rather than overcoming issues in its usage).

```Go
func Encrypt(text []rune, keyWord string) ([]rune, error) {
```
instead of 
```Go
func Encrypt(text []rune, keyWord string) (string, error) {
```

and 
```Go
func Decrypt(text []rune, keyWord string) ([]rune, error) {
```
instead of 
```Go
func Decrypt(text []rune, keyWord string) ([]rune, error) {
```

Once you have that, you can either use `reflect.DeepEqual` or cast them to strings and compare like so `string(decrypt) != string(input)` or something along those lines.

mcaci: You're welcome! And thanks for your explanation and suggestion as well: in this situation I'm ok to modify the signature of the function and adapt the tests accordingly. I updated the PR, let me know if you agree with the kind of change I did. Thanks again!

tjgurwara99:
I've seen the PR - I haven't been able to test the functionality in my environment yet. I'll post an update after I analyse this algorithm properly.

tjgurwara99 (approved):

raklaptudirm (approved):

PR #602, opened 2022-10-31
feat(cipher/diffiehellman): add fuzz test to key generators in diffiehellman algorithm

mcaci:
Hello, I think this is the last subpackage for the cipher package where the fuzz tests were not yet present. I added two quick fuzz tests for the key generators and tested them with 'go test ./cipher/diffiehellman/ -fuzz=FuzzGenM' and 'go test ./cipher/diffiehellman/ -fuzz=FuzzGenS' to check that the execution was getting along just fine.
Let me know if you have any feedbacks on it. Thanks again!

PR linked to issue #480

### DIFF
--- a/cipher/diffiehellman/diffiehellmankeyexchange_test.go
+++ b/cipher/diffiehellman/diffiehellmankeyexchange_test.go
@@ -39,3 +39,21 @@ func TestDiffieHellmanKeyExchange(t *testing.T) {
 		}
 	})
 }
+
+func FuzzGenSharedKey(f *testing.F) {
+	f.Add(int64(13))
+	f.Fuzz(func(t *testing.T, prvKey int64) {
+		// explicitly ignoring the return value
+		// here we want to test that no strange behaviors are raised when executing GenerateShareKey
+		_ = GenerateShareKey(prvKey)
+	})
+}
+
+func FuzzGenMutualKey(f *testing.F) {
+	f.Add(int64(5), int64(17))
+	f.Fuzz(func(t *testing.T, prvKey, shrdKey int64) {
+		// explicitly ignoring the return value
+		// here we want to test that no strange behaviors are raised when executing GenerateMutualKey
+		_ = GenerateMutualKey(prvKey, shrdKey)
+	})
+}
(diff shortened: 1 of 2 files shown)


cipher/diffiehellman/diffiehellmankeyexchange_test.go
```diff
@@ -39,3 +39,21 @@ func TestDiffieHellmanKeyExchange(t *testing.T) {
...
 	})
 }
+
+func FuzzGenSharedKey(f *testing.F) {
+	f.Add(int64(13))
+	f.Fuzz(func(t *testing.T, prvKey int64) {
+		// explicitly ignoring the return error
+		// here we want to test that no strange behaviors are raised when executing GenerateShareKey
+		_ = GenerateShareKey(prvKey)
+	})
```
tjgurwara99, line 49: This is unnecessary I think... Is there a specific reason why fuzzing is even needed here if we are not checking the error?

cipher/diffiehellman/diffiehellmankeyexchange_test.go
```diff
@@ -39,3 +39,21 @@ func TestDiffieHellmanKeyExchange(t *testing.T) {
...
+}
+
+func FuzzGenMutualKey(f *testing.F) {
+	f.Add(int64(5), int64(17))
+	f.Fuzz(func(t *testing.T, prvKey, shrdKey int64) {
+		// explicitly ignoring the return error
+		// here we want to test that no strange behaviors are raised when executing GenerateMutualKey
```
tjgurwara99, line 56: Can you define "strange" behaviour?

I'm of the opinion that these fuzz tests are for checking the unexpected behaviour - that could only be detected once you check the errors (since there are no `panic`s in the algorithm.)

mcaci: About the reason for these tests: it is mostly to run the function and see if it breaks for unexpected reasons like if there are any issue during the computation done in the algorithm, i.e. divisions by zero, overflows and so on. Let me know what are your thoughts on this.

cipher/diffiehellman/diffiehellmankeyexchange_test.go
mcaci: Actually I wrote it wrong: it's not return error but the return value. I can update that if you agree with it.

tjgurwara99: I didn't understand what you meant, but please update it so that I can see what you mean and we can make progress on this further 😄

cipher/diffiehellman/diffiehellmankeyexchange_test.go
tjgurwara99: I don't think there is anything that is being checked for us to conclude that there is nothing "strange" happening. I don't think this is a change with any merit... 😓

PR #603, opened 2022-11-01
Added string test-cases in sort algorithms

gauravgupta98:
Hi,

I was just exploring the code and found out that we are missing string test cases for sorting algorithms. I've raised this PR to add few test cases for string literals. I'd dig deeper into the code to see where we can improve and would also love to hear back on anything I can contribute.

Looking forward to the feedback!

### DIFF
--- a/sort/sorts_test.go
+++ b/sort/sorts_test.go
@@ -76,7 +76,6 @@ func testFramework(t *testing.T, sortingFunction func([]int) []int) {
 	}
 }
 
-
 func testFrameworkWithStrings(t *testing.T, sortingFunction func([]string) []string) {
 	sortTests := []struct {
 		input    []string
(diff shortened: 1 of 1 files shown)


tjgurwara99:
This PR is doing nothing more than testing a language feature. There have been previous PRs (such as [this](https://github.com/TheAlgorithms/Go/pull/538)) to make the same changes but have been rejected since its testing something that the language guarantees. Closing this PR as this isn't something that needs testing.

gauravgupta98:
Hi @tjgurwara99 , thanks for reviewing the PR.
I agree with your point that it is something that language guarantees but my understanding is that if we have a generic function that accepts any input then there should be test cases which makes sure that code works fine with a bunch of datatypes. Let's say today your method is generic and tomorrow for some reason  you decide to only make it accept integers (would not really happen here  but talking about a general scenario) then your test cases should fail.

tjgurwara99:
Maybe you're worried about regression - you may be thinking if we support `string` sorting then we _must_ keep supporting `string` sorting; that argument only applies to libraries, this is not a library - we don't claim that our changes would be backwards compatible and hence I don't think we need another test. Here we care more about correct implementation not backwards compatibility. In fact we're happy to break backwards compatibility in favour of better implementations.

If regression is not what you're talking about but in general you think that all generic things must be tested, then I argue against it and propose a challenge. If I provide you with a function

```Go
func doSomething(w io.Writer) {
    // do something to the writer w
}
```

Then if I asked you to test this function. How would you go about it? According to your current understanding, the function must be tested against all things that it accepts. How many tests cases would you need to consider? What would you test - the implementation or which types can be passed to the function? and what would your assumptions be?

Similar to the above situation, the type constraints in Go are very (very) similar to interfaces (just a bit more generalised). The only different is that with this new type constraint (think interface) accepts only types such that "underlying" type has `<` operator defined on it, nothing more nothing less.

gauravgupta98:
Ah alright. Got your point now. Thanks for clarifying mate. I'll see where I can contribute in the project.

PR #607, opened 2022-11-20
feat: add heap implementation using generic.

qizhaoTan:
add heap implementation using generic.

Hello, This is the first time I contribute my submission on github, if there is any mistake, please correct me, thanks.


structure/heap/heap.go
```diff
@@ -0,0 +1,79 @@
+package heap
+
+// The Interface type describes the requirements
+// for a type using the routines in this package.
+type Interface[T any] interface {
+	Less(o T) bool
+}
```
tjgurwara99, line 7: I do not think you need this interface... The type constraint `constraints.Ordered` in the constraints package within this repo is what you need I think.

structure/heap/heap.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Heap heap implementation using generic.
+type Heap[T Interface[T]] struct {
+	heaps []T
+}
```
tjgurwara99, line 12:
```suggestion
type Heap[T constraints.Ordered] struct {
	heaps []T
}
```

structure/heap/heap.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (h *Heap[T]) down(parent int) {
+	lessIdx := parent
+	lChild, rChild := (parent<<1)+1, (parent<<1)+2
+	if lChild < len(h.heaps) && h.heaps[lChild].Less(h.heaps[lessIdx]) {
+		lessIdx = lChild
+	}
+	if rChild < len(h.heaps) && h.heaps[rChild].Less(h.heaps[lessIdx]) {
+		lessIdx = rChild
+	}
```
tjgurwara99, line 73:
```suggestion
	if lChild < len(h.heaps) && h.heaps[lChild] < h.heaps[lessIdx] {
		lessIdx = lChild
	}
	if rChild < len(h.heaps) && h.heaps[rChild] < h.heaps[lessIdx] {
		lessIdx = rChild
	}
```

Once you have the `constraints.Ordered` type constraint in the definition for the `Heap`.

structure/heap/heap_test.go
```diff
@@ -0,0 +1,100 @@
+package heap
+
+import (
+	"testing"
+)
+
+type _int int
```
tjgurwara99, line 7: Why? This convention is very different from Go standards, please don't use `underscore` to mark things private 🙏🏼 

```suggestion
type testInt int
```

structure/heap/heap_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (u _int) Less(o _int) bool {
+	return u < o
+}
```
tjgurwara99, line 11: Once you use the type constraint you would not need this...

structure/heap/heap_test.go
```diff
@@ -0,0 +1,100 @@
...
+	_pop   = 2
+	_test  = 3
+	_empty = 4
+)
+
+type _op struct {
+	typ     _opType
+	x       _int
+	isEmpty bool
+}
```
tjgurwara99, line 43: Again, please don't use such convention and refer to some community guidelines on naming convention in Go. Example: [Google's style guide](https://google.github.io/styleguide/go/).

structure/heap/heap_test.go
```diff
@@ -0,0 +1,100 @@
...
+				case _test:
+					if got := h.Top(); got != op.x {
+						t.Errorf("op %d Test() = %v, want %v", i, got, op.x)
+					}
+				case _empty:
+					if got := h.Empty(); got != op.isEmpty {
+						t.Errorf("op %d Empty() = %v, want %v", i, got, op.isEmpty)
+					}
+				}
+			}
```
tjgurwara99, line 97: You're performing your operations but not checking they are correct. Doesn't that defeat the purpose of testing...

structure/heap/heap_test.go
```diff
@@ -0,0 +1,100 @@
+package heap
```
tjgurwara99, line 1:
```suggestion
package heap_test
```

Prefer black box testing for tests that are only testing the exported API.

tjgurwara99 (changes requested):

structure/heap/heap.go
qizhaoTan: We can do this like testing example 2 if we have this interface, but the `constraints.Ordered` can't.

structure/heap/heap.go
qizhaoTan: You can see TestHeapExample2 function in heap_test.go,  the `constraints.Ordered` can't do this.

structure/heap/heap_test.go
qizhaoTan: I think the type constraint can't be used in many scenarios. The `Less` method can be used in any scenarios.

structure/heap/heap.go
qizhaoTan: If we need this Interface,  this modificarion should not be made.

structure/heap/heap.go
qizhaoTan: Thank you so much for your detailed reply, I benefited a lot from this commit. For the two options you mentioned, I provide two methods `New` and `NewAny`, users can choose according to their needs.

tjgurwara99 (approved):
Looks good to me! Thank you for your contribution 😄

raklaptudirm (approved):

PR #611, opened 2022-11-25
algorithm: Hexagonal number

itsAkshayDubey:
This PR contains commits for hexagonal number. Kindly have a look.

Thanks.

### DIFF
--- a/math/hexagonalnumber.go
+++ b/math/hexagonalnumber.go
@@ -1,0 +1,23 @@
+// hexagonalnumber.go
+// description: Returns nth hexagonal number
+// details:
+//  Hexagonal number: The nth hexagonal number hn is the
+//  number of distinct dots in a pattern of dots consisting
+//  of the outlines of regular hexagons with sides up to n dots,
+//  when the hexagons are overlaid so that they share one vertex.
+// wikipedia: https://en.wikipedia.org/wiki/Hexagonal_number
+// author: Akshay Dubey (https://github.com/itsAkshayDubey)
+// see hexagonalnumber_test.go
+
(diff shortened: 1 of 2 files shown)


math/hexagonalnumber.go
```diff
@@ -0,0 +1,23 @@
...
+package math
+
+// HexagonalNumber returns nth hexagonal number.
+func HexagonalNumber(n int) (int, error) {
+	if n < 0 {
+		return 0, ErrPosArgsOnly
+	}
+	if n == 0 {
+		return 0, ErrNonZeroArgsOnly
+	}
```
tjgurwara99, line 21: I really feel like all the algorithms (even the earlier ones) that you've proposed could just as easily benefit from `uint` instead of an `int`, with the added condition that providing `n` gives you the result of `n+1`th item (think arrays, index 0 is index 1). Earlier I was convinced that `int` is the more common choice but upon reflection, casting a `uint` back to `int` is trivial enough so we should just use `uint`. But maybe you have other reasons to support `int` instead, so let me know if you have any arguments against this.

For example in this case,

```Go
// HexagonalNumber returns (n+1)th hexagonal number.
func HexagonalNumber(n uint) uint {
	return (n+1) * (2*n + 1)
}
```

And in which case this algorithm really just becomes a formula. Now whether this adds value for people learning programming (go) is a different topic, I'm personally biased since I'm a mathematician that it **does add value** but at the same time I can see why _it_ would become a bloat on the repository if we keep adding trivial formulas here.

tjgurwara99 (changes requested):

PR #613, opened 2022-12-02
Added sqrt decomposition

lorenzotinfena:
Added the data structure like the one described here (not including mo's algorithm) https://cp-algorithms.com/data_structures/sqrt_decomposition.html.
But I made it more general as possible, using generic functions to query, merge 2 queries, and change the value of an element


lorenzotinfena (commented):
this comment is a test

sqrt_decomposition/sqrt_decomposition_simple.go
```diff
@@ -0,0 +1,105 @@
+package sqrt_decomposition
```
tjgurwara99, line 1: This is a data structure, please put this algorithm inside the `structure` package.

sqrt_decomposition/sqrt_decomposition_simple.go
```diff
@@ -0,0 +1,105 @@
...
+Usage example:
+	slice := []int32{1, 2, 3, 4, 5, 6}
+	s := NewSqrtDecompositionSimple(slice,
+		func(e int32) int32 { return e },
+		func(q1, q2 int32) int32 { return q1 + q2 },
+		func(q, a, b int32) int32 { return q - a + b },
+	)
+	s.Update(3, 10)
+	println(s.Query(1, 6)) // 26
+*/
```
tjgurwara99, line 29: The commenting style is not following the Go standards defined in the `gofmt` program, please fix these issues.

structure/heap/binary_heap.go
```diff
@@ -0,0 +1,98 @@
...
+
+	func (a MyInt32) PriorTo(b MyInt32) bool {
+		return a < b
+	}
+	func main() {
+		h := heap.NewBinaryHeapFromSlice[MyInt32, int32]([]MyInt32{4, 2})
+		h.Push(3)
+		fmt.Println(h.Pop())
+	}
+*/
```
tjgurwara99, line 22: Usage example are presented separately in a Example function inside the test file, please follow the Go standards whenever possible - comments have a tendency to go out of date very quickly without rigorous checks and since we don't have those checks here, an example function is much better than this.

tjgurwara99 (changes requested):
Please read our contribution guidelines. I have reviewed this but I was going to avoid it reviewing this PR because there is no context provided on what you're trying to do and achieve in this PR. You are expecting the maintainers to spend time figuring out most of your code when you haven't even provided a decent description (no description in fact). Just opening a PR without description is not a very good way of contributing to projects.

lorenzotinfena:
I fixed the style

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
```diff
@@ -0,0 +1,95 @@
...
+//
+// if $\exist unionQ:Q,Q\rightarrow Q$
+//
+// s.t.
+//
+// - $\forall n\in \N > 1, 1\le i<n, E_1,..., E_n\in E \\ query(E_1,..., E_n)=unionQ(query(E_1,..., E_i), query(E_{i+1},...,E_n))$
+//
+// - (Only if you want use $update$ function)
+// $\forall n\in \N > 0, E_1,..., E_n\in E \\ query(E_1,...,E_{new},..., E_n)=updateQ(query(E_1,...,E_{old},...,E_n), indexof(E_{old}), E_{new})$
+package sqrtdecompositionsimple
```
tjgurwara99, line 16: The package name is poorly chosen. Remember that programmers are lazy 😂 Could you think of no other name that would be suitable for your data structure?

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
```diff
@@ -0,0 +1,95 @@
...
+
+import "math"
+
+type SqrtDecompositionSimple[E any, Q any] struct {
```
tjgurwara99, line 20: The name of the package and this struct are almost identical - naming things wisely and appropriately as a programmer is a skill that would help you in the long run.

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
```diff
@@ -0,0 +1,95 @@
...
+	unionQ             func(q1 Q, q2 Q) Q
+	updateQ            func(oldQ Q, oldE E, newE E) (newQ Q)
+
+	elements  []E
+	blocks    []Q
+	blockSize uint64
+}
+
+// Assumption: len(elements) > 0
+func NewSqrtDecompositionSimple[E any, Q any](
```
tjgurwara99, line 31: Same applies to the naming of this.

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
```diff
@@ -0,0 +1,95 @@
...
+	elements []E,
+	querySingleElement func(element E) Q,
+	unionQ func(q1 Q, q2 Q) Q,
+	updateQ func(oldQ Q, oldE E, newE E) (newQ Q),
```
tjgurwara99, line 35: I think it would be rather better to have the functions as types themselves (though I have checked whether that would suit with your current generic design). For example

```Go
type QueryFunc[E any, Q any] func(elem E) Q

type UnionFunc[Q any] func(a, b Q) Q

type UpdateFunc[E any, Q any] func(oldQ Q, oldE E, newE E) Q

func NewSqrtDecompositionSimple[E any, Q any](
	elements []E,
	querySingleElement QueryFunc,
 	unionQ UnionFunc,
 	updateQ UpdateFunc,
) ...
```

This way you could even add more control on the exposure of the api.

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
```diff
@@ -0,0 +1,95 @@
...
+	querySingleElement func(element E) Q
+	unionQ             func(q1 Q, q2 Q) Q
+	updateQ            func(oldQ Q, oldE E, newE E) (newQ Q)
+
+	elements  []E
+	blocks    []Q
+	blockSize uint64
+}
+
+// Assumption: len(elements) > 0
```
tjgurwara99, line 30: Please use Go style commenting - read up on `godoc` and how the comments are used to write documentation automatically for public consumption.

structure/sqrtdecompositionsimple/sqrtdecompositionsimple_test.go
```diff
@@ -0,0 +1,79 @@
+package sqrtdecompositionsimple
```
tjgurwara99, line 1: Use the black box testing method for this structure, this is a powerful structure but very fiddly so there may be things that you may have overlooked until you test it as a consumer.

```suggestion
package sqrtdecompositionsimple_test
```
or add `_test` to package name when you change the package name.

tjgurwara99 (changes requested):

lorenzotinfena:
(comment to not close this) I wish I can work on this when I have time

lorenzotinfena:
(comment to not close this)

ankit-gautam23 (commented):
LGTM

siriak:
Don't forget to re-request a review once you are done with the changes :)

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
lorenzotinfena: But since those functions are passed only to a singular method, I think is overengineering doing so

lorenzotinfena:
I tried to do little fixes on comments, and moved to a new folder called sqrt, it may contains in future others algorithms/data structures which use "sqrt advantages", for example mo's algorithm

structure/sqrtdecompositionsimple/sqrtdecompositionsimple.go
tjgurwara99: Sure I agree

lorenzotinfena:
What should I do?

siriak:
@tjgurwara99 could you review again?

tjgurwara99:
I haven't forgotten this PR - I will review as soon as I get some time 🙏🏼

sqrt/sqrtdecomposition.go
```diff
@@ -0,0 +1,102 @@
...
+//
+// if $\exist unionQ:Q,Q\rightarrow Q$
+//
+// s.t.
+//
+// - $\forall n\in \N > 1, 1\le i<n, E_1,..., E_n\in E \\ query(E_1,..., E_n)=unionQ(query(E_1,..., E_i), query(E_{i+1},...,E_n))$
+//
+// - (Only if you want use $update$ function)
+// $\forall n\in \N > 0, E_1,..., E_n\in E \\ query(E_1,...,E_{new},..., E_n)=updateQ(query(E_1,...,E_{old},...,E_n), indexof(E_{old}), E_{new})$
+type SqrtDecomposition[E any, Q any] struct {
```
tjgurwara99, line 21:
```suggestion
type Decomposition[E any, Q any] struct {
```

sqrt is in the name of the package 😄

sqrt/sqrtdecomposition.go
```diff
@@ -0,0 +1,102 @@
...
+
+	elements  []E
+	blocks    []Q
+	blockSize uint64
+}
+
+// Create a new SqrtDecomposition instance with the parameters as specified by SqrtDecomposition comment
+// Assumptions:
+//   - len(elements) > 0
+func NewSqrtDecomposition[E any, Q any](
```
tjgurwara99, line 34:
```suggestion
func NewDecomposition[E any, Q any](
```

Again sqrt is in the name

And maybe call this `Decomp`, instead of `Decomposition`

tjgurwara99 (commented):
Other than those two things, everything else looks good to me. You can also feel free to ignore it since it's an opinions related comment, in which case let me know and I'll approve it 😄

lorenzotinfena:
I did a little naming fix, but not the ones you exposed.
I'm not sure, but I think it make sense to interprete "sqrt" package as an abbreviation of something like "sqrt decomposition based algorithms and ds", and the base data structure (unfortunately) has the same name of the family of algorithms (sqrt decomposition) (or has no name: see https://cp-algorithms.com/data_structures/sqrt_decomposition.html). Can be included also in the same packages other sqrt-decomposition algorithms like mo's algorithm, so call it only Decomposition is not too intuitive to me.

FInally I think we have 2 choices: Let the name as it is, or call it like "BaseDS" (base data structure), what do you like more?

lorenzotinfena:
With all this effort in this conversation can be built tens of sqrt decompositions xd

tjgurwara99 (approved):
The choice of naming is unfortunate and that can also be said to be true for `quicksort` hence we also made an exception there as well, so don't worry too much about naming 😄 It was just a suggestion because I rarely come across these sqrt decomposition based data structures so I know very little about them.

raklaptudirm (approved):

PR #614, opened 2022-12-05
feat: add GUID algorithm

cheatsnake:
Algorithm implementation for GUID strings generation.

Generation occurs for the following pattern `xxxxxxxxxx-xxxxxxx-4xxxxx-xxxxxxx-xxxxxxxxx`, where a random hexadecimal value (from 0 to f) is substituted for the `x` values. The robust `rand` function from the `crypto` package was used to get the random values.

Read more about GUID: https://en.wikipedia.org/wiki/Universally_unique_identifier
Other implementation from TheAlgorithms website: https://the-algorithms.com/search?q=GUID

### DIFF
--- a/strings/guid/guid.go
+++ b/strings/guid/guid.go
@@ -21,11 +21,10 @@ import (
 	"strings"
 )
 
-// GUID v4
 const pattern string = "xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx"
 
-// Return a random globally unique identifier.
-func New() string {
+// New returns a randomly generated global unique identifier.
+func New() (string, error) {
 	var guid strings.Builder
 
 	for i := range pattern {
@@ -34,7 +33,7 @@ func New() string {
 		if char == "x" {
 			random, err := rand.Int(rand.Reader, big.NewInt(16))
 			if err != nil {
-				panic(err)
+				return "", err
 			}
 
 			char = fmt.Sprintf("%x", random.Int64())
@@ -43,5 +42,5 @@ func New() string {
 		guid.WriteString(char)
 	}
 
-	return guid.String()
+	return guid.String(), nil
 }
--- a/strings/guid/guid.go
+++ b/strings/guid/guid.go
@@ -1,0 +1,47 @@
+// guid.go
+// description: Generate random globally unique identifiers (GUIDs).
+// details:
+// A GUID (globally unique identifier) is a 128-bit text string that
+// represents an identification (ID). Organizations generate GUIDs when
+// a unique reference number is needed to identify information on
+// a computer or network. A GUID can be used to ID hardware, software,
+// accounts, documents and other items. The term is also often used in
+// software created by Microsoft.
+// See more information on: https://en.wikipedia.org/wiki/Universally_unique_identifier
+// author(s) [cheatsnake](https://github.com/cheatsnake)
+// see guid_test.go
+
+// Package guid provides facilities for generating random globally unique identifiers.
+package guid
+
+import (
+	"crypto/rand"
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+// GUID v4
+const pattern string = "xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx"
+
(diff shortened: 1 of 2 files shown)


tjgurwara99 (changes requested):

tjgurwara99 (commented):
Please read our contribution guidelines - just providing an algorithm without any PR description wastes our time because we don't know what to expect in the algorithm (even if its documented in the file itself).

cheatsnake:
> Please read our contribution guidelines - just providing an algorithm without any PR description wastes our time because we don't know what to expect in the algorithm (even if its documented in the file itself).

Sorry. Thanks for the review.

strings/guid/guid_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		if expected != result {
+			t.Errorf(`the length of the string should be "%d", but got %d`, len(pattern), len(guid))
+		}
+	})
+}
+
+func BenchmarkNew(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		New()
```
tjgurwara99, line 66:
```suggestion
		_ = New()
```

tjgurwara99 (changes requested):

strings/guid/guid_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		if expected != result {
+			t.Errorf(`the length of the string should be "%d", but got %d`, len(pattern), len(guid))
+		}
+	})
+}
+
+func BenchmarkNew(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		_ = New()
```
tjgurwara99, line 66:
```suggestion
		_, _ = New()
```

Sorry my mistake

tjgurwara99 (changes requested):

strings/guid/guid.go
```diff
@@ -0,0 +1,46 @@
...
+			random, err := rand.Int(rand.Reader, big.NewInt(16))
+			if err != nil {
+				return "", err
+			}
+
+			char = fmt.Sprintf("%x", random.Int64())
+		}
+
+		guid.WriteString(char)
+	}
```
tjgurwara99, line 43: The following are non-blocking suggestions.

```suggestion
	for i, ch := range pattern {
		if i == 14 {
			guid.WriteRune(ch)
			continue
		}
		if ch == '-' {
			guid.WriteRune(ch)
			continue
		}
		random, err := rand.Int(rand.Reader, big.NewInt(16))
		if err != nil {
			return "", err
		}
		guid.WriteString(fmt.Sprintf("%x", random.Int64()))
	}
```

This adds a few lines but makes it somewhat easy to notice what your main intent of the function is. If you notice, there were two cases which were previously hidden from the previous implementation (case when `ch` was in index 14 or when `ch` was `-`), making it somewhat difficult if it turns out there was a problem. Having the problems conditions written as guard statements are much easier to wrap your head around. (Not something that is always possible but should be preferred over simply hiding edge cases)

tjgurwara99 (dismissed):

strings/guid/guid.go
cheatsnake: In that case, it may be a little unclear where the number 14 came from. Perhaps you should put the version index in a separate constant to avoid "magic numbers" in the code.

```suggestion
const pattern string = "xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx"
const versionIndex int = 14

// New returns a randomly generated global unique identifier.
func New() (string, error) {
	var guid strings.Builder

	for i, ch := range pattern {
	        if i == versionIndex {
		        guid.WriteRune(ch)
		        continue
	        }
	        if ch == '-' {
		        guid.WriteRune(ch)
		        continue
	        }
	        random, err := rand.Int(rand.Reader, big.NewInt(16))
	        if err != nil {
		        return "", err
	        }
	        guid.WriteString(fmt.Sprintf("%x", random.Int64()))
        }
	return guid.String(), nil
}
```

tjgurwara99 (approved):

cheatsnake:
@raklaptudirm waiting for approve

PR #615, opened 2022-12-06
fix: fix the problem that the Get method is incorrect when there is a…

xcrossed:
… hashmap conflict

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea/
 .vscode/
+.history/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea/
 .vscode/
+.history/
```
tjgurwara99, line 3: Why was this added? I don't think the repository needs this, you can keep it in your local and not commit it to the repository.

tjgurwara99 (changes requested):
Please provide a description explaining why the change was made. From the outset, looking at the tests, I can't see why this change was necessary. Please provide a description and a test case where you expect the previous implementation to fail and then we can review this further 😄

.gitignore
xcrossed: this dir  is a vscode extension local history use This is a directory used by a vscode or idea plugin called local history, which I can remove.

xcrossed:
> 
The scenario where this happens is: when there is a hash key conflict, the hashmap implementation at that time stores the conflicting nodes through a linked list. Therefore, when the Get method is executed, the linked list needs to be traversed.

PR #617, opened 2022-12-20
add Fibonacci

KhanbalaRashidov:


math/fibonacci/fibonacci.go
```diff
@@ -45,3 +45,18 @@ func Formula(n uint) uint {
 	powPhi := math.Pow(phi, float64(n))
 	return uint(powPhi/sqrt5 + 0.5)
 }
+
+func ForSlice(n uint) uint {
+
+	if n == 0 {
+		return 0
+	}
+	k := make([]uint, n+1)
```
tjgurwara99, line 54: If you're making a slice here and not using it in any meaningful way, what is the point of wasting all that memory? It would be much better to just have three variables without wasting memory.

tjgurwara99 (commented):
I'm not sure this algorithm in its current implementation is providing much value 😅 As you can see, the previous implementations are using simple formula representation of the fibonacci sequence to generate them. The simple fibonacci is missing, I argue that it would have been better to implement the simpler case rather than using a slice in a situation where it isn't necessary.

```Go
func Nth(n uint) uint {
    var a, b, c uint = 1, 1, 2
    for i := 2; i < n + 1; i++ {
        a, b, c = b, c, b+c
    }
    return c
}
```

or something along those lines (as I haven't tested this). As you can see, there is no need for having a separate slice to keep track of `all` previous values in the sequence.

math/fibonacci/fibonacci.go
KhanbalaRashidov: Thank you for the valuable information ☺

PR #620, opened 2023-02-04
Adding Krishnamurthy number algorithm

GooMonk:
Added the Krishnamurthy Algorithm implementation along with the tests and benchmarks for the code. Please review the pull request.

### DIFF
--- a/math/krishnamurthy.go
+++ b/math/krishnamurthy.go
@@ -6,8 +6,10 @@
 // see krishnamurthy_test.go
 package math
 
+import "github.com/TheAlgorithms/Go/constraints"
+
 // IsKrishnamurthyNumber returns if the provided number n is a Krishnamurthy number or not.
-func IsKrishnamurthyNumber[T int32 | int64 | int | uint32 | uint64 | uint](n T) bool {
+func IsKrishnamurthyNumber[T constraints.Integer](n T) bool {
 	if n <= 0 {
 		return false
 	}
--- a/math/krishnamurthy.go
+++ b/math/krishnamurthy.go
@@ -1,0 +1,32 @@
+// filename : krishnamurthy.go
+// description: A program which contains the function that returns true if a given number is Krishnamurthy number or not.
+// details: A number is a Krishnamurthy number if the sum of all the factorials of the digits is equal to the number.
+// Ex: 1! = 1, 145 = 1! + 4! + 5!
+// author(s): [GooMonk](https://github.com/GooMonk)
+// see krishnamurthy_test.go
+package math
+
+import "github.com/TheAlgorithms/Go/constraints"
+
+// IsKrishnamurthyNumber returns if the provided number n is a Krishnamurthy number or not.
+func IsKrishnamurthyNumber[T constraints.Integer](n T) bool {
+	if n <= 0 {
+		return false
+	}
+
+	// Preprocessing: Using a slice to store the digit Factorials
+	digitFact := make([]T, 10)
+	digitFact[0] = 1 // 0! = 1
+
+	for i := 1; i < 10; i++ {
+		digitFact[i] = digitFact[i-1] * T(i)
+	}
+
+	// Subtract the digit Facotorial from the number
+	nTemp := n
+	for n > 0 {
+		nTemp -= digitFact[n%10]
+		n /= 10
+	}
+	return nTemp == 0
+}
--- a/math/krishnamurthy.go
+++ b/math/krishnamurthy.go
@@ -6,10 +6,8 @@
 // see krishnamurthy_test.go
 package math
 
-import "github.com/TheAlgorithms/Go/constraints"
-
 // IsKrishnamurthyNumber returns if the provided number n is a Krishnamurthy number or not.
-func IsKrishnamurthyNumber[T constraints.Integer](n T) bool {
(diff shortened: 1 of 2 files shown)


siriak (approved):
Looks good, thanks!

math/krishnamurthy.go
```diff
@@ -0,0 +1,30 @@
+// filename : krishnamurthy.go
+// description: A program which contains the function that returns true if a given number is Krishnamurthy number or not.
+// details: A number is a Krishnamurthy number if the sum of all the factorials of the digits is equal to the number.
+// Ex: 1! = 1, 145 = 1! + 4! + 5!
+// author(s): [GooMonk](https://github.com/GooMonk)
+// see krishnamurthy_test.go
+package math
+
+// IsKrishnamurthyNumber returns if the provided number n is a Krishnamurthy number or not.
+func IsKrishnamurthyNumber[T int32 | int64 | int | uint32 | uint64 | uint](n T) bool {
```
tjgurwara99, line 10: I don't think there is any real benefit to using this instead of the `contrainst.Integer` constraint. Is there a reason you changed it?

GooMonk: Hi @tjgurwara99, sorry for the late reply. I thought of using int8 and int16 would give wrong results for the number 40585, so thought only allowing the function call for int32 and int64 would be better.

tjgurwara99: Then how about `type myInt int64`? Do you think your algorithm would work for this?

The answer is no 😄 - the point of using a generic constraint to use the same algorithm on similar types but if the algorithm by nature is not benefiting from a generic type then don't use generics.

Also, I don't see why it won't work on `int8` and `int64`. Can you elaborate on why you think it won't work on them? (I even tested them and they seem to be working - so wanted to confirm with you)

math/krishnamurthy_test.go
```diff
@@ -0,0 +1,51 @@
...
+		{3421488712, false, "is not"},
+	}
+}
+
+func TestIsKrishnamurthyNumber(t *testing.T) {
+	for _, test := range retCases() {
+		t.Run(fmt.Sprintf("%d ", test.input)+test.outputString+" a Krishnamurthy Number", func(t *testing.T) {
+			res := IsKrishnamurthyNumber(test.input)
+			if res != test.output {
+				t.Errorf("For input " + fmt.Sprintf("%d ", test.input) + "Expected: " + fmt.Sprintf("%t, ", test.output) + "Found: " + fmt.Sprintf("%t, ", res))
```
tjgurwara99, line 37: `Errorf` is a formatting enabled error printing function. You should use the `fmt` package [verbs](https://pkg.go.dev/fmt#hdr-Printing) for these since that is the purpose of this functionality. This is quite confusing to look at and read.

math/krishnamurthy_test.go
```diff
@@ -0,0 +1,51 @@
...
+		{6327, false, "is not"},
+		{40585, true, "is"},
+		{9743821, false, "is not"},
+		{3421488712, false, "is not"},
+	}
+}
+
+func TestIsKrishnamurthyNumber(t *testing.T) {
+	for _, test := range retCases() {
+		t.Run(fmt.Sprintf("%d ", test.input)+test.outputString+" a Krishnamurthy Number", func(t *testing.T) {
```
tjgurwara99, line 34: You're already using the `Sprintf`, please use the verbs instead of string concatenating, this causes confusion to the reader.

tjgurwara99 (changes requested):

math/krishnamurthy.go
GooMonk: I was thinking about the overflow error it would give for the larger numbers but it would not make sense for the user to store large numbers in the low bounded int types. I will use the contrainst.Integer again and also create work on the other suggestions as mentioned by you. Thank you for the insight.

math/krishnamurthy_test.go
GooMonk: Hi Taj, please review the latest changes.

math/krishnamurthy_test.go
```diff
@@ -0,0 +1,51 @@
...
+		{3421488712, false, "is not"},
+	}
+}
+
+func TestIsKrishnamurthyNumber(t *testing.T) {
+	for _, test := range retCases() {
+		t.Run(fmt.Sprintf("%d %s a Krishnamurthy Number", test.input, test.outputString), func(t *testing.T) {
+			res := IsKrishnamurthyNumber(test.input)
+			if res != test.output {
+				t.Errorf("For input %d ,Expected: %t ,Found: %t", test.input, test.output, res)
```
tjgurwara99, line 37: This is minor and I'm approving it anyways:

The error strings in Go are usually kept lowercase since we can wrap these errors into other errors and it doesn't make sense to have uppercase words in the middle (unless we're talking about nouns).

tjgurwara99:
```suggestion
				t.Errorf("for input %d, expected: %t, found: %t", test.input, test.output, res)
```

Like so ^

Also, the comma's are in the wrong place.

tjgurwara99 (approved):

tjgurwara99 (approved):

siriak (approved):
Looks good, thanks!

PR #624, opened 2023-02-11
fix: fix deprecated rand.Seed()

hidenari-yuda:
rand.Seed() has been deprecated since Go 1.20, though this project is 1.19. But programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. 
To avoid such breakages, Programs that need a specific deterministic sequence can do that more reliably using NewRand(NewSource(seed)), which they can be sure no other packages are consuming values from.

### DIFF
--- a/cipher/caesar/caesar_test.go
+++ b/cipher/caesar/caesar_test.go
@@ -156,10 +156,10 @@ func Example() {
 }
 
 func FuzzCaesar(f *testing.F) {
-	rand.Seed(time.Now().Unix())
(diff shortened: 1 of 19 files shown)


siriak (approved):
Looks good, thanks!

raklaptudirm (approved):

tjgurwara99:
We are on `1.19` right now, which means this merge causes the random number to not correct anymore. I think we should have waited until we moved to `1.20` before merging this 😅

siriak:
Let's do this sooner then :)

PR #625, opened 2023-02-13
feat: adds Huffman coding algorithm

pedromsrocha:
Implements Huffman compression, associated encoding and decoding algorithms. 

The implementation follows the technique described [here](https://en.wikipedia.org/wiki/Huffman_coding#Basic_technique), in particular it implements the described compression variant that uses two queues to build the Huffman tree,  where the the given input symbols are already sorted by frequency. 

I've also added some basic tests that check that the compression is lossless, i.e. that the string obtained by encoding and then decoding is the same string we started with.


siriak (dismissed):
Looks good, thanks!

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
...
+	symbol rune
+	freq   int
+}
+
+// The algorithm operates on []SymbolFreq sorted by frequency
+type ByFreq []SymbolFreq
+
+func (x ByFreq) Len() int           { return len(x) }
+func (x ByFreq) Less(i, j int) bool { return x[i].freq < x[j].freq }
+func (x ByFreq) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
```
tjgurwara99, line 34: From what I can tell, you're making an assumption that the provided list of symbol-frequency is already sorted, therefore I don't see why you this type has to satisfy the `sort.Interface`. And even more so when you're not even importing the `sort` package....

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
...
+		node2, q1, q2 = least(q1, q2)
+		node := Node{left: &node1, right: &node2,
+			symbol: -1, weight: node1.weight + node2.weight}
+		q2 = append(q2, node)
+	}
+	if len(q1) == 1 { // returns the remaining node in q1, q2
+		return q1[0]
+	} else {
+		return q2[0]
+	}
```
tjgurwara99, line 61:
```suggestion
	if len(q1) == 1 { // returns the remaining node in q1, q2
		return q1[0]
	}
	return q2[0]
```

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
...
+		return q2[0], q1, q2[1:]
+	}
+	if len(q2) == 0 {
+		return q1[0], q1[1:], q2
+	}
+	if q1[0].weight <= q2[0].weight {
+		return q1[0], q1[1:], q2
+	} else {
+		return q2[0], q1, q2[1:]
+	}
```
tjgurwara99, line 77: Simply your if else statements - the return statements inside an if statement means that the rest of the code is never going to be executed when `if` condition is met.
```suggestion
	if q1[0].weight <= q2[0].weight {
		return q1[0], q1[1:], q2
	}
	return q2[0], q1, q2[1:]
```

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
...
+	} else { // inductive step
+		prefixLeft := make([]bool, len(prefix))
+		copy(prefixLeft, prefix)
+		prefixLeft = append(prefixLeft, false)
+		BuildDict(node.left, prefixLeft, dict)
+		prefixRight := make([]bool, len(prefix))
+		copy(prefixRight, prefix)
+		prefixRight = append(prefixRight, true)
+		BuildDict(node.right, prefixRight, dict)
+	}
```
tjgurwara99, line 95: This type of scenario is what a guard statement is for. You already have what your need in the `if` block so return :smile:
```suggestion
	if node.symbol != -1 { //base case
		dict[node.symbol] = prefix
        return
	}
    // inductive step
    prefixLeft := make([]bool, len(prefix))
    copy(prefixLeft, prefix)
	prefixLeft = append(prefixLeft, false)
	BuildDict(node.left, prefixLeft, dict)
	prefixRight := make([]bool, len(prefix))
	copy(prefixRight, prefix)
	prefixRight = append(prefixRight, true)
	BuildDict(node.right, prefixRight, dict)
```

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
...
+		return q1[0], q1[1:], q2
+	} else {
+		return q2[0], q1, q2[1:]
+	}
+}
+
+// BuildDict recursively traverses the Huffman tree pointed by node to build dict, that maps symbols to binary codes,
+// each code being prefixed by prefix.
+// left and right children are labelled with the booleans false and true, respectively.
+func BuildDict(node *Node, prefix []bool, dict map[rune][]bool) {
```
tjgurwara99, line 83: Dict is a term that refers to a concept in `Python`, this is a Go repository. Try to name things using the Go conventions :pray: :smile:

Can you think of the actual name of this step of the code? Even the python [implementation](https://github.com/TheAlgorithms/Python/blob/67676c3b790d9631ea99c89f71dc2bf65e9aa2ca/compression/huffman.py#L69) is not using this particular step (from what I can tell - I skimmed it so maybe I missed something :smile:).
PS: dict is called a map in Go.

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
+// huffman.go
```
tjgurwara99, line 1: I believe having a separate directory for `compression` makes sense. I believe you should move this package inside a new `compression` package by creating a new directory `compression` in the root of this repository, and call that package `compression` (though feel free to create a new subpackage inside the `compression` package called `huffman` if that's easier - but in Go we try to keep similar things as close together as possible without the possibility of name clashes)

other/huffmancoding/huffmancoding_test.go
```diff
@@ -0,0 +1,53 @@
+// huffmancoding_test.go
+// description: Tests the compression, encoding and decoding algorithms of huffmancoding.go.
+// author(s) [pedromsrocha](https://github.com/pedromsrocha)
+// see huffmancoding.go
+
+package huffman
```
tjgurwara99, line 6: This is an algorithm that is quite involved for people unfamilar to this (I myself took quite some time to review this thoroughly), so I recomment creating blackbox testing by changing the package name to
```suggestion
package huffman_test
```

That way, people who want to understand the algorithm from a high level perspective, it would be easier, and it would also ensure that there is nothing in code that has leaky abstraction :smile:

other/huffmancoding/huffmancoding_test.go
```diff
@@ -0,0 +1,53 @@
...
+		sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+		Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+		aliquip ex ea commodo consequat.`,
+	}
+
+	for _, message := range messages {
+		t.Run("huffman: "+message, func(t *testing.T) {
+			tree := BuildTree(SymbolCountOrd(message))
+			dict := make(map[rune][]bool)
+			BuildDict(&tree, nil, dict)
```
tjgurwara99, line 44: I really think that the name of this function is quite bad... (just my personal opinion).

tjgurwara99 (changes requested):
Please review the comments that I have left in your PR. Overall this is quite promising but at the same time, I felt that there was a tendency to use other language concepts in Go :sweat_smile:. Maybe try to look at some well documented Go code to get a feel for the way we Go :wink: - anyways, I'd be happy to review any changes you make and answer any comments and consider your suggestions (whenever I get some time)

other/huffmancoding/huffmancoding_test.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, s := range message {
+		runeCount[s]++
+	}
+	listfreq := make(ByFreq, len(runeCount))
+	i := 0
+	for s, n := range runeCount {
+		listfreq[i] = SymbolFreq{symbol: s, freq: n}
+		i++
+	}
+	sort.Sort(listfreq)
```
tjgurwara99, line 25: Forgot to mention. Could you try to use the sorting algorithm that is inside this repository - we have tried very much to avoid the standard library sort package because I think if there is something that our sort functions are missing then it makes sense to adapt them to improve this repo 😄

other/huffmancoding/huffmancoding.go
```diff
@@ -0,0 +1,123 @@
...
+func (x ByFreq) Len() int           { return len(x) }
+func (x ByFreq) Less(i, j int) bool { return x[i].freq < x[j].freq }
+func (x ByFreq) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
+
+// BuildTree returns the root Node of the Huffman tree by compressing listfreq.
+// The compression produces the most optimal code lengths, provided listfreq is ordered,
+// i.e.: listfreq[i] <= listfreq[j], whenever i < j.
+func BuildTree(listfreq ByFreq) Node {
+	if len(listfreq) < 1 {
+		panic("huffman: calling BuildTree with an empty list of symbol-frequency pairs")
```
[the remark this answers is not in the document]
pedromsrocha: The commit 684980f replaces the panic call by returning an appropriate error.

other/huffmancoding/huffmancoding.go
pedromsrocha: I've opted to use the term `dict` in accordance with the terminology used in the [wiki page](https://en.wikipedia.org/wiki/Huffman_coding) of the Huffman coding algorithm, on which this implementation is based. However, I understand that this might violate Go conventions, so commit 7193e6e renames method `BuldDitct` to - I hope - a more descriptive name `GetEncoding`. This corresponds to the step of the algorithm in which the Huffman tree is traversed (a kind of tree flattening) to obtain the encoding map.

other/huffmancoding/huffmancoding_test.go
pedromsrocha: I guess you're referring to `BuildDict`. I've already changed it to `GetEncoding`.

other/huffmancoding/huffmancoding.go
pedromsrocha: You're completely right! Commit 318d135 moves the sort interface method instantiation to the test file, the place where it's used.

compression/huffmancoding_test.go
```diff
@@ -0,0 +1,60 @@
+// huffmancoding_test.go
+// description: Tests the compression, encoding and decoding algorithms of huffmancoding.go.
+// author(s) [pedromsrocha](https://github.com/pedromsrocha)
+// see huffmancoding.go
+
+package compression
```
tjgurwara99, line 6: Could you use the test package instead?
```suggestion
package compression_test
```

compression/huffmancoding_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"sort"
+	"testing"
+)
+
+// The algorithm operates on []SymbolFreq sorted by frequency
+type ByFreq []SymbolFreq
+
+func (x ByFreq) Len() int           { return len(x) }
+func (x ByFreq) Less(i, j int) bool { return x[i].freq < x[j].freq }
+func (x ByFreq) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
```
tjgurwara99, line 18: If you were to use the `sort.Slice` you don't need to implement the `sort.Interface` though I would much rather prefer if you use this repositories sorting functions - might have to modify them in a similar way to the standard library but I think it might also be possible to completely avoid it.

other/huffmancoding/huffmancoding_test.go
pedromsrocha: Done with 6cd8d3a. I rename the functions of the Huffman API to avoid future clashes as new compression methods are added to the package.

compression/huffmancoding_test.go
pedromsrocha: Thanks for your suggestion of using sort.Slice instead, commit 4dc9b7d does that.

I think that the sorting functions in this repository's sort package apply to slices of constraints.Ordered elements, which are only basic datatypes (Int, String, Float), whereas in this case I need to sort a slice of pairs. I don't know which changes these would imply to the sort package.

raklaptudirm (approved):

PR #626, opened 2023-03-03
feat: add dutchNationalFlagSort implementation

romankurnovskii:
#### Checklist

Notes:


sort/dutchsationalslagsort.go
```diff
@@ -0,0 +1,25 @@
+// Implementation of Dutch National Flag Sort algorithm
+// Reference: https://en.wikipedia.org/wiki/Dutch_national_flag_problem
+
+package sort
+
+import "github.com/TheAlgorithms/Go/constraints"
+
+// DutchNationalFlagSort sorts an array of integers with the Dutch National Flag algorithm
+func DutchNationalFlagSort[T constraints.Signed](arr []T) []T {
```
tjgurwara99, line 9: Sort is in the name of the package - calling it `sort.DutchNationalFlag` suffices.

sort/dutchsationalslagsort_test.go
```diff
@@ -0,0 +1,34 @@
+package sort
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDutchNationalFlagSort(t *testing.T) {
```
tjgurwara99, line 8: There is already a [testing framework](https://github.com/TheAlgorithms/Go/blob/4f43cc154fccf98cccd49c1b1fdba21cba8e4a5f/sort/sorts_test.go#L11) in this package to test sorting algorithms matching the sort function signature, you only need to provide it with your function and test on that.


All you need is this:

```go
func TestDutchNationalFlag(t *testing.T) {
	testFramework(t, sort.DutchNationalFlag[int])
}
```

I would also prefer to keep this test closer to all the other tests in the `sort_test.go` file - since that is the convention we have used for this package but moving this to that file is a non-blocker.

sort/dutchsationalslagsort_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := DutchNationalFlagSort(tc.input)
+			if !reflect.DeepEqual(actual, tc.expected) {
+				t.Errorf("Expected %v, but got %v", tc.expected, actual)
+			}
+		})
+	}
+}
```
tjgurwara99, line 34: Please also write a benchmark for this!

tjgurwara99 (changes requested):

sort/dutchsationalslagsort.go
romankurnovskii: fixed, thank you

sort/dutchsationalslagsort_test.go
romankurnovskii: added

sort/dutchnationalflagsort_test.go
```diff
@@ -0,0 +1,34 @@
+package sort
```
tjgurwara99, line 1:
```suggestion
package sort_test
```

sort/dutchsationalslagsort.go
tjgurwara99, line 9: This algorithm seems more a category classification algorithm than a sorting algorithm (even though it looks very similar to quick sort, I think). It might be worth it moving to the `other` package instead of adding it into the `sort` package instead. What do you think @raklaptudirm @siriak ??

siriak: I agree, it can only sort arrays with 3 distinct elements, so I wouldn't call it a sorting algorithm

PR #634, opened 2023-04-28
feat: add cycle sort algorithm

PalAditya:
This adds the cycle sort algorithm. Link for reference: https://en.wikipedia.org/wiki/Cycle_sort

### DIFF
--- a/sort/cyclesort.go
+++ b/sort/cyclesort.go
@@ -14,7 +14,7 @@ func Cycle[T constraints.Number](arr []T) []T {
 	if len <= 1 {
 		return arr
 	}
-	
+
 	for cycle = 0; cycle < len-1; cycle++ {
 		elem := arr[cycle]
 		// Find total smaller elements to right
--- a/sort/cyclesort.go
+++ b/sort/cyclesort.go
@@ -10,12 +10,11 @@ import (
 func Cycle[T constraints.Number](arr []T) []T {
 
 	counter, cycle, len := 0, 0, len(arr)
-
 	// Early return if the array too small
 	if len <= 1 {
 		return arr
 	}
-
+	
 	for cycle = 0; cycle < len-1; cycle++ {
 		elem := arr[cycle]
 		// Find total smaller elements to right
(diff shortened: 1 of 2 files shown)


sort/cyclesort.go
```diff
@@ -0,0 +1,67 @@
...
+			// Ignore all duplicate elements, if any still appear
+			for elem == arr[pos] {
+				pos++
+			}
+
+			//We can do this unconditionally, but the check helps prevent redundant writes to the array
+			if elem != arr[pos] {
+				arr[pos], elem = elem, arr[pos]
+			}
+		}
```
tjgurwara99, line 63: You have too many newlines separating sections that shouldn't be separated. Please can you remove those unnecessary newlines.

tjgurwara99 (changes requested):
Good comments in the code! I usually always write to people to explain the question of "why you're doing" in comments instead of "what you're doing" since the latter can be explained by the code itself. You seem to have a correct balance (except for a few places).

sort/cyclesort.go
PalAditya: @tjgurwara99 - Thanks for the comment. I was separating all comments via newlines, but it indeed was a bit long. Have now used them for distinct segments only - initialization, first loop, nested loop and the return. Hopefully that makes it more concise.

tjgurwara99: Just a request, could you try and not tag people unless it's really necessary for a conversation 🙏🏼 I say this because I would have reviewed this PR again without the tag/mention which creates an unnecessary notification noise when working 😅

sort/cyclesort.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+// Cycle sort is an in-place, unstable sorting algorithm that is particularly useful
+// when sorting arrays containing elements with a small range of values. It is theoretically
+// optimal in terms of the total number of writes to the original array.
+func Cycle[T constraints.Number](arr []T) []T {
+
+	counter, cycle, len := 0, 0, len(arr)
```
tjgurwara99, line 12:
```suggestion
	counter, cycle, len := 0, 0, len(arr)
```

tjgurwara99: This is a small nit so I'm committing - please make a pull on your fork to reflect these changes in your local 🙏🏼

tjgurwara99 (approved):

tjgurwara99:
@siriak @raklaptudirm since I added a commit to the PR, could you both review this as well 🙏🏼

siriak (approved):
Looks good, thanks!

raklaptudirm (approved):

PR #638, opened 2023-05-29
Check if godocmd runs

siriak:


vil02:
it might be a good idea to add `workflow_dispatch:` into the `on:` section - then one can trigger the workflow manually.

PR #647, opened 2023-08-24
Add fast inverse square root algorithm

yan-aint-nickname:
Almost identical to C implementation but with golang specific unsafe package
https://en.wikipedia.org/wiki/Fast_inverse_square_root


siriak (dismissed):
Looks good, thanks!

raklaptudirm (changes requested):
Please make the changes to the comments as I suggested in the discussion.

tjgurwara99:
I think this already exists in a different form - https://github.com/TheAlgorithms/Go/blob/6429dfdb87ad0b2ae462f42f5307b6c0bcaca5c4/math/binary/sqrt.go

yan-aint-nickname:
@tjgurwara99 there is some problem with the algorithm what you send. `1/y division` is very expensive operation that's why `right-shift >>` is used in floating point bit level hacking line. It's better to multiply by passed value. This algorithm was made to [reduce divisions](https://en.wikipedia.org/wiki/Fast_inverse_square_root#Motivation).  And I return `1/y` value not a `y`, because it's reverse square root. I'm going to clear comments @raklaptudirm , I just wanted to pay homage to the original algorithm. May I change `what the f***` to `what the hell`? If you have any questions, please ask

tjgurwara99:
I was in a hurry so I didn't write a descriptive comment but I pointed out the other implementation not because I want you to remove your implementation but rather do it without `unsafe`. As you can see from that function it can be done!!

Furthermore, if you look closely, it is almost identical to your function so if you move it (and name it appropriately, eg `InverseSqrt` or something) and change the other function to be something like `func Sqrt(n float32) float32 { return 1 / InverseSqrt(x) }` you solved two problems: 1. Simple is better than complex, and 2. Keeping similar logical blocks together in the same package.

yan-aint-nickname:
> I was in a hurry so I didn't write a descriptive comment but I pointed out the other implementation not because I want you to remove your implementation but rather do it without `unsafe`. As you can see from that function it can be done!!
> 
> Furthermore, if you look closely, it is almost identical to your function so if you move it (and name it appropriately, eg `InverseSqrt` or something) and change the other function to be something like `func Sqrt(n float32) float32 { return 1 / InverseSqrt(x) }` you solved two problems:
> 
> 1. Simple is better than complex by removing the unsafe package, and
> 2. Keeping similar logical blocks together in the same package.

`Float32bits` it's just an alias for `*(*int32)(unsafe.Pointer(&y))` what I did wrong that I used `int32` instead of `uint32`
https://github.com/golang/go/blob/master/src/math/unsafe.go#L12

So, I need to:
1. Rename `Q_rsqrt` to `FastInverseSqrt`
2. Move `FastInverseSqrt` to math/binary package
3. Replace `unsafe` with simple `Float32bits` and `Float32frombits`
4. Replace Sqrt in math/binary/sqrt.go with `func Sqrt(n float32) float32 { return 1 / FastInverseSqrt(x) }`
5. Improve comments for better godoc generation

Am I right?

tjgurwara99:
> So, I need to:
>
> Rename Q_rsqrt to FastInverseSqrt
> Move FastInverseSqrt to math/binary package
> Replace unsafe with simple Float32bits and Float32frombits
> Replace Sqrt in math/binary/sqrt.go with func Sqrt(n float32) float32 { return 1 / FastInverseSqrt(x) }
> Improve comments for better godoc generation
> Am I right?

Yes 😄

math/binary/fast_inverse_sqrt.go
```diff
@@ -0,0 +1,29 @@
+// Calculating the inverse square root
+// [See more](https://en.wikipedia.org/wiki/Fast_inverse_square_root)
+package binary
+
+import (
+	"math"
+)
+
+// Assumes that number always positive
```
tjgurwara99, line 9: A minor nit, Go's documenting comments are supposed to be written in a specific way, please refer to the documenting [tips here](https://tip.golang.org/doc/comment).
Eg:

```suggestion
// FastInverseSqrt assumes that number always positive
```

math/binary/fast_inverse_sqrt.go
```diff
@@ -0,0 +1,29 @@
+// Calculating the inverse square root
+// [See more](https://en.wikipedia.org/wiki/Fast_inverse_square_root)
+package binary
```
tjgurwara99, line 3: This will become the package documentation comment which already exists - you can see it [here](https://pkg.go.dev/github.com/TheAlgorithms/Go@v0.0.3-alpha/math/binary). To keep the same comment but not add it to the documentation, you would need to add a newline between the package declaration and the comments. Eg:

```suggestion
// Calculating the inverse square root
// [See more](https://en.wikipedia.org/wiki/Fast_inverse_square_root)

package binary
```

tjgurwara99 (dismissed):

siriak (approved):
Looks good, thanks!

tjgurwara99 (approved):
LGMT

PR #652, opened 2023-09-06
Added median algorithm

domo536:
Hey, I would like to contribute the median algorithm.

Thank you!

file added: median.go and median_test.go


domo536:
sorry for testing only

PR #655, opened 2023-09-19
add bogo sort

suhailmalik07:
This PR adds bogo sort.


suhailmalik07:
Wasn't able to add it into sorts_test.go as it takes a lot time even for 20 elements array (tried 120s, got timeout).

Let me know if anything better can be done.

tjgurwara99:
Just passing by, will review it later but I think you can still write the test for it and add the `t.Skip()` call in the beginning of the test to skip it. Bogo sort is one of those algorithms that's fun to think about but not really good for CI and their associated compute power spent.

siriak (approved):
Looks good, thanks!

siriak:
@tjgurwara99 could you review?

tjgurwara99 (approved):

PR #656, opened 2023-09-26
feat: add set implementation using generic

nikhrom:
Modified implementation of set using generics instead of any

Hello, if there is any mistake, please correct me, thanks!


raklaptudirm (changes requested):
Format your code with `gofmt`

siriak (approved):

raklaptudirm (approved):

PR #658, opened 2023-09-29
optimized stack array implementation 

11-aryan:
- Optimized the push operation by appending to the last element of the array instead of copying the whole array
- Removed the global variable stackArray and created a struct StackArray for it
- Modified stack_test.go file to reflect these changes

### DIFF
--- a/structure/stack/stack_test.go
+++ b/structure/stack/stack_test.go
@@ -67,7 +67,7 @@ func TestStackLinkedList(t *testing.T) {
 
 // TestStackArray for testing Stack with Array
 func TestStackArray(t *testing.T) {
-	newStack := NewStack() 
+	newStack := NewStack()
 	t.Run("Stack With Array", func(t *testing.T) {
 
 		newStack.push(2)
--- a/structure/stack/stack_test.go
+++ b/structure/stack/stack_test.go
@@ -67,53 +67,52 @@ func TestStackLinkedList(t *testing.T) {
 
 // TestStackArray for testing Stack with Array
 func TestStackArray(t *testing.T) {
+	newStack := NewStack() 
 	t.Run("Stack With Array", func(t *testing.T) {
 
-		stackPush(2)
-		stackPush(3)
+		newStack.push(2)
+		newStack.push(3)
 
 		t.Run("Stack Push", func(t *testing.T) {
-			if !reflect.DeepEqual([]any{3, 2}, stackArray) {
-				t.Errorf("Stack Push is not work we expected %v but got %v", []any{3, 2}, stackArray)
+			if !reflect.DeepEqual([]any{2, 3}, newStack.elements) {
+				t.Errorf("Stack Push is not work we expected %v but got %v", []any{2, 3}, newStack)
 			}
 		})
 
-		pop := stackPop()
+		pop := newStack.pop()
 
 		t.Run("Stack Pop", func(t *testing.T) {
-
-			if stackLength() == 2 && pop != 3 {
+			if newStack.length() == 2 && pop != 3 {
 				t.Errorf("Stack Pop is not work we expected %v but got %v", 3, pop)
 			}
 		})
 
-		stackPush(2)
-		stackPush(83)
+		newStack.push(2)
+		newStack.push(83)
 
 		t.Run("Stack Peak", func(t *testing.T) {
-
-			if stackPeak() != 83 {
-				t.Errorf("Stack Peak is not work we expected %v but got %v", 83, stackPeak())
+			if newStack.peek() != 83 {
+				t.Errorf("Stack Peek is not work we expected %v but got %v", 83, newStack.peek())
 			}
 		})
 
 		t.Run("Stack Length", func(t *testing.T) {
-			if stackLength() != 3 {
-				t.Errorf("Stack Length is not work we expected %v but got %v", 3, stackLength())
+			if newStack.length() != 3 {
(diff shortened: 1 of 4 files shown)


11-aryan:
Hi, 
I've made the changes as mentioned. Will this be considered this as a part of the Hacktoberfest if accepted?

11-aryan:
Can the code style tests that failed be resolved by running gofmt -s -w stackarray.go ?

raklaptudirm:
Yes they can be fixed that way.

raklaptudirm (changes requested):
Please export all the methods and convert the tests into black box ones.

raklaptudirm:
In go, you can write tests either in the `packagename` package or the `packagename_test` package. Tests are written under the `packagename` package when you want to test internal functionality of a package, and under the `packagename_test` package when you want to test the public API of a package. In your case you are testing the public API (which is why I asked you to export all the methods) and therefore should be tested under the `packagename_test` package, also known as black box testing.

11-aryan:
Oh, got it! Thanks for the explanation

structure/stack/stackarray.go
```diff
@@ -9,31 +9,45 @@
 
 package stack
 
-var stackArray []any
+type StackArray struct {
```
raklaptudirm, line 12: Don't repeat the package name in an exported identifier. `stack.StackArray` has stack twice. `stack.Array` is shorter and gives the same idea.

structure/stack/stackarray.go
```diff
@@ -9,31 +9,45 @@
...
+	elements []interface{}
```
raklaptudirm, line 13: Use a generic type parameter instead of `any` (`interface{}`).

raklaptudirm (changes requested):

structure/stack/stack_test.go
```diff
@@ -7,23 +7,24 @@
 // author [Milad](https://github.com/miraddo)
 // see stackarray.go, stacklinkedlist.go, stacklinkedlistwithlist.go
 
-package stack
+package stack_test
 
 import (
 	"container/list"
+	st "github.com/TheAlgorithms/Go/structure/stack"
```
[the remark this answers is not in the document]
11-aryan: I thought it would be easier to use the short form. I'll fix it.

structure/stack/stackarray.go
11-aryan: Okay, I'll change stackArray to Array

structure/stack/stackarray.go
11-aryan: `
  type Array[T any] struct {
	  elements []T
  }
`
Will this be fine or should I create a separate struct for the allowed data types ?

raklaptudirm: A Stack Array should support any data type correct? So you should use `any`.

tjgurwara99: I think that the usage of `type Array[T any] struct { elements []T }` might be better. Consider this example:

```go
...
import "fmt"

type Array[T any] struct{ elements []T }

func main() {
	data := Array[any]{
		elements: []any{1, "a"},
	}
	fmt.Println(data)
}
...
```

Here we have the same usage as the current implementation with the added benefit that the user should be able to make the `Stack` a typed stack. Though overall I don't know whether any one way is better or worse than the other... 😄

raklaptudirm: @tjgurwara99 Thats what they are implementing rn? Not sure how your suggestion is different.

tjgurwara99: > @tjgurwara99 Thats what they are implementing rn? Not sure how your suggestion is different.

The only extra benefit for using the approach mentioned is that the user would have to choice to make the this stack a "typed" stack or keep the stack dynamic in nature.

For example

```
...
import "fmt"

type Array[T any] struct{ elements []T }

func Example() {
	data := Array[any]{
		elements: []any{1, "a"},
	}
	fmt.Printf("%T\n", data.elements)
	data2 := Array[int]{
		elements: []int{1, 2, 3},
	}
	fmt.Printf("%T\n", data2.elements)
	// Output:
	// []interface {}
	// []int
}
...
```

You can imagine that if a user only needs to use the `int` type inside the stack, then they won't need to do type casting and type checks like `d, ok := data.elements[0].(int)` all the time... Though arguably it is indeed a contrived example, so not sure how useful it would be out in the wild.

raklaptudirm: Yeah but thats what we discussed about implementing right? Is your suggestion any different from:

> Use a generic type parameter instead of `any` (`interface{}`).

> `type Array[T any] struct { elements []T }`
> Will this be fine or should I create a separate struct for the allowed data types ?

tjgurwara99: > Use a generic type parameter instead of `any` (`interface{}`).

Probably not, I guess there is some miscommunication here. I assumed when you mentioned the above, you were talking about just replacing the `interface{}` to `any`. And I got further confused by the following where I assumed that you wanted to keep the older implementation rather than what the author suggested...

> A Stack Array should support any data type correct? So you should use any.

Anyways, I think we're both on the same page which means my suggestion was no different from what was discussed 😄

raklaptudirm: Yeah seems the issue has been resolved :)

tjgurwara99 (approved):

raklaptudirm (approved):

PR #660, opened 2023-09-30
Add Cocktail Sort

migueldalberto:
The cocktail sort (also known as cocktail shaker sort, bidirectional bubble sort, shaker sort, ripple sort, shuffle sort, or shuttle sort) is a variant of the bubble sort algorithm. It extends the bubble sort by operating in both directions (from array's beginning to its end, and from array's end to its beginning).

I based the implementation on the [Wikipedia page about Cocktail Sort](https://en.wikipedia.org/wiki/Cocktail_shaker_sort).

Also added this sort on the `sort_tests.go`

### DIFF
--- a/sort/cocktailsort.go
+++ b/sort/cocktailsort.go
@@ -1,0 +1,40 @@
+// Implementation of Cocktail sorting
+// reference: https://en.wikipedia.org/wiki/Cocktail_shaker_sort
+
+package sort
+
+import "github.com/TheAlgorithms/Go/constraints"
+
+func Cocktail[T constraints.Integer](arr []T) []T {
+	if len(arr) == 0 { // guard clause for 0 length arrays
+		return arr
+	}
+
+	swapped := false
+
+	for ok := true; ok; ok = swapped {
+		for i := 0; i < len(arr)-1; i++ { // loop from array's beginning to end
+			if arr[i] > arr[i+1] { // swap unsorted elements
+				arr[i], arr[i+1] = arr[i+1], arr[i]
+				swapped = true
+			}
+		}
+
+		if !swapped {
+			// if swapped == false, first loop didn't find any unsorted elements.
+			// array is sorted
+			break
+		}
+
+		swapped = false
+
+		for i := len(arr) - 1; i > 0; i-- { // loop from array's end to beginning
+			if arr[i] < arr[i-1] { // swap unsorted elements
+				arr[i], arr[i-1] = arr[i-1], arr[i]
+				swapped = true
+			}
+		}
+	}
+
+	return arr
+}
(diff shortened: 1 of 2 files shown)


raklaptudirm (commented):
Add some inline comments explaining the algorithm as far as possible.

sort/cocktailsort.go
```diff
@@ -0,0 +1,41 @@
...
+// Cocktail sort is a variation of bubble sort, operating in two directions (beginning to end, end to beginning)
+func Cocktail[T constraints.Integer](arr []T) []T {
+	if len(arr) == 0 { // ignore 0 length arrays
+		return arr
+	}
+
+	swapped := false // true if swapped two or more elements in the last loop
+	// if it loops through the array without swapping, the array is sorted
+
+	for ok := true; ok; ok = swapped { // exits when swapped == false
```
raklaptudirm, line 17: Isn't this whole mess just equivalent to:
```go
for swapped {
```
Not sure what I am missing.

I guess that's to ensure the loop doesn't break in the first iteration? That can be fixed by simply setting the initial value of `swapped` to `true`.

sort/cocktailsort.go
```diff
@@ -0,0 +1,41 @@
...
+
+		for i := 0; i < len(arr)-1; i++ { // first loop, from array's beginning to end
```
raklaptudirm, line 19: You probably need to set swapped to false before this loop.

raklaptudirm (changes requested):
I would suggest adding `start` and `end` variables to make the sort more efficient.

sort/cocktailsort.go
migueldalberto: I thought about doing that but preferred the way I did instead. Maybe a more readable solution is to just put in a `for {}` with another if exit at the end.

migueldalberto:
> I would suggest adding `start` and `end` variables to make the sort more efficient.

Added

sort/cocktailsort.go
raklaptudirm: This code:
https://github.com/TheAlgorithms/Go/blob/4c573a955c61932ec1513d90f92f5177636fa3cb/sort/cocktailsort.go#L52
Is equivalent to:
```go
for swapped {
```
It just uses more code and locs, so I am not sure why you think its better.

migueldalberto: Done.

raklaptudirm (approved):

sort/cocktailsort.go
```diff
@@ -0,0 +1,54 @@
+// Implementation of Cocktail sorting
+// reference: https://en.wikipedia.org/wiki/Cocktail_shaker_sort
+
+package sort
+
+import "github.com/TheAlgorithms/Go/constraints"
+
+// Cocktail sort is a variation of bubble sort, operating in two directions (beginning to end, end to beginning)
+func Cocktail[T constraints.Integer](arr []T) []T {
```
tjgurwara99, line 9: From my understanding, isn't this algorithm applicable to all `Ordered` types?

migueldalberto: Right.

tjgurwara99 (approved):

raklaptudirm (approved):

PR #662, opened 2023-09-30
feat: Implement Matrix , its Methods and Some Functions including Strassen Matrix Multiplication in Go

### LINKED ISSUE #661: [ENHANCEMENT]: Strassen Matrix Multiplication Implementation in Go
### What would you like to share?

**Strassen Matrix Multiplication** is an algorithm for multiplying two matrices that uses a **divide-and-conquer approach** to reduce the number of basic multiplications required, making it more efficient than the standard matrix multiplication algorithm, especially for large matrices. 

It reduces the time complexity to _O(n^log2(7))_ compared to naive approach _(O(n^3))_.

### Extra issue details

Refactoring the file and folders for better structure.
- Adding the code file in math/matrix folder    

### Additional information

_No response_

mohit07raghav19:
### **Title**: Implementation of Matrix and Associated Methods in Go

**Description**:

This pull request introduces several enhancements to the math/matrix directory:

- **Matrix Data Type**: Implemented a new Matrix data type in Go, providing a foundation for matrix operations.
- **Matrix Methods**: Developed associated methods for the Matrix data type, enabling operations such as addition, subtraction, and multiplication.
- **Strassen's Algorithm**: Included an efficient implementation of Strassen's Matrix Multiplication algorithm.
- **Error Handling**: Integrated robust error handling for matrix operations to manage incompatible inputs effectively.
- **Concurrency Utilization**: Leveraged Go's concurrency features for improved performance on large matrix computations.
- **Test Coverage**: Added comprehensive unit tests to ensure the correctness of the matrix operations and Strassen's algorithm.

Closes #661 (if applicable)


math/matrix/strassenmatrixmultiply.go
```diff
@@ -0,0 +1,138 @@
...
+// details:
+// This program takes two matrices as input and performs matrix multiplication
+// using the Strassen algorithm, which is an optimized divide-and-conquer
+// approach. It allows for efficient multiplication of large matrices.
+// author(s): Mohit Raghav(https://github.com/mohit07raghav19)
+// See strassenmatrixmultiply_test.go for test cases
+package matrix
+
+// Perform matrix multiplication using Strassen's algorithm
+func strassenMatrixMultiply(A, B [][]int) [][]int {
```
raklaptudirm, line 12: This function should probably be exported. Rename it to `strassenMatrixNultiply`.

raklaptudirm, line 12: Instead of just supporting matrices of type `[][]int`, you should parameterize the numeric type and work on `[][]T`, where `T` is a type parameter which implements `constraints.Integer`.

math/matrix/strassenmatrixmultiply.go
```diff
@@ -0,0 +1,138 @@
...
+				C[i][j] = C22[i-mid][j-mid]
+			}
+		}
+	}
+
+	return C
+}
+
+// Add two matrices element-wise
+func addMatrices(A, B [][]int) [][]int {
```
raklaptudirm, line 94: YOu should probably add this as a separate function in the `math/matrix` package. Same for the `subtractMatrices` and the `standardMatrixMultiply` function. All of them should be exported.

raklaptudirm, line 94: You should at least assert (panic on fail) that the matrices are compatible for addition, even if you are not returning an error.

math/matrix/strassenmatrixmultiply_test.go
```diff
@@ -0,0 +1,91 @@
+package matrix
```
raklaptudirm, line 1: Tests should be black box, i.e. under the package `matrix_test` cause you don't want to deal with internal stuff here.

math/matrix/strassenmatrixmultiply_test.go
```diff
@@ -0,0 +1,91 @@
...
+		for j := 0; j < size; j++ {
+			if result[i][j] != expected[i][j] {
+				t.Errorf("Mismatch at position (%d, %d). Expected %d, but got %d.", i, j, expected[i][j], result[i][j])
+			}
+		}
+	}
+}
+
+// Generate a random matrix of the given size
+func makeRandomMatrix(size int) [][]int {
```
raklaptudirm, line 57: Try and use the [Go Fuzz Testing](https://go.dev/doc/tutorial/fuzz) system instead of generating your own random numbers.

math/matrix/strassenmatrixmultiply_test.go
```diff
@@ -0,0 +1,91 @@
...
+	for i := 0; i < size; i++ {
+		matrix[i] = make([]int, size)
+		for j := 0; j < size; j++ {
+			matrix[i][j] = rand.Intn(10) // Generate random integers between 0 and 9
+		}
+	}
+	return matrix
+}
+
+func TestMain(m *testing.M) {
```
raklaptudirm, line 68: Unnecessary to define a `TestMain`, we don't need a failure exit code.

math/matrix/strassenmatrixmultiply_test.go
```diff
@@ -0,0 +1,91 @@
...
+	matrixB := makeRandomMatrix(size)
+
+	b.ResetTimer() // Reset the benchmark timer before the actual benchmark
+	for i := 0; i < b.N; i++ {
+		_ = strassenMatrixMultiply(matrixA, matrixB)
+	}
+}
+
+// Print a matrix
+func printMatrix(matrix [][]int) {
```
raklaptudirm, line 87: Make this another function under the `math/matrix` maybe?

raklaptudirm (changes requested):

math/matrix/strassenmatrixmultiply_test.go
mohit07raghav19: @raklaptudirm , I am working to fix all the requested changes but I am stuck at writing Fuzz Testing for generic functions like
```
package matrix

import "github.com/TheAlgorithms/Go/constraints"

type Number interface {
	constraints.Integer
}

func AddMatrices[T Number](A, B [][]T) [][]T {
	n := len(A)
	result := make([][]T, n)

	for i := 0; i < n; i++ {
		result[i] = make([]T, n)
		for j := 0; j < n; j++ {
			result[i][j] = A[i][j] + B[i][j]
		}
	}

	return result
}
```
Would you please help me out .

raklaptudirm: Hmm, I guess it is a bit complicated to generate fuzzed matrices. If you are having trouble with this change, feel free to ignore it, the other changes are good enough to warrant a merge.

mohit07raghav19: Ohk @raklaptudirm , Can you please help me with **Unit Tests** for generic functions . I searched about it but couldn't understand it.

raklaptudirm: Well testing for a single generic case should be fine, so just test for `[][]int` and it should imply that all the other numeric types also work.

math/matrix/strassenmatrixmultiply.go
mohit07raghav19: Changed the `strassenMatrixMultiply` to `StrassenMatrixMultiply` which is now exported.

mohit07raghav19: Added Generic Type to function `StrassenMatrixMultiply`and parameterized it with [][]T, where T is a type parameter which implements `constraints.Integer`. Also made the helping functions type parameterized

math/matrix/strassenmatrixmultiply.go
mohit07raghav19: Added `AddMatrices`,`SubtractMatrices`,`MultiplyMatrices`,`PrintMatrix`and`EqualMatrix` in different files which all are exportable and type parameterized. Also added their respective __test.go_ files.

mohit07raghav19: Added checks for **compatibility** for all functions .

math/matrix/strassenmatrixmultiply_test.go
mohit07raghav19: Updated All the **__test.go_** files under package `matrix_test`.

math/matrix/strassenmatrixmultiply_test.go
mohit07raghav19: Added Appropriate Test functions.

math/matrix/strassenmatrixmultiply_test.go
mohit07raghav19: Removed the Unnecessary `TestMain` function.

math/matrix/strassenmatrixmultiply_test.go
mohit07raghav19: Added `PrintMatrix` function as another function in `printmatrix.go` file under the `math/matrix`

mohit07raghav19:
@raklaptudirm , I have made the required changes in the code. 
I request you to Verify the same.
Please let me know if I could do something related to it further.
Thank YOU :)

math/matrix/addmatrices.go
```diff
@@ -0,0 +1,27 @@
+package matrix
+
+import (
+	"errors"
+
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+type Number interface {
```
raklaptudirm, line 9: Why make a new interface? You can directly use `constraints.Integer` in the source.

math/matrix/equalmatrix.go
```diff
@@ -0,0 +1,15 @@
+package matrix
+
+func MatrixEqual[T comparable](mat1, mat2 [][]T) bool {
```
raklaptudirm, line 3: You seem to be using `MatrixEqual` and `reflect.DeepEqual`. Be consisted and use a single function for equality checks.

math/matrix/printmatrix.go
```diff
@@ -0,0 +1,30 @@
+package matrix
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// PrintMatrix prints a 2D matrix of any type.
+func PrintMatrix(matrix interface{}) {
```
raklaptudirm, line 9: Use the `any` keyword instead of `interface{}`. Also, `[][]any` would allow you to skip those type checks.

math/matrix/strassenmatrixmultiply.go
```diff
@@ -0,0 +1,120 @@
...
+		A10, _ := AddMatrices(A9, M6)
+		// Calculate result submatrices
+		C11, _ := SubtractMatrices(A8, M5)
+		C12, _ := AddMatrices(M3, M5)
+		C21, _ := AddMatrices(M2, M4)
+		C22, _ := SubtractMatrices(A10, M2)
+
+		// Combine submatrices into the result matrix
+		C := make([][]T, n)
+		// for i := 0; i < n; i++ {
```
raklaptudirm, line 93: Why is this code commented out?

raklaptudirm (changes requested):
Note the Go idiom that you should not repeat the package name. For example, in `matrix.AddMatrices`, you are using the word "Matrix" twice. `matrix.Add` is a much simpler name and it doesn't lose any information. Change the names accordingly.

Also, you could make some simple functions that find the lenght/breadth of a matrix, given you are trying to find that multiple times.

mohit07raghav19:
> Also, you could make some simple functions that find the length/breadth of a matrix, given you are trying to find that multiple times.

@raklaptudirm 
For getting the `Breadth` of a matrix (only valid matrix.),  
```
// Breadth returns the number of columns in a valid matrix.
func Breadth[T any](matrix [][]T) (int, error) {
	if err := IsValid(matrix); err != nil {
		return 0, err
	}

	if len(matrix) == 0 {
		return 0, nil // Return 0 if the matrix is empty and valid.
	}

	return len(matrix[0]), nil
}

```
in which  `IsValid` func is used as 
```
// IsValid checks if a matrix is valid, meaning all rows have the same number of columns.
func IsValid[T any](matrix [][]T) error {
	if len(matrix) == 0 {
		return nil // an empty matrix is considered valid
	}

	columns := len(matrix[0]) // Number of columns in the first row.

	for _, row := range matrix {
		if len(row) != columns {
			return errors.New("invalid matrix: rows have different numbers of columns")
		}
	}

	return nil
}
```
and `Len` func returns the length only if matrix is valid.
```
func Len[T any](matrix [][]T) (int, error) {
	if err := IsValid(matrix); err != nil {
		return 0, err
	}
	return len(matrix), nil
}
```

BUT now in many functions simple `len(matrix)` is used After verifying that it is a valid matrix. The code can become unnecessarily extended like for 
`result := make([][]T, len(matrix1))` , The code would be 
```
// checked for IsValid(matrix1) above.
    length , err := Len(matrix1) // There is _no need_ to check for a valid matrix again
    if err!=nil {
         result := make([][]T, length)
         // ... appropriate code
    }
```
Do I need to change them to `Len`  or `Breadth` ?

raklaptudirm:
Don't do a validity check in the `Length` or `Breadth` functions, let the user decide when to do that.

Even better, you could make matrix a type with a constructor that ensures that the number of columns in each row is the same.

math/matrix/add.go
```diff
@@ -0,0 +1,36 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+// Add adds the elements of the current matrix (m1) to another matrix (m2) element-wise and returns a new matrix.
+// The two matrices must have the same dimensions.
+// If the matrices have different dimensions or if an error occurs during addition, an error is returned.
+func Add[T constraints.Integer](matrix1, matrix2 *Matrix[T]) (*Matrix[T], error) {
```
raklaptudirm, line 13: Unless you are actually changing the inputs, using pointers for a small structure such as Matrix is unnecessary. Remember that the slice is already a pointer.

raklaptudirm (changes requested):

raklaptudirm:
Also fix the CI errors.

math/matrix/add.go
mohit07raghav19: Yes..Changed.👍

math/matrix/samedimensions.go
```diff
@@ -0,0 +1,18 @@
+package matrix
+
+import "fmt"
+
+// SameDimensions checks if two matrices have the same dimensions.
+func (m *Matrix[T]) SameDimensions(matrix2 Matrix[T]) (bool, error) {
```
tjgurwara99, line 6: I'm kind of on the fence about the naming of this method. Do you think that `MatchDimensions` or `VerifySameDimensions` are better names when reading the code through?

math/matrix/add.go
```diff
@@ -0,0 +1,39 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/TheAlgorithms/Go/constraints"
+)
+
+// Add adds the elements of the current matrix (m1) to another matrix (m2) element-wise and returns a new matrix.
+// The two matrices must have the same dimensions.
+// If the matrices have different dimensions or if an error occurs during addition, an error is returned.
+func Add[T constraints.Integer](matrix1, matrix2 Matrix[T]) (Matrix[T], error) {
```
tjgurwara99, line 13: You seem to be using both a mixture of methods and functions to do matrix operations. Should these be consistent do you think?

I say this because `Add` can just as easily be a method on matrices rather than a function in the package since you are using the `SameDimensions` method... So what I'm trying to say is that your implementation should be consistent so that users know what to expect where 😄

math/matrix/samedimensions.go
```diff
@@ -0,0 +1,18 @@
...
+	rows1 := m.rows
+	rows2 := matrix2.rows
+
+	columns1 := m.columns
+	columns2 := matrix2.columns
+
+	if rows1 == rows2 && columns1 == columns2 {
+		return true, nil
+	} else {
+		return false, fmt.Errorf("matrices have different dimensions: rows=%d vs columns=%d", rows1, columns1)
```
tjgurwara99, line 16: Your false scenario only has one type of failure, that is the dimensions don't match. Have you considered only sending back the `false` without any error type? `false` should be indicative of the error... no?

math/matrix/add.go
```diff
@@ -0,0 +1,39 @@
...
+	// Check if the matrices have the same dimensions.
+	sameDimensions, err := matrix1.SameDimensions(matrix2)
+	if err != nil {
+		return Matrix[T]{}, fmt.Errorf("invalid matrices: %v", err)
+	}
+
+	if !sameDimensions {
+		return Matrix[T]{}, errors.New("matrices are not compatible for addition")
+	}
```
tjgurwara99, line 22: Should `SameDimensions` really need to return an `error` along with `false`?

math/matrix/add.go
```diff
@@ -0,0 +1,39 @@
...
+	for i := range matrix1.elements {
+		for j := range matrix1.elements[i] {
+			// Perform addition element-wise and set the result in the new matrix.
+			sum := matrix1.elements[i][j] + matrix2.elements[i][j]
+			err := result.Set(i, j, sum)
+			if err != nil {
+				return Matrix[T]{}, fmt.Errorf("invalid matrices: %v", err)
+			}
+		}
+	}
```
tjgurwara99, line 36: Good opportunity to maybe add go routines here...

tjgurwara99 (commented):
I haven't reviewed it fully, so will come back to this later. But I've left some thoughts on the PR 😄

math/matrix/add.go
mohit07raghav19: I tried doing this but as I have generalised the Matrix type for `any`, I want `Add` and other mathematical operations to be performed for `[T constraints.Integer]` ? How do I implement it ?
Should I use `func (m *Matrix[int]) Add(other Matrix[int]) (Matrix[int], error)`

tjgurwara99: It's interesting to see that you've chosen `any` for the Matrix implementation. So I ask a question, what kind of "thing" would you store in a matrix for it to be useful? Should you create a matrix of the following type? Is it going to be useful?

```go
type MyType struct {
   something string
}
```

mohit07raghav19: No..Got it.So I should Change `any` to `constraints.Integer` . Right?

math/matrix/add.go
mohit07raghav19: Is this correct ? Should I apply it to similar scenarios ?
```
var wg sync.WaitGroup
for i := 0; i < m1.rows; i++ {
	wg.Add(1)
	go func(i int) {
		defer wg.Done()
		for j := 0; j < m1.columns; j++ {
			sum := m1.elements[i][j] + m2.elements[i][j]
			err := result.Set(i, j, sum)
			if err != nil {
				fmt.Println("Error:", err)
				return
			}
		}
	}(i)
}

wg.Wait()
```

math/matrix/add.go
tjgurwara99: Personally I'd like to create another constraint called `Arithmetic` in the place where we have `Integer` defined. This can include the complex numbers as well as floats might be very useful in the context of matrices. But `Integer` would indeed suffice for now 😄

math/matrix/add.go
mohit07raghav19: @tjgurwara99 , Please See the sample code with go-routines and suggest any improvements ,So that I can make the required changes in similar files for better optimization.

mohit07raghav19: Ohk..But as the `Set` function is defined as 
```
func (m Matrix[T]) Set(row, col int, val T) error {
	if row < 0 || row >= m.rows || col < 0 || col >= m.columns {
		return errors.New("index out of bounds")
	}

	m.elements[row][col] = val
	return nil
}
```
It gives error only when User tries to set the value which doesn't fit into the **matrix dimensions**.But as in `Add` func, it is assured that it will be called for valid values of i and j . So, there will not be any error here.However, I have to check for the error due to _Golang CI lint checkers._
Anyways, it can be implemented like 
```
 var wg sync.WaitGroup
    var mu sync.Mutex // Mutex for synchronization of error slice
    var errors []error // Slice to collect errors

    for i := 0; i < m1.rows; i++ {
        wg.Add(1)

        go func(i int) { 
            defer wg.Done()

            for j := 0; j < m1.columns; j++ {
                sum := m1.elements[i][j] + m2.elements[i][j]
                err := result.Set(i, j, sum)
                if err != nil {
                    mu.Lock()
                    errors = append(errors, err) // Collect errors in the slice
                    mu.Unlock()
                    return
                }
            }
        }(i)
    }

    wg.Wait()

    if len(errors) > 0 {
        return Matrix[T]{}, errors
}
```

tjgurwara99: Your approach is fine but it can still be improved. Specially the use of Mutex when you don't really need to use it (its fine if you go with this approach actually - I'm just giving you an idea  an alternative approach)

My thinking is that you don't need a `slice` there for errors. Why? In Go, we use the channels to communicate so if there is an error, we can communicate that something failed using an `error` channel 😄

```Go
var wg sync.WaitGroup
	var errCh = make(chan error)
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel() // Make sure it's called to release resources even if no errors

	for i := 0; i < m1.rows; i++ {
		wg.Add(1)

		go func(i int) {
			defer wg.Done()

			for j := 0; j < m1.columns; j++ {
				select {
				case <-ctx.Done():
					return
				default:
				}
				sum := m1.elements[i][j] + m2.elements[i][j]
				err := result.Set(i, j, sum)
				if err != nil {
					errCh <- err
					cancel()
					return
				}
			}
		}(i)
	}

	wg.Wait()

	if ctx.Err() != nil {
		return nil, <-errCh
	}
	return result, nil
```

I haven't tested it but you would need context to make sure that you cancel on the first error and finish all the work in other goroutines. The errCh is to get the actual error that occurred.

Also, I can see that it might get very difficult to manage the complexity so I don't mind if you use the `golang.org/x/sync/errgroup` package personally but other maintainers might disagree. I personally think `errgroup` package should have been part of the standard library but alas, it's not.

mohit07raghav19: @tjgurwara99 , I have tried implementing the way for the functions. But I find it difficult for implementing for `StrassenMatrixMultiply` func.

tjgurwara99: Don't worry about making that method concurrent. Matrix multiplication is never trivial specially when dealing with concurrency.

math/matrix/strassenmatrixmultiply.go
```diff
@@ -0,0 +1,101 @@
...
+		A2, _ := B11.Add(B22)
+		A3, _ := A21.Add(A22)
+		A4, _ := A11.Add(A12)
+		A5, _ := B11.Add(B12)
+		A6, _ := B21.Add(B22)
+		//
+		S1, _ := B12.Subtract(B22)
+		S2, _ := B21.Subtract(B11)
+		S3, _ := A21.Subtract(A11)
+		S4, _ := A12.Subtract(A22)
```
tjgurwara99, line 46: I don't think you should be ignoring these errors...

math/matrix/submatrix.go
```diff
@@ -0,0 +1,64 @@
...
+	subMatrix := New(numRows, numCols, zeroVal)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // Make sure it's called to release resources even if no errors
+
+	var eg errgroup.Group
+	var firstErr error
+
+	for i := 0; i < numRows; i++ {
+		i := i // Capture the loop variable for the goroutine
```
tjgurwara99, line 34: 😂 This has caused so many bugs for me before 😂

Fortunately there is already a kinda fix in go1.21 - [ref](https://github.com/golang/go/wiki/LoopvarExperiment) to the experiment which is tracking the issue.

mohit07raghav19: @tjgurwara99 , According to the references, i think we have to use line 34 or run `GOEXPERIMENT=loopvar go test` instead of `go test`? Or there is any other possible solution which i didn't get? as the change need to be made is not yet done and is expected to be done in go1.22 or later (which I have understood).

tjgurwara99: Oh sorry, I think I miscommunicated here - I didn't ask you to change anything here. Sorry if that confused you

math/matrix/strassenmatrixmultiply.go
```diff
@@ -9,93 +9,219 @@
...
 		// Combine subMatrices into the result matrix
 		var zeroVal T
 		C := New(n, n, zeroVal)
 
 		for i := 0; i < mid; i++ {
 			for j := 0; j < mid; j++ {
-				val, _ := C11.Get(i, j)
+				val, err := C11.Get(i, j)
+				if err != nil {
+					panic("copyMatrix.Set error: " + err.Error())
```
tjgurwara99, line 188: I don't know how I missed it - you can surely do something better than `panic` here.

mohit07raghav19: Yeah..I missed it too. Thanks for correction.🙂

math/matrix/add.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Create a new matrix to store the result.
+	var zeroVal T
+	result := New(m1.Rows(), m1.Columns(), zeroVal)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // Make sure it's called to release resources even if no errors
+
+	var eg errgroup.Group
```
tjgurwara99, line 24: Ok, so you're conflating two concepts into one and making a change that is not needed - you should either explicitly use the `context` package or `errgroup` package. Since you're using the `eg.Go()` if one of the `func`s returns an error, the whole execution chain stops. So essentially you have one implementation that is working as expected but you are also using context to do the exact same thing - making that kind of redundant. So if you're using the `errgroup` package, you do not need to have a `select` statement inside the for loop and if you're using the `select` statement in conjunction with using a context cancellation upon error. So do just one 😄 I preferred your approach without the errgroup package 😄

mohit07raghav19: Yeah,my mistake..Corrected it.

tjgurwara99 (approved):

tjgurwara99:
Well done! This is looking much better now 😄 

PS: you might need to make an empty commit - the update GoDoc action has stopped the execution of other actions for some reason 😓

raklaptudirm (approved):
A very good quality pr, LGTM :)

PR #663, opened 2023-09-30
feat: reactivate `horspool`

vil02:
#### Description of Change

There is some commented out code in [`horspool.go`](https://github.com/TheAlgorithms/Go/blob/58bb31e80f8347c91db049a09d32a4989c06b485/strings/horspool/horspool.go).
This PR _reactivates_ it by making all necessary changes.

When the given pattern is not present in the input string, I used a similar approach to the one in [./search](https://github.com/TheAlgorithms/Go/tree/master/search).

#### Checklist

Notes: 
_Reactivates_ [`horspool.go`](https://github.com/TheAlgorithms/Go/blob/58bb31e80f8347c91db049a09d32a4989c06b485/strings/horspool/horspool.go).

### DIFF
--- a/strings/horspool/horspool.go
+++ b/strings/horspool/horspool.go
@@ -1,95 +1,39 @@
+// Implementation of the
+// [Boyer–Moore–Horspool algorithm](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore%E2%80%93Horspool_algorithm)
+
 package horspool
 
-// User defined.
-// Set to true to read input from two command line arguments
-// Set to false to read input from two files "pattern.txt" and "text.txt"
-// const commandLineInput bool = false
+import "errors"
 
-// Implementation of Boyer-Moore-Horspool algorithm (Suffix based approach).
-// Requires either a two command line arguments separated by a single space,
-// or two files in the same folder: "pattern.txt" containing the string to
-// be searched for, "text.txt" containing the text to be searched in.
-// func main() {
-// 	if commandLineInput == true { // case of command line input
-// 		args := os.Args
-// 		if len(args) <= 2 {
-// 			log.Fatal("Not enough arguments. Two string arguments separated by spaces are required!")
-// 		}
-// 		pattern := args[1]
-// 		s := args[2]
-// 		for i := 3; i < len(args); i++ {
-// 			s = s + " " + args[i]
-// 		}
-// 		if len(args[1]) > len(s) {
-// 			log.Fatal("Pattern  is longer than text!")
-// 		}
-// 		fmt.Printf("\nRunning: Horspool algorithm.\n\n")
-// 		fmt.Printf("Search word (%d chars long): %q.\n", len(args[1]), pattern)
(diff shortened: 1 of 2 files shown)


raklaptudirm (changes requested):
Add some inline comments to the algorithm implementation explaining it as far as possible.

vil02:
@raklaptudirm I have added some comments and restructured the code a bit. Now it should be more _readable_.

strings/horspool/horspool.go
```diff
@@ -1,95 +1,60 @@
...
-// Set to false to read input from two files "pattern.txt" and "text.txt"
-// const commandLineInput bool = false
+import "errors"
+
+var ErrNotFound = errors.New("pattern was not found in the input string")
+
+func Horspool(t, p string) (int, error) {
+	// in order to handle multy-byte character properly
+	// the input is converted into rune arrays
+	return horspool([]rune(t), []rune(p))
```
raklaptudirm, line 13: Implementing the functionality in a separate function seems unnecessary. Just directly implement it here.

raklaptudirm (changes requested):

strings/horspool/horspool.go
raklaptudirm: Introducing `tRunes`/`pRunes` seems fine to me. @siriak what do you think?

siriak: I prefer the current implementation because it clearly assigns concerns to different functions. The strings function accepts input and converts it to a different format, while the runes function implements the algorithm. Otherwise, there would be `t, p, tRune and pRune` variables that span the whole method, but only `tRune and pRune` variables are used.

raklaptudirm (approved):

strings/horspool/horspool.go
tjgurwara99: I'm also kind of against this... but why not just use the `[]rune` on the main `Horspool` instead?

vil02: For some reason `horspool` is in `strings` directory. I think that `string` is a more _natural_ type to represent string than `[]rune` - I would be annoyed if I would need to convert my inputs to `[]rune` every single time when I want to use this function.

In this case it is a pity that go does not have overloading.

tjgurwara99: > I would be annoyed if I would need to convert my inputs to []rune every single time

I get annoyed at that too 😂 until there are points where the conversion makes sense (not just the `[]rune` conversion but `[]byte` too). In this case however, it's fine I don't mind it. I just didn't like another allocation happening for that function call but that is a very nitpicky 😂

tjgurwara99 (approved):

strings/horspool/horspool.go
vil02: > I just didn't like another allocation happening for that function call but that is a very nitpicky 😂

I guess the only solution for that would be to create a `SimpleString` class not allowing multi-byte characters or `TurboString`, which has just a `[]rune` wrapper.

I any case: I will mark this conversation as _resolved_. Thanks for interesting conversation.

siriak (approved):

PR #667, opened 2023-10-04
refactor: refactor comments and variable declaration compliant with godoc

himanshu-patel-dev:


tjgurwara99 (dismissed):

graph/topological.go
```diff
@@ -1,7 +1,7 @@
 package graph
 
-// Assumes that graph given is valid and possible to
-// get a topo ordering.
+// Topological Assumes that graph given is valid and possible to
```
raklaptudirm, line 3:
```suggestion
// Topological assumes that graph given is valid and that its possible to
```

raklaptudirm (changes requested):

tjgurwara99 (approved):

raklaptudirm (approved):

PR #672, opened 2023-10-06
Add `FieldsN` function to `strings` package.

celestix:
FieldsN splits a string into n fields.
It is similar to strings.Fields, but it splits the string into n fields.
If the string has less than n fields, the last field will contain the rest of the string.
If the string has more than n fields, the last field will contain the rest of the string.
If n is -1, it will behave like strings.Fields.

### DIFF
--- a/strings/fieldsn.go
+++ b/strings/fieldsn.go
@@ -20,14 +20,20 @@ func FieldsN(s string, n int) []string {
 	a := make([]string, n+1)
 	b := ""
 	index := 0
+	wf := false
 	for _, c := range s {
 		if index >= n || !unicode.IsSpace(c) {
 			b += string(c)
+			wf = false
+			continue
+		}
+		if wf {
 			continue
 		}
 		a[index] = b
 		index++
 		b = ""
+		wf = true
 	}
 	a[index] = b
 	return a[:index+1]
--- a/strings/fieldsn.go
+++ b/strings/fieldsn.go
@@ -4,7 +4,9 @@
 // see fields_test.go
 package strings
 
-import "strings"
+import (
+	"unicode"
+)
 
 // FieldsN splits a string into n fields.
 // It is similar to strings.Fields, but it splits the string into n fields.
@@ -13,25 +15,24 @@ import "strings"
 // If n is -1, it will behave like strings.Fields.
 func FieldsN(s string, n int) []string {
 	if n < 0 {
-		return strings.Fields(s)
+		n = len(s)
 	}
 	a := make([]string, n+1)
-	b := strings.Builder{}
+	b := ""
 	index := 0
 	for _, c := range s {
 		if index+1 > n {
-			b.WriteRune(c)
+			b += string(c)
 			continue
 		}
-		switch c {
-		case '\n', '\t', '\f', '\r', ' ':
-			a[index] = b.String()
-			index++
-			b.Reset()
+		if !unicode.IsSpace(c) {
+			b += string(c)
 			continue
 		}
-		b.WriteRune(c)
+		a[index] = b
+		index++
+		b = ""
 	}
-	a[index] = b.String()
+	a[index] = b
 	return a[:index+1]
 }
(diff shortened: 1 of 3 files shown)


strings/fieldsn.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	a := make([]string, n+1)
+	b := strings.Builder{}
+	index := 0
+	for _, c := range s {
+		if index+1 > n {
+			b.WriteRune(c)
+			continue
+		}
+		switch c {
```
raklaptudirm, line 26: Instead of a switch, `unicode.IsSpace(r)` should be better.

strings/fieldsn.go
```diff
@@ -0,0 +1,37 @@
...
+// It is similar to strings.Fields, but it splits the string into n fields.
+// If the string has less than n fields, the last field will contain the rest of the string.
+// If the string has more than n fields, the last field will contain the rest of the string.
+// If n is -1, it will behave like strings.Fields.
+func FieldsN(s string, n int) []string {
+	if n < 0 {
+		return strings.Fields(s)
+	}
+	a := make([]string, n+1)
+	b := strings.Builder{}
```
raklaptudirm, line 19: A string builder seems overkill in this situation, just a simple string should suffice.

strings/fieldsn.go
```diff
@@ -0,0 +1,37 @@
...
+import "strings"
+
+// FieldsN splits a string into n fields.
+// It is similar to strings.Fields, but it splits the string into n fields.
+// If the string has less than n fields, the last field will contain the rest of the string.
+// If the string has more than n fields, the last field will contain the rest of the string.
+// If n is -1, it will behave like strings.Fields.
+func FieldsN(s string, n int) []string {
+	if n < 0 {
+		return strings.Fields(s)
```
raklaptudirm, line 16: Change the algorithm so that you don't depend on the standard library implementation.

raklaptudirm (changes requested):

souhardya79:
I want to contribute in this issue

celestix:
@raklaptudirm I've made the changes you asked, please check if it is fine now.

strings/fieldsn.go
```diff
@@ -0,0 +1,38 @@
...
+// If the string has less than n fields, the last field will contain the rest of the string.
+// If the string has more than n fields, the last field will contain the rest of the string.
+// If n is -1, it will behave like strings.Fields.
+func FieldsN(s string, n int) []string {
+	if n < 0 {
+		n = len(s)
+	}
+	a := make([]string, n+1)
+	b := ""
+	index := 0
```
raklaptudirm, line 22: Instead of an index variable, you can just use the first value from the range expression.

strings/fieldsn.go
```diff
@@ -0,0 +1,38 @@
...
+	for _, c := range s {
+		if index+1 > n {
```
raklaptudirm, line 24: This is equivalent to `index >= n`

strings/fieldsn.go
```diff
@@ -0,0 +1,38 @@
...
+			b += string(c)
+			continue
+		}
+		if !unicode.IsSpace(c) {
```
raklaptudirm, line 28: Both the if statements have the same body, so you should merge the conditions with a logical or `||`.

raklaptudirm (changes requested):

strings/fieldsn.go
celestix: that index variable is not the index for character element in the string, It's incremented on every whitespace.

strings/fieldsn.go
celestix: Sure!

celestix:
Hey @raklaptudirm, please check now.

raklaptudirm (changes requested):
The code doesn't properly handle cases where there are multiple spaces between fields and cases where there are leading and trailing spaces.

celestix:
@raklaptudirm Thanks for pointing out the multiple whitespaces issue! I've fixed it and added few testcases for those conditions too, please check if everything is fine now or not.

raklaptudirm (changes requested):
Instead of the wf variable, you could just check if `b == ""`.

Also note the if `NumFields(s) <= n` then also `len(n) == n + 1`, which is probably unintended.

PR #689, opened 2023-10-19
feat: graph cycle detection

kiarash8112:
#### Description of Change

i added cycle detection in graph package and other function that list all detected cycles in graph also added tests and docs
#### Checklist

Notes:

### DIFF
--- a/graph/cycle.go
+++ b/graph/cycle.go
@@ -1,0 +1,40 @@
+package graph
+
+func (g *Graph) HasCycle() bool {
+
+	all := map[int]struct{}{}
+	visiting := map[int]struct{}{}
+	visited := map[int]struct{}{}
+
+	for v := range g.edges {
+		all[v] = struct{}{}
+	}
+
+	for current := range all {
+		if g.HasCycleHelper(current, all, visiting, visited) {
+			return true
+		}
+	}
+
+	return false
+
+}
+
+func (g Graph) HasCycleHelper(v int, all, visiting, visited map[int]struct{}) bool {
+	delete(all, v)
+	visiting[v] = struct{}{}
+
+	neighbors := g.edges[v]
+	for v := range neighbors {
+		if _, ok := visited[v]; ok {
+			continue
+		} else if _, ok := visiting[v]; ok {
+			return true
+		} else if g.HasCycleHelper(v, all, visiting, visited) {
+			return true
+		}
+	}
+	delete(visiting, v)
+	visited[v] = struct{}{}
+	return false
+}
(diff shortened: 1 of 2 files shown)


graph/cycle.go
```diff
@@ -0,0 +1,119 @@
...
+		if g.HasCycleHelper(current, all, visiting, visited) {
+			return true
+		}
+	}
+
+	return false
+
+}
+
+func (g Graph) HasCycleHelper(v int, all, visiting, visited map[int]struct{}) bool {
```
tjgurwara99, line 30:
```suggestion
func (g Graph) hasCycleHelper(v int, all, visiting, visited map[int]struct{}) bool {
```

This function is not needed by anything outside of this package and as such, should not be exported.

graph/cycle.go
```diff
@@ -0,0 +1,119 @@
...
+
+		}
+
+	}
+
+	return allCycles
+
+}
+
+func (g Graph) FindAllCyclesHelper(current int, all, visiting, visited map[int]struct{}) (bool, [][]int) {
```
tjgurwara99, line 86: Same for this.
```suggestion
func (g Graph) findAllCyclesHelper(current int, all, visiting, visited map[int]struct{}) (bool, [][]int) {
```

graph/cycle.go
```diff
@@ -0,0 +1,119 @@
...
+			parents = append(parents, []int{v, current})
+			return true, parents
+		}
+	}
+	delete(visiting, current)
+	visited[current] = struct{}{}
+	return false, parents
+}
+
+func ConvertToString(parents [][]int) string {
```
tjgurwara99, line 110: Cycles are graphs by themselves, as such I think you should return a fully fledged graph instead of strings when you detect cycles.

graph/cycle_test.go
```diff
@@ -0,0 +1,53 @@
...
+		graph.AddEdge(edge[0], edge[1])
+	}
+
+	res := graph.FindAllCycles()
+
+	if len(res) != 1 {
+		t.Error("number of cycles is not correct")
+	}
+
+	firstCycle := res[0]
```
tjgurwara99, line 40: Is this always going to be deterministic? The first cycle is going to always be the same?

tjgurwara99 (approved):

graph/cycle_test.go
kiarash8112: this test has only one cycle so It's vertices should be connected in just one way (graph is directed)

raklaptudirm (approved):

PR #699, opened 2023-12-10
feat: add Kosaraju Algoritm

11-aryan:


11-aryan:
@raklaptudirm Could you please review it ?

graph/kosaraju.go
```diff
@@ -0,0 +1,61 @@
...
+	*currentSCC = append(*currentSCC, node)
+	for _, nbr := range adj[node] {
+		if visited[nbr] {
+			continue
+		}
+		dfs(nbr, adj, visited, currentSCC)
+	}
+}
+
+func GetStronglyConnectedComponents(n int, adj [][]int) [][]int {
```
tjgurwara99, line 40: Could you add a comment here to describe that this is getting it through the Kosaraju algorithm

tjgurwara99, line 40: I think it would have been much more easier to understand this algorithm if you had used the `Graph` data structure defined in the `graph.go` file. There are already a lot of helpers for that graph as well.

tjgurwara99: If you don't need the weighted edges, then don't use the weighted edges part of the graph 😄.

PR #703, opened 2024-01-23
feat: Add btree with insertion, deletion, and search

caojoshua:
see https://en.wikipedia.org/wiki/B-tree

### DIFF
--- a/structure/tree/btree.go
+++ b/structure/tree/btree.go
@@ -73,7 +73,7 @@ func (node *BTreeNode[T]) Search(key T) bool {
 
 func (tree *BTree[T]) Search(key T) bool {
 	if tree.root == nil {
-		return false;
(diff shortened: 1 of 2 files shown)


caojoshua:
Could I get a review for this?

caojoshua:
Ping to prevent PR from auto closing. Would still like reviews.

raklaptudirm (approved):

caojoshua:
Code coverage helped me find bugs where root is null after deleting everything. Added a testcase to handle this case. This should also help with line coverage, but I'm not sure how to check that locally.

raklaptudirm:
Seems you haven't formatted some of the changed files, please do so.

raklaptudirm (approved):

PR #705, opened 2024-02-12
style: handle edge cases uniformly in all implementations of factorial

vil02:
Current implementations of the factorial react differently for negative inputs. This PR fixes that. Furthermore it reduced the code duplication in the corresponding test file by expressing tests and benchmarks as _parameterized_.

### DIFF
--- a/math/factorial/factorial.go
+++ b/math/factorial/factorial.go
@@ -13,12 +13,11 @@ import (
 )
 
 var ErrNegativeArgument = errors.New("input argument must be non-negative integer")
-var erroredResult = -1
 
 // Iterative returns the iteratively brute forced factorial of n
 func Iterative(n int) (int, error) {
 	if n < 0 {
-		return erroredResult, ErrNegativeArgument
+		return 0, ErrNegativeArgument
 	}
 	result := 1
 	for i := 2; i <= n; i++ {
@@ -30,7 +29,7 @@ func Iterative(n int) (int, error) {
 // Recursive This function recursively computes the factorial of a number
 func Recursive(n int) (int, error) {
 	if n < 0 {
-		return erroredResult, ErrNegativeArgument
+		return 0, ErrNegativeArgument
 	}
 	if n <= 1 {
 		return 1, nil
@@ -42,7 +41,7 @@ func Recursive(n int) (int, error) {
 // UsingTree This function finds the factorial of a number using a binary tree
 func UsingTree(n int) (int, error) {
 	if n < 0 {
-		return erroredResult, ErrNegativeArgument
+		return 0, ErrNegativeArgument
 	}
 	if n == 0 {
 		return 1, nil
--- a/math/factorial/factorial.go
+++ b/math/factorial/factorial.go
@@ -12,13 +12,13 @@ import (
 	"errors"
 )
 
-var InputMustBeNonnegative = errors.New("Input must be non-negative")
+var ErrNegativeArgument = errors.New("input argument must be non-negative integer")
 var erroredResult = -1
 
 // Iterative returns the iteratively brute forced factorial of n
 func Iterative(n int) (int, error) {
 	if n < 0 {
-		return erroredResult, InputMustBeNonnegative
+		return erroredResult, ErrNegativeArgument
 	}
 	result := 1
 	for i := 2; i <= n; i++ {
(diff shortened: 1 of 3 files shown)


math/factorial/factorial.go
```diff
@@ -8,36 +8,49 @@
 // Package factorial describes algorithms Factorials calculations.
 package factorial
 
+import (
+	"errors"
+)
+
+var InputMustBeNonnegative = errors.New("Input must be non-negative")
```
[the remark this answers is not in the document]
tjgurwara99: `Err*` suffix is a common convention when defining errors.

math/factorial/factorial_test.go
```diff
@@ -1,81 +1,75 @@
...
-func TestBruteForceFactorial(t *testing.T) {
-	tests := getTests()
-	for _, tv := range tests {
-		t.Run(tv.name, func(t *testing.T) {
-			result := Iterative(tv.n)
-			if result != tv.result {
-				t.Errorf("Wrong result! Expected:%d, returned:%d ", tv.result, result)
-			}
-		})
+func implementations() map[string]factorialFun {
```
tjgurwara99, line 12: I think you can define it as a variable inside this test package if you change it to `factorial_test` (this is also fine and one more function call is not gonna affect the performance). I generally ask a question when it comes to these things, is this return value always going to remain static? If so, using un-exported `var` makes sense. If it changes on each function, keep it inside a function because we don't want side-effects.

math/factorial/factorial_test.go
```diff
@@ -1,81 +1,75 @@
...
+			_, error := implFunction(-1)
+			if error != InputMustBeNonnegative {
+				t.Errorf("No error captured for negative input")
 			}
 		})
+		for _, tc := range testCases() {
+			t.Run(fmt.Sprintf("%s_with_input_%d", implName, tc.n), func(t *testing.T) {
+				actual, err := implFunction(tc.n)
+				if err != nil {
+					t.Errorf("Unexpected error captured")
```
tjgurwara99, line 55:
```suggestion
					t.Errorf("unexpected error captured")
```

For some reason, quite a few of linting tools complain about capitalisation of errors (I have no personal preference but my IDE starts complaining 😂 It didn't used to before so I never caught these)

math/factorial/factorial_test.go
```diff
@@ -1,81 +1,75 @@
...
-		Recursive(10)
-	}
-}
-
-func BenchmarkCalculateFactorialUseTree(b *testing.B) {
-	for i := 0; i < b.N; i++ {
-		Recursive(10)
+func BenchmarkFactorial(b *testing.B) {
+	for _, input := range []int{5, 10, 15} {
+		for implName, implFunction := range implementations() {
```
tjgurwara99, line 67: I'm not sure if it does what you think it does (if so, great! If not, this could be a big bug). Due to the inherent nature of for loops pre Go v1.22, the things that run inside goroutines (`b.Run` runs the tests inside `goroutines` if I remember correctly) so it might be doing something that could possibly cause problems (though I'm not sure I have time to track down the problem)

ref: https://go.dev/blog/loopvar-preview

math/factorial/factorial_test.go
vil02: I am not that familiar with Go ecosystem to have a strong opinion here.

math/factorial/factorial.go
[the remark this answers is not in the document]
vil02: Done in 9297cf1611e4e6fe40dbee76e3b8277198f6bc68.

math/factorial/factorial_test.go
vil02: I like to have _immutable_ constants. In any case, have a look at 66e653933c45e95dce442386d043183c9fa5f7b1.

math/factorial/factorial_test.go
vil02: To be sure, I simply did some experiments (with `go version go1.18.1 linux/amd64`). I was adding snippets like
```go
if n == 10 {
    panic("oh no!")
}
```
into the respective functions. While running the benchmarks, the crashes were happening in the expected places.

raklaptudirm (changes requested):
Instead of `erroredResult, ErrNegativeArgument` just return `0, ErrNegativeArgument` since the caller is not supposed to use the result value if `err != nil`.

vil02:
> Instead of `erroredResult, ErrNegativeArgument` just return `0, ErrNegativeArgument` since the caller is not supposed to use the result value if `err != nil`.

Done in e9c894faf8e9aff5cc145731121731b597371c16.

raklaptudirm (approved):

PR #714, opened 2024-04-07
Implement Trapped Rain Water algorithm.

TruongNhanNguyen:
Provides a function to calculate the amount of trapped rainwater between bars represented by an elevation map using dynamic programming.
## Details:
- The `TrapRainWater` function calculates the amount of trapped rainwater between the bars represented by the given elevation map.
- It uses dynamic programming to precompute the maximum height of bars to the left and right of each position. Then, it iterates through the array to calculate the amount of trapped rainwater at each position based on the minimum of the left and right maximum heights.
- Finally, it sums up the trapped rainwater for all positions and returns the total amount


TruongNhanNguyen:
Let's take a look on this PR @raklaptudirm.

raklaptudirm (approved):

PR #716, opened 2024-04-17
feat(uniquepaths): add solution for unique paths problem

rares985:
Provides a solution to the "Unique Paths" problem. That is, in how many ways can you traverse an m x n matrix from (0, 0) to (m-1, n-1)

### DIFF
--- a/dynamic/uniquepaths.go
+++ b/dynamic/uniquepaths.go
@@ -8,9 +8,9 @@ func UniquePaths(m, n int) int {
 		return 0
 	}
 
-	grid := make([][]int, m+1)
+	grid := make([][]int, m)
 	for i := range grid {
-		grid[i] = make([]int, n+1)
+		grid[i] = make([]int, n)
 	}
 
 	for i := 0; i < m; i++ {
@@ -21,8 +21,8 @@ func UniquePaths(m, n int) int {
 		grid[0][j] = 1
 	}
 
-	for i := 1; i <= m; i++ {
-		for j := 1; j <= n; j++ {
+	for i := 1; i < m; i++ {
+		for j := 1; j < n; j++ {
 			grid[i][j] = grid[i-1][j] + grid[i][j-1]
 		}
 	}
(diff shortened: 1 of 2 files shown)


TruongNhanNguyen:
This looks like a good implementation of the dynamic programming solution for the "Unique Paths" problem. One minor adjustment I would suggest is to remove the `+1` when creating the grid. Since you're using zero-based indexing, allocating `m+1` and `n+1` for the grid creates unnecessary space. Simply use `m` and `n` instead. This adjustment makes the code a bit more concise. Other than that, the logic and implementation seem correct!

dynamic/uniquepaths.go
```diff
@@ -0,0 +1,31 @@
...
+// author: Rares Mateizer (https://github.com/rares985)
+package dynamic
+
+// UniquePaths implements the solution to the "Unique Paths" problem
+func UniquePaths(m, n int) int {
+	if m <= 0 || n <= 0 {
+		return 0
+	}
+
+	grid := make([][]int, m+1)
```
TruongNhanNguyen, line 11:
```suggestion
	grid := make([][]int, m)
```

dynamic/uniquepaths.go
```diff
@@ -0,0 +1,31 @@
...
+	for i := range grid {
+		grid[i] = make([]int, n+1)
```
TruongNhanNguyen, line 13:
```suggestion
		grid[i] = make([]int, n)
```

dynamic/uniquepaths.go
```diff
@@ -0,0 +1,31 @@
...
+	for i := 0; i < m; i++ {
+		grid[i][0] = 1
+	}
+
+	for j := 0; j < n; j++ {
+		grid[0][j] = 1
+	}
+
+	for i := 1; i <= m; i++ {
+		for j := 1; j <= n; j++ {
```
TruongNhanNguyen, line 25:
```suggestion
	for i := 1; i < m; i++ {
		for j := 1; j < n; j++ {
```

rares985:
@yanglbme, @raklaptudirm Could you please review?

raklaptudirm (approved):

PR #720, opened 2024-05-03
feat: add next permutation problem

jafar75:
A practice to find lexicographically next greater permutation of the given array of integers.

- If there does not exist any greater permutation, then print the lexicographically smallest permutation of the given array.
- The implementation below, finds the next permutation in linear time and constant memory and returns in place.


jafar75:
sounds odd! one of the tests if failing that haven't changed since years ago

raklaptudirm (approved):

PR #721, opened 2024-05-09
Project Euler - Problem 1 solution

ddaniel27:
Creation of Project Euler folder and solution for problem 1

Tests and benchmarks added


raklaptudirm (commented):
This seems like a very trivial algorithm which does not provide any educational content, so closing.

PR #727, opened 2024-06-30
[IMPLEMENTATION] RSA algorithm

ddaniel27:
This is my implementation for RSA crypto algorithm. There is another implementation in the repo, but that file just contains the encryption/decryption algorithm assuming that we have the private/public key and modulus. My implementation contains:

- RSA struct to easily manage keys and methods.
- Keys Generation (with random prime numbers selection).
- String encryption/decryption.
- Number encryption/decryption.
- Getters for both (public key, modulus) and (private key).

Tests and benchmarks are included :)

### DIFF
--- a/cipher/rsa/rsa2.go
+++ b/cipher/rsa/rsa2.go
@@ -1,13 +1,24 @@
+/*
+rsa2.go
+description: RSA encryption and decryption including key generation
+details: [RSA wiki](https://en.wikipedia.org/wiki/RSA_(cryptosystem))
+author(s): [ddaniel27](https://github.com/ddaniel27)
+*/
 package rsa
 
 import (
 	"encoding/binary"
 	"fmt"
 	"math/big"
+	"math/rand"
 
 	"github.com/TheAlgorithms/Go/math/gcd"
+	"github.com/TheAlgorithms/Go/math/lcm"
+	"github.com/TheAlgorithms/Go/math/modular"
+	"github.com/TheAlgorithms/Go/math/prime"
 )
 
+// rsa struct contains the public key, private key and modulus
 type rsa struct {
 	publicKey  uint64
 	privateKey uint64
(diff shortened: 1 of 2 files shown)


cipher/rsa/rsa2.go
```diff
@@ -0,0 +1,131 @@
...
+// rsa struct contains the public key, private key and modulus
+type rsa struct {
+	publicKey  uint64
+	privateKey uint64
+	modulus    uint64
+}
+
+// InitRSA initializes the RSA algorithm
+// returns the RSA object
+func InitRSA() *rsa {
```
raklaptudirm, line 30: These kinds of functions usually have the naming convention `New<Thing>`, due to the similar behaviour of the `new` builtin function. However when `<Thing>` is the package name, just `New` also works.

cipher/rsa/rsa2.go
```diff
@@ -0,0 +1,131 @@
...
+	rsa.keyGen()
+	return rsa
+}
+
+// KeyGen generates a key for encryption and decryption with RSA
+// 1. Choose two large prime numbers (p and q) and compute n = p * q
+// 2. Compute the totient of n, lcm(p-1, q-1)
+// 3. Choose an integer e such that 1 < e < lcm(p-1, q-1) and gcd(e, lcm(p-1, q-1)) = 1
+// 4. Compute d such that d * e ≡ 1 (mod lcm(p-1, q-1))
+func (rsa *rsa) keyGen() {
```
raklaptudirm, line 41: This functions seems to be only used in the `InitRSA` function, so you can inline it there.

raklaptudirm (changes requested):

cipher/rsa/rsa2.go
ddaniel27: Makes sense.

cipher/rsa/rsa2.go
ddaniel27: Done!

cipher/rsa/rsa2.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// 4. Compute d such that d * e ≡ 1 (mod totient(n))
+	inv, _ := modular.Inverse(int64(publicKey), int64(totient))
+	privateKey := uint64(inv)
+
+	rsa.publicKey = publicKey
+	rsa.privateKey = privateKey
+	rsa.modulus = modulus
+
+	return rsa
```
raklaptudirm, line 58: Instead of creating a variable setting its values and returning, you can create an `rsa` value directly in the return statement.
```go
return rsa {
    publicKey: publicKey,
    privateKey: privateKey,
    modulus: modulus,
}
```

cipher/rsa/rsa2.go
```diff
@@ -0,0 +1,125 @@
...
+func encryptDecryptInt(e, n, data uint64) uint64 {
+	pow := new(big.Int).Exp(big.NewInt(int64(data)), big.NewInt(int64(e)), big.NewInt(int64(n)))
+	return pow.Uint64()
+}
+
+// randomPrime returns two random prime numbers
+func randomPrime() (uint64, uint64) {
+	sieve := prime.SieveEratosthenes(1000)
+	sieve = sieve[10:] // remove first 10 prime numbers (small numbers)
+	index1 := rand.Intn(len(sieve) - 10)
```
raklaptudirm, line 117: Why the `- 10` here? the length of the sieve already decreased when you resliced it:
```go
// https://go.dev/play/p/yo39EDQbrnA
slice := []int{1, 2, 3, 4, 5, 6, 6, 8, 9, 10}
fmt.Printf("Length of the slice: %d\n", len(slice)) // 10
slice = slice[5:]
fmt.Printf("Length of the new slice: %d\n", len(slice)) // 5
```

cipher/rsa/rsa2.go
```diff
@@ -0,0 +1,125 @@
...
+	index2 := rand.Intn(len(sieve) - 10)
```
raklaptudirm, line 118: See above comment.

raklaptudirm (approved):

cipher/rsa/rsa2.go
ddaniel27: Makes sense, done.

cipher/rsa/rsa2.go
ddaniel27: Sorry, mb, I didn't see I left that `- 10` when fixing some inconsistencies. Fixed.

cipher/rsa/rsa2.go
ddaniel27: Fixed

raklaptudirm (approved):

PR #729, opened 2024-07-10
Improved HashMap implementation

manuelrojas19:
As I was going through the current implementation of HashMap in Go, I noticed that the following improvements could be made:

**Enhanced Collision Handling:**

- Modified the Put method to handle collisions using linked lists (chaining).
- Ensured that new key-value pairs are correctly inserted even when collisions occur.

**Resizing Logic**

- Refactored the resize method to correctly rehash and relocate all nodes.
- The hash map now doubles its capacity when the load factor exceeds 0.75, ensuring efficient operations.

**Refactored Node Retrieval:**

- Implemented getNodeByKey to traverse the linked list at a given hash index to find the node with the matching key.
- Updated the Get method to use getNodeByKey for retrieving values.

**Code Cleanup:**

- Removed redundant functions (newNode and newNodeWithNext).
- Added comments.

**Miscellaneous:**
- Corrected the Make function to return a pointer to the HashMap.
- Improved hashing function for better distribution of keys.

This PR ensures that the HashMap correctly handles collisions, dynamically resizes based on the load factor, and maintains efficient access and modification operations.

### DIFF
--- a/structure/hashmap/hashmap.go
+++ b/structure/hashmap/hashmap.go
@@ -20,16 +20,16 @@ type HashMap struct {
 	table    []*node
 }
 
-// New returns a new HashMap instance
-func New() *HashMap {
+// DefaultNew returns a new HashMap instance with default values
+func DefaultNew() *HashMap {
 	return &HashMap{
 		capacity: defaultCapacity,
 		table:    make([]*node, defaultCapacity),
 	}
 }
 
-// Make creates a new HashMap instance with the specified size and capacity
-func Make(size, capacity uint64) *HashMap {
+// New creates a new HashMap instance with the specified size and capacity
+func New(size, capacity uint64) *HashMap {
 	return &HashMap{
 		size:     size,
 		capacity: capacity,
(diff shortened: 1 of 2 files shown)


yurnero13 (approved):

structure/hashmap/hashmap.go
```diff
@@ -13,119 +13,100 @@ type node struct {
...
 		capacity: defaultCapacity,
 		table:    make([]*node, defaultCapacity),
 	}
 }
 
-// Make creates a new HashMap instance with input size and capacity
-func Make(size, capacity uint64) HashMap {
-	return HashMap{
+// Make creates a new HashMap instance with the specified size and capacity
+func Make(size, capacity uint64) *HashMap {
```
raklaptudirm, line 32: Since it returns a pointer now, rename the function to `New`.

raklaptudirm (changes requested):

structure/hashmap/hashmap.go
manuelrojas19: Renamed function to New on a new commit

raklaptudirm (approved):

PR #735, opened 2024-09-03
feat: add Kahn's algorithm for topological sort

soondubu137:
## Additions
Currently, there is only a DFS implementation for topological sort. Therefore, I'm adding Kahn's algorithm.

`graph/kahn.go`: an implementation of the Kahn's algorithm for topological sort, complemented with detailed comments. My code conforms to the `Graph` struct defined in `graph/graph.go`.
`graph/kahn_test.go`: 11 test cases.

## Notes
1. I suggest renaming `graph/topological.go` to something else to indicate its DFS implementation.
2. `graph/topological_test.go` has a package-level variable called `testCases`. I think it's better to move it into the test function.

### DIFF
--- a/graph/kahn.go
+++ b/graph/kahn.go
@@ -1,0 +1,40 @@
+package graph
+
+func Kahn(n int, dependencies [][]int) []int {
+	g := Graph{vertices: n, Directed: true}
+	inDegree := make([]int, n)
+
+	for _, d := range dependencies {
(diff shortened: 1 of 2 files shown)


soondubu137:
It seems the current codebase has some linting issue. I checked the errors, and I think they arise from existing files. (Same for PR #734)

raklaptudirm (approved):

PR #737, opened 2024-09-14
[IMPLEMENTATION] DSA algorithm implementation

ddaniel27:
Hi team, this is my [DSA](https://en.wikipedia.org/wiki/Digital_Signature_Algorithm) implementation. This PR needs a little bit of context.

I tried to follow the exact same algorithm steps as Wikipedia says but this has a big problem: **efficiency**. I'm talking about waiting around 1.5 mins to get the keys generated. Even this is not the exact amount of time because having randomness adds inconsistency (this is the keygen took 10s or 10m but on average 1.5m).

This is why I took a look at the [current go dsa implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/crypto/dsa/dsa.go). As you might see, we have the same aproach in the generation parameters loop (i.e. GPLoop for me, GeneratePrimes for DSA Go implementation), this is so far the most optimal way I found to get the `p` and `q` params, so I tried to replicated it in my own way also with the added value of explaining most decisions because some decisions seems arbitray but it's all for (again) **efficiency**.

I hope the context helps a little bit to clarify why the core algorithm appears to be a "copy" of the DSA std library. Honestly, after studying many different implementations, this is the only viable one.

### DIFF
--- a/cipher/dsa/dsa.go
+++ b/cipher/dsa/dsa.go
@@ -1,0 +1,196 @@
+/*
+dsa.go
+description: DSA encryption and decryption including key generation
+details: [DSA wiki](https://en.wikipedia.org/wiki/Digital_Signature_Algorithm)
+author(s): [ddaniel27](https://github.com/ddaniel27)
+*/
+package dsa
+
+import (
+	"crypto/rand"
+	"io"
+	"math/big"
+)
+
+const (
+	numMRTests = 64   // Number of Miller-Rabin tests
+	L          = 1024 // Number of bits in p
+	N          = 160  // Number of bits in q
+)
+
+type (
+	// parameters represents the DSA parameters
+	parameters struct {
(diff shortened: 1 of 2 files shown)


ddaniel27:
Hey @raklaptudirm @yanglbme hope you are well. I re-open a PR for this implementation. Last one was closed by `stale` status. I hope you can take a look at this PR this time. Also the pipeline is failing because some lint errors not related with my PR. I could track two errors, both in `math/matrix/strassenmatrixmultiply_test.go` line `69` and `106`.

ddaniel27:
@raklaptudirm @yanglbme I created a [PR](https://github.com/TheAlgorithms/Go/pull/743) to fix the tests failing for lint. After merge that PR You/I can update this PR to merge this too :)

raklaptudirm (approved):

PR #738, opened 2024-09-23
LeetCode 198-House Robber

dcode-github:
Used pick or not pick approach.
If we pick a house **i**, then we skip the next one and move to **i+2**.
If we don't pick a house, then we move to **i+1**.


dcode-github:
Some tests are wrong 
Also didn't follow the pull request guidelines

PR #739, opened 2024-09-23
LeetCode 198-House Robber

dcode-github:
Used Pick or Not Pick approach
If we pick a house **i**, we add the amount of money to our answer and move to **i+2**
If we don't pick house **i**, we move to **i+1**


dcode-github:
Did some mistake

PR #740, opened 2024-09-23
LeetCode 198-House Robber, 139-Word Break

dcode-github:
House Robber-
Used Pick or Not Pick approach
If we pick a house **i**, we add the amount of money to our answer and move to **i+2**
If we don't pick house **i**, we move to **i+1**

Word Break-
Comparing all possibilities and storing the answer in an array for further reference to reduce number of computations.


raklaptudirm:
IMO this doesn't seem to add any educational value to this repository. Would you argue otherwise?

PR #762, opened 2024-10-23
add odd-even-sort

SimonWaldherr:
add ~~sleepsort and~~ odd-even-sort


raklaptudirm:
Sleep sort was by consensus removed from all repositories, so we won't be accepting any implementations of that particular algorithm.

raklaptudirm (approved):

PR #787, opened 2025-10-01
chore: added gnome sort algo

Sachinn-64:
Implemented Gnome sort!!!


Sachinn-64:
@raklaptudirm @yanglbme

PR #800, opened 2026-05-08
test(search): add duplicate-element tests for binary search

dvanhu:
## Summary

Added duplicate-element test coverage for binary search implementations.

Some search algorithms may return the first occurrence of a target, while others may return any valid matching index when duplicate values exist. These tests validate correctness by checking the returned value rather than enforcing a specific duplicate index.

## Changes

* Added duplicate-element tests for:

  * Binary
  * BinaryIterative
* Validated returned indices safely against target values
* Preserved current implementation semantics


dvanhu:
Closing this PR to reopen a cleaner version with isolated test additions only.

